PR #1, opened 2020-11-17
init repo with bare bones project structure

tomsmyers:
the aim of this PR is to seed the repo with a bare bones express project with the GDS and MOJ design assets bundled and ready to use. 

i used https://github.com/ministryofjustice/use-of-force as a starting point, but took out lots of the boilerplate code. much of it will need adding in again, but wanted to keep the initial checkin as simple as possible.


.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
+  "trailingComma": "es5",
+  "singleQuote": true,
+  "printWidth": 120,
+  "semi": false,
```
Gweaton, line 5: How tied are we to this `prettierrc`? Would be nice to use the defaults as much as possible unless people have any strong opinions otherwise?

.eslintrc.json
```diff
@@ -0,0 +1,111 @@
...
+            "argsIgnorePattern": "res|next|^err|_",
+            "ignoreRestSiblings": true
+          }
+        ],
+        "@typescript-eslint/semi": 0,
+        "import/no-unresolved": "error",
+        "prettier/prettier": [
+          "error",
+          {
+            "trailingComma": "es5",
```
Gweaton, line 53: Again, it would be nice to use the defaults if possible!

Gweaton:
Feel free to ignore me on this (you're the tech lead!), but I'm wondering if there's an argument for building this repo from the ground up (in a few separate PRs, commits etcs) using the existing repo as an example, rather than a starting point?

There's a lot going on here, e.g. setting up CI, a new Express App (with some routes), Typescript, Linting, some Logging tools, testing frameworks, docker, which I feel would make sense to add sequentially as and when there's case for them (or even with a few people working on individual bits in parallel), rather than all in one PR.

tomsmyers:
closing in favour of more 'official' template

PR #4, opened 2020-11-20
Remove pre-commit hook

Gweaton:

Removes the husky pre-commit hook that runs all tests and linting before code can be committed.

Although ideally we'd have all tests passing on every commit, we'd like to leave it down to individual devs to run tests locally, or on CI, as running them on every commit slows the development process down.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,11 @@
     "delay": 2500,
     "ext": "js,json,html,njk"
   },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
   "lint-staged": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -72,11 +72,6 @@
     "delay": 2500,
     "ext": "js,json,html,njk"
   },
-  "husky": {
-    "hooks": {
-      "pre-commit": "lint-staged && tsc && npm test"
-    }
-  },
   "lint-staged": {
```
tomsmyers, line 75: we can remove this block too since we are no longer using it

we can also remove the dev dependencies on husky i think

tomsmyers (approved):

PR #5, opened 2020-11-20
add helm config for initial deploys

tomsmyers:

add boilerplate helm config

enable deploys to cloud platform

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,9 +53,6 @@ COPY --from=build --chown=appuser:appgroup \
 COPY --from=build --chown=appuser:appgroup \
         /app/node_modules ./node_modules
 
-COPY --from=build --chown=appuser:appgroup \
-        /app/server/views ./server/views
-
 RUN npm prune --production
 
 EXPOSE 3000
--- a/helm_deploy/hmpps-interventions-ui/templates/_envs.tpl
+++ b/helm_deploy/hmpps-interventions-ui/templates/_envs.tpl
@@ -1,0 +1,62 @@
+{{/* vim: set filetype=mustache: */}}
+{{/*
+Environment variables for web and worker containers
+*/}}
+{{- define "deployment.envs" }}
+env:
+  - name: INGRESS_URL
+    value: "https://{{ .Values.ingress.host }}/"
+
+  - name: CLIENT_ID
+    valueFrom:
+      secretKeyRef:
+        name: {{ template "app.name" . }}
+        key: CLIENT_ID
+
+  - name: CLIENT_SECRET
+    valueFrom:
+      secretKeyRef:
+        name: {{ template "app.name" . }}
+        key: CLIENT_SECRET
+
+  - name: APPINSIGHTS_INSTRUMENTATIONKEY
+    valueFrom:
+      secretKeyRef:
+        name: {{ template "app.name" . }}
+        key: APPINSIGHTS_INSTRUMENTATIONKEY
+
+  - name: SESSION_SECRET
+    valueFrom:
+      secretKeyRef:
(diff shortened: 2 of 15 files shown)


helm_deploy/values-dev.yaml
```diff
@@ -0,0 +1,28 @@
...
+image:
+  repository: quay.io/hmpps/hmpps-interventions-ui
+  tag: latest
+  pullPolicy: IfNotPresent
+  port: 3000
+
+ingress:
+  enabled: true
+  enable_whitelist: true
+  host: hmpps-interventions-ui-dev.apps.live-1.cloud-platform.service.justice.gov.uk
```
tomsmyers, line 15: i changed this to match the change @sldblog made to hmpps-interventions-service. 

we need to track the fact that this needs changing when we have our real urls set up

Dockerfile
```diff
@@ -53,9 +53,6 @@ COPY --from=build --chown=appuser:appgroup \
 COPY --from=build --chown=appuser:appgroup \
         /app/node_modules ./node_modules
 
-COPY --from=build --chown=appuser:appgroup \
```
tomsmyers, line 56: this leaked in from elsewhere. but it might as well get merged. this COPY is superfluous.

Gweaton (approved):
I don't know if you want @sldblog to take a look at this as he'll definitely know more, but as far as I can tell you've replaced the important bits with our new service 🤷

Gweaton (approved):
Approving again - I'm wondering if, in future, it might make sense to rebase with `main` instead of `merge`ing, as we've now got 2 merge commits going into the code base here - not a blocker at all, but something to consider!

tomsmyers:
@Gweaton yeh good point. i'm not sure github likes rebasing much. and i did this merge via the online conflict resolution tool which i don't think has an option for merge/rebase.

helm_deploy/hmpps-interventions-ui/templates/_envs.tpl
```diff
@@ -0,0 +1,62 @@
...
+        key: auth_token
+
+  - name: REDIS_TLS_ENABLED
+    value: {{ .Values.env.REDIS_TLS_ENABLED }}
+    value: "true"
+
+  - name: HMPPS_AUTH_URL
+    value: {{ .Values.env.HMPPS_AUTH_URL | quote }}
+
+  - name: TOKEN_VERIFICATION_API_URL
```
sldblog, line 53: In `feature.env`, this variable has the same host as `HMPPS_AUTH_URL` with a different path.

This intrigues me
- this name is functional -- we're using it for token verification
- the other (`HMPPS_AUTH_URL`) is a system name -- should it also be functional, like `AUTH_URL`?

helm_deploy/hmpps-interventions-ui/templates/ingress.yaml
```diff
@@ -0,0 +1,27 @@
...
+apiVersion: networking.k8s.io/v1beta1
+kind: Ingress
+metadata:
+  name: {{ $fullName }}
+  labels:
+    {{- include "app.labels" . | nindent 4 }}
+  annotations:
+    kubernetes.io/ingress.class: "nginx"
+    nginx.ingress.kubernetes.io/custom-http-errors: "418"
+    {{ if .Values.ingress.enable_whitelist }}nginx.ingress.kubernetes.io/whitelist-source-range: {{ include "app.joinListWithComma" .Values.whitelist | quote }}{{ end }}
```
sldblog, line 13: 🙋‍♂️ This line can be removed if we don't plan to use `.Values.whitelist`

sldblog (commented):
Curiosity questions :)

PR #6, opened 2020-11-20
add snyk scanning of node dependencies

tomsmyers:

adds snyk scanning to build and nightly pipelines:

- scans on PRs show results and fail builds if any known high severity CVEs are found.
- scans on main also fail builds on high severity CVEs, but also post the results to the snyk platform for monitoring
- nightly scans run on the docker image and app dependencies

the job runs on every branch, but only pushes results to the snyk platform on main. 

the build is failed if high severity CVEs are found. 

be aware of CVEs before they end up in production. block deploys with known high severity CVEs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,7 +47,7 @@ jobs:
         default: false
     steps:
       - checkout
-      - npm install -q
+      - run: npm install -q
       - snyk/scan:
           project: '${CIRCLE_PROJECT_REPONAME}'
           monitor-on-build: << parameters.monitor >>
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,6 +2,7 @@ version: 2.1
 
 orbs:
   hmpps: ministryofjustice/hmpps@1.1.3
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -38,17 +39,44 @@ jobs:
...
+          monitor: true
+          filters:
+            branches:
+              only:
+                - main
+      - vulnerability_scan:
+          filters:
+            branches:
+              ignore:
+                - main
```
Gweaton, line 79: Sorry, I'm not an expert on circle ci, so I'm a bit confused about these 2 sections - does this not undo the running of `vulnerability_scan`'s`only` above i.e. not run on `main`?

tomsmyers: the filters are mutually exclusive; one runs only on main, the other on everything except main. the idea is just to run the job with different params for PR builds.

Gweaton: Thanks for clarifying!

Gweaton (approved):

PR #7, opened 2020-11-20
Document key tooling decisions in ADRs

Gweaton:

Adds a bunch of ADRs (Architecture decision records) to document what tools we're using and why.

To document decisions we've made early on in the project and make sure there's a reason for using each tool. This may seem obvious to some, but will be useful for newcomers to the project or those less familiar with MOJ standard tooling.

## To add, if these are at all useful:

- Docker
- Kubernetes
- Helm
- Bunyan

### DIFF
--- a/.adr-dir
+++ b/.adr-dir
@@ -1,0 +1,1 @@
+doc/architecture/decisions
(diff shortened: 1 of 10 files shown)


tomsmyers (approved):
this is awesome. you should open this as a PR on the hmpps-typescript-template repo

tomsmyers:
re: your other suggestions, the only i think we should add is bunyan - the others are sort of out of the scope of our own decision making since they are requirements to run in cloud platform

PR #8, opened 2020-11-24
fix helm deploy to dev

tomsmyers:

3 things:
- use k8s secrets throughout
- remove ingress whitelist 
- fix dev env var names

unblock the dev deploy


helm_deploy/values-dev.yaml
```diff
@@ -11,18 +11,16 @@ image:
...
   HMPPS_COOKIE_NAME: hmpps-session-dev
   REMOTE_AUTH_STRATEGY: "true"
   WEB_SESSION_TIMEOUT_IN_MINUTES: "60"
   VIDEO_LINK_ENABLED_FOR: "WWI"
-  TOKENVERIFICATION_API_URL: https://token-verification-api-dev.prison.service.justice.gov.uk
-  TOKENVERIFICATION_API_ENABLED: true
+  TOKEN_VERIFICATION_API_URL: https://token-verification-api-dev.prison.service.justice.gov.uk
+  TOKEN_VERIFICATION_API_ENABLED: true
   REDIS_ENABLED: true
   SUPPORT_URL: https://support-dev.hmpps.service.justice.gov.uk/
```
sldblog, line 26: I couldn't find `HMPPS_COOKIE_NAME`, `REMOTE_AUTH_STRATEGY`, `VIDEO_LINK_ENABLED_FOR` and `SUPPORT_URL` in the codebase. Could we remove those too?

helm_deploy/hmpps-interventions-ui/templates/_envs.tpl
```diff
@@ -10,26 +10,20 @@ env:
...
-        key: APPINSIGHTS_INSTRUMENTATIONKEY
+        name: hmpps-auth
+        key: client-secret
 
   - name: SESSION_SECRET
     valueFrom:
       secretKeyRef:
-        name: {{ template "app.name" . }}
-        key: SESSION_SECRET
+        name: session
```
sldblog, line 25: _not a blocker_

Should we scope these names somehow so it's obvious they belong to the UI? The kubernetes namespace we use is shared between the UI, and the service deployments, so it may get confusing after a while.

helm_deploy/hmpps-interventions-ui/templates/_envs.tpl
```diff
@@ -10,26 +10,20 @@ env:
...
+        key: secret
 
   - name: REDIS_HOST
```
sldblog, line 28: _not a blocker_

🙋‍♂️ This will be populated from a secret provisioned by an ElastiCache resource in [our namespace](https://github.com/ministryofjustice/cloud-platform-environments/tree/main/namespaces/live-1.cloud-platform.service.justice.gov.uk/hmpps-interventions-dev/resources).
(We don't have this resource yet.)

❓ I wonder if it would make sense to somehow include the purpose in one of these names, so it's immediately visible that we use it for session storage?

helm_deploy/values-dev.yaml
```diff
@@ -11,18 +11,16 @@ image:
 
 ingress:
   enabled: true
-  enable_whitelist: true
```
sldblog, line 14: Could we also remove the `if .Values.ingress.enable_whitelist` block in the `ingress.yaml`, please?

sldblog (commented):
Looks great!

Couple of small suggestions and a question around `API_CLIENT_ID` vs `CLIENT_ID` in secrets.

helm_deploy/secrets-example.yaml
```diff
@@ -1,6 +1,8 @@
 ---
 secrets:
   APPINSIGHTS_INSTRUMENTATIONKEY: somesecretvalue
-  CLIENT_ID: somesecretvalue
-  CLIENT_SECRET: somesecretvalue
+  API_CLIENT_ID: clientid
+  API_CLIENT_SECRET: clientsecret
```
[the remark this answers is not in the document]
tomsmyers: sorry i'm getting ahead of myself here. in an upcoming PR i'm going to add seperate sets of creds for user facing login vs oauth2 secured api access. since this file is inconsequential in terms of deploy we can leave it for now if it's ok with you.

helm_deploy/values-dev.yaml
tomsmyers: can do

helm_deploy/hmpps-interventions-ui/templates/_envs.tpl
tomsmyers: oh damn - i thought we had seperate namespaces. i need to fix up some of the secret names in that case.

and yes to scoping these types of var.

helm_deploy/values-dev.yaml
tomsmyers: yep

helm_deploy/hmpps-interventions-ui/templates/_envs.tpl
sldblog: To be honest, I made an (invisible) assumption there that we don't need separate ones -- sorry about that. If it makes more sense we can definitely have everything on its own?

tomsmyers: i'm fine with it as it is - especially since k8s secrets are really just a stop-gap(?)

helm_deploy/hmpps-interventions-ui/templates/_envs.tpl
tomsmyers: yeh we can do. these secret names have no bearing on what we call the secrets in AWS, so we can address this when we migrate off k8s secrets?

helm_deploy/hmpps-interventions-ui/templates/_envs.tpl
sldblog: I think, ultimately, the deployment _must_ populate Kubernetes secrets. I think we'll change the mechanism of _how_ they get there (currently: manual, later: CI pulls them down from AWS then pushes it through Helm) so names will still be a concern. However, this isn't a blocker for this PR 👍

helm_deploy/hmpps-interventions-ui/templates/_envs.tpl
sldblog: We may need to address it before (but definitely not in this PR 😄) as Kubernetes secrets will stay as a mechanism used in runtime to inject secrets.

When we look at the Kubernetes namespace (runtime) and the codebase, I feel we should be able to reason about how they connect together.

AWS secrets manager values will be invisible in both contexts and to avoid an additional level of mapping in the helm templates we will probably have those values match the ENV var names or the secret names -- so I get the sense we will not completely detach ourselves from these names.

sldblog (approved):

PR #9, opened 2020-11-25
Add sample code for delius integration via community API.

tomsmyers:

This consists of:
 - differentiate between oauth clients for login vs API access
 - add community api docker service
 - add 'service' for calling community api endpoint
 - add class for sample integration routes

demonstrate delius integration + discuss coding patterns

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,5 @@
 version: '3.1'
 services:
-
   redis:
     image: 'bitnami/redis:5.0'
     networks:
@@ -17,12 +16,26 @@ services:
       - hmpps
     container_name: hmpps-auth 
     ports:
-      - "9090:8080"
+      - "8090:8080"
     healthcheck:
       test: ["CMD", "curl", "-f", "http://localhost:8080/auth/health"]
     environment:
+      - SERVER_PORT=8080
       - SPRING_PROFILES_ACTIVE=dev
-      - DELIUS_ENABLED=false
+      - SPRING_JPA_PROPERTIES_HIBERNATE_SHOW_SQL=false
+
+  community-api:
+    image: quay.io/hmpps/community-api:latest
+    networks:
+      - hmpps
+    container_name: community-api
+    ports:
+      - "8091:8080"
+    healthcheck:
+      test: [ "CMD", "curl", "-f", "http://localhost:8080/health" ]
+    environment:
+      - SERVER_PORT=8080
+      - SPRING_PROFILES_ACTIVE=dev
 
 networks:
   hmpps:
--- a/feature.env
+++ b/feature.env
@@ -1,6 +1,4 @@
 PORT=3007
-HMPPS_AUTH_URL=http://localhost:9091/auth
-TOKEN_VERIFICATION_API_URL=http://localhost:9091/verification
+HMPPS_AUTH_URL=http://localhost:8090/auth
+TOKEN_VERIFICATION_API_URL=http://localhost:8092/verification
 NODE_ENV=development
-CLIENT_ID=clientid
-CLIENT_SECRET=clientsecret
--- a/helm_deploy/hmpps-interventions-ui/templates/_envs.tpl
+++ b/helm_deploy/hmpps-interventions-ui/templates/_envs.tpl
@@ -7,23 +7,35 @@ env:
   - name: INGRESS_URL
     value: "https://{{ .Values.ingress.host }}/"
 
-  - name: CLIENT_ID
+  - name: API_CLIENT_ID
     valueFrom:
       secretKeyRef:
         name: hmpps-auth
-        key: client-id
+        key: interventions-ui-client-id
 
-  - name: CLIENT_SECRET
+  - name: API_CLIENT_SECRET
     valueFrom:
       secretKeyRef:
         name: hmpps-auth
-        key: client-secret
+        key: interventions-ui-client-secret
+
+  - name: LOGIN_CLIENT_ID
+    valueFrom:
+      secretKeyRef:
+        name: hmpps-auth
+        key: interventions-login-client-id
+
+  - name: LOGIN_CLIENT_SECRET
+    valueFrom:
+      secretKeyRef:
+        name: hmpps-auth
+        key: interventions-login-client-secret
 
   - name: SESSION_SECRET
     valueFrom:
       secretKeyRef:
         name: session
-        key: ui-session-secret
+        key: interventions-ui-session-secret
 
   - name: REDIS_HOST
     valueFrom:
--- a/server/config.ts
+++ b/server/config.ts
@@ -50,16 +50,25 @@ export default {
     expiryMinutes: get('WEB_SESSION_TIMEOUT_IN_MINUTES', '120'),
   },
   apis: {
+    communityApi: {
+      url: get('COMMUNITY_API_URL', 'http://localhost:8091', requiredInProduction),
+      timeout: {
+        response: get('COMMUNITY_API_TIMEOUT_RESPONSE', 10000),
+        deadline: get('COMMUNITY_API_TIMEOUT_DEADLINE', 10000),
+      },
+      agent: new AgentConfig(),
+    },
     hmppsAuth: {
-      url: get('HMPPS_AUTH_URL', 'http://localhost:9090/auth', requiredInProduction),
-      externalUrl: get('HMPPS_AUTH_EXTERNAL_URL', get('HMPPS_AUTH_URL', 'http://localhost:9090/auth')),
+      url: get('HMPPS_AUTH_URL', 'http://localhost:8090/auth', requiredInProduction),
       timeout: {
         response: get('HMPPS_AUTH_TIMEOUT_RESPONSE', 10000),
         deadline: get('HMPPS_AUTH_TIMEOUT_DEADLINE', 10000),
       },
       agent: new AgentConfig(),
-      clientId: get('CLIENT_ID', 'sentence-plan-client', requiredInProduction),
-      clientSecret: get('CLIENT_SECRET', 'clientsecret', requiredInProduction),
+      apiClientId: get('API_CLIENT_ID', 'interventions', requiredInProduction),
+      apiClientSecret: get('API_CLIENT_SECRET', 'clientsecret', requiredInProduction),
+      loginClientId: get('LOGIN_CLIENT_ID', 'interventions', requiredInProduction),
+      loginClientSecret: get('LOGIN_CLIENT_SECRET', 'clientsecret', requiredInProduction),
     },
     tokenVerification: {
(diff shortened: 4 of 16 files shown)


docker-compose.yml
```diff
@@ -17,12 +16,26 @@ services:
       - hmpps
     container_name: hmpps-auth 
     ports:
-      - "9090:8080"
+      - "8090:8080"
```
tomsmyers, line 19: consistency with hmpps-interventions-service docker compose will make developing on both repos easier

docker-compose.yml
```diff
@@ -17,12 +16,26 @@ services:
...
     healthcheck:
       test: ["CMD", "curl", "-f", "http://localhost:8080/auth/health"]
     environment:
+      - SERVER_PORT=8080
+      - SPRING_PROFILES_ACTIVE=dev
+      - SPRING_JPA_PROPERTIES_HIBERNATE_SHOW_SQL=false
```
tomsmyers, line 25: i find this log output annoying when trying to verify token exchange in logs

feature.env
```diff
@@ -1,6 +1,4 @@
 PORT=3007
-HMPPS_AUTH_URL=http://localhost:9091/auth
-TOKEN_VERIFICATION_API_URL=http://localhost:9091/verification
+HMPPS_AUTH_URL=http://localhost:8090/auth
+TOKEN_VERIFICATION_API_URL=http://localhost:8092/verification
 NODE_ENV=development
-CLIENT_ID=clientid
```
tomsmyers, line 5: i actually don't know what this stuff is really for. removed the env vars to hopefully break something then we'll find out...

server/config.ts
```diff
@@ -50,16 +50,25 @@ export default {
...
+      url: get('COMMUNITY_API_URL', 'http://localhost:8091', requiredInProduction),
+      timeout: {
+        response: get('COMMUNITY_API_TIMEOUT_RESPONSE', 10000),
+        deadline: get('COMMUNITY_API_TIMEOUT_DEADLINE', 10000),
+      },
+      agent: new AgentConfig(),
+    },
     hmppsAuth: {
-      url: get('HMPPS_AUTH_URL', 'http://localhost:9090/auth', requiredInProduction),
-      externalUrl: get('HMPPS_AUTH_EXTERNAL_URL', get('HMPPS_AUTH_URL', 'http://localhost:9090/auth')),
```
tomsmyers, line 55: couldn't see the reason for having two urls here

server/config.ts
```diff
@@ -50,16 +50,25 @@ export default {
...
+      url: get('HMPPS_AUTH_URL', 'http://localhost:8090/auth', requiredInProduction),
       timeout: {
         response: get('HMPPS_AUTH_TIMEOUT_RESPONSE', 10000),
         deadline: get('HMPPS_AUTH_TIMEOUT_DEADLINE', 10000),
       },
       agent: new AgentConfig(),
-      clientId: get('CLIENT_ID', 'sentence-plan-client', requiredInProduction),
-      clientSecret: get('CLIENT_SECRET', 'clientsecret', requiredInProduction),
+      apiClientId: get('API_CLIENT_ID', 'interventions', requiredInProduction),
```
tomsmyers, line 68: these are the same when running locally - but changes to helm config coming soon will illustrate that these are different clients.

server/routes/index.ts
```diff
@@ -1,13 +1,23 @@
 import type { RequestHandler, Router } from 'express'
 
 import asyncMiddleware from '../middleware/asyncMiddleware'
+import CommunityApiService from '../services/communityApiService'
+import IntegrationSamplesRoutes from './integrationSamples'
 
-export default function routes(router: Router): Router {
+interface RouteProvider {
```
tomsmyers, line 7: trying to make a start on nailing down exactly how we want to split out our controllers.

we can make this more specific, e.g. we could map http verbs (an enum of get,put,delete etc) to request handlers or whatever we want. 

this is just to get us thinking about it

server/routes/index.ts
```diff
@@ -1,13 +1,23 @@
...
+export default function routes(router: Router, communityApiService: CommunityApiService): Router {
   const get = (path: string, handler: RequestHandler) => router.get(path, asyncMiddleware(handler))
 
+  const integrationSamples: RouteProvider = IntegrationSamplesRoutes(communityApiService)
+
   get('/', (req, res, next) => {
     res.render('pages/index')
   })
 
+  get('/integrations/delius/user', integrationSamples.viewDeliusUserSample)
```
tomsmyers, line 20: again, just getting the ball rolling - instead of http verb i tried a more descriptive name

server/routes/testutils/appSetup.ts
```diff
@@ -8,29 +8,9 @@ import allRoutes from '../index'
...
-
-const user = {
-  name: 'john smith',
-  firstName: 'john',
-  lastName: 'smith',
-  username: 'user1',
-  displayName: 'John Smith',
-}
-
-class MockUserService extends UserService {
```
tomsmyers, line 22: argh - mocks!

server/routes/testutils/appSetup.ts
```diff
@@ -8,29 +8,9 @@ import allRoutes from '../index'
...
-
-  async getUser(token: string) {
-    return {
-      token,
-      ...user,
-    }
-  }
-}
+import { user, MockUserService } from './mocks/mockUserService'
+import MockCommunityApiService from './mocks/mockCommunityApiService'
```
tomsmyers, line 13: argh more *NEW* mocks !

server/routes/testutils/mocks/mockCommunityApiService.ts
```diff
@@ -0,0 +1,19 @@
...
+  constructor() {
+    super(undefined)
+  }
+
+  async getUserByUsername(username: string) {
+    return {
+      userId: '987123876',
+      username: 'maijam',
+      firstName: 'Maija',
+      surname: 'Meikäläinen',
```
tomsmyers, line 13: always good to test unicode support :)

server/services/communityApiService.ts
```diff
@@ -0,0 +1,33 @@
+import type HmppsAuthClient from '../data/hmppsAuthClient'
+import RestClient from '../data/restClient'
+import config from '../config'
+import logger from '../../log'
+
+interface DeliusUser {
```
tomsmyers, line 6: is this an appropriate place for these ?

server/routes/index.ts
```diff
@@ -1,13 +1,23 @@
...
 
-export default function routes(router: Router): Router {
+interface RouteProvider {
+  [key: string]: RequestHandler
+}
+
+export default function routes(router: Router, communityApiService: CommunityApiService): Router {
   const get = (path: string, handler: RequestHandler) => router.get(path, asyncMiddleware(handler))
 
+  const integrationSamples: RouteProvider = IntegrationSamplesRoutes(communityApiService)
```
lawrence-forooghian, line 14: I'm wondering what the `RouteProvider` interface gives us here - it seems to me like we're throwing away some useful type information by downcasting the `IntegrationSamplesRoutes` to a `RouteProvider` - namely that it has a `viewDeliusUserSample` property. But I'm still very much getting to grips with TS and possibly applying faulty mappings from type systems in other languages…

server/services/communityApiService.ts
lawrence-forooghian: I think we briefly talked about this the other day when we were wondering why that `types.ts` file existed in the repo George was showing us. It probably comes down to whether we think that every file that uses `DeliusUser` is also going to use `CommunityApiService`. Like, if `DeliusUser` is going to end up being passed between files all around the app, then do we want all of those files to have to import `CommunityApiService` even if they're not actually making API calls? (The answer may be "yes, that's fine", in which case this seems like a fine place to have these types)

server/routes/testutils/appSetup.ts
```diff
@@ -58,5 +38,5 @@ function appSetup(route: Router, production: boolean): Express {
 
 export default function appWithAllRoutes({ production = false }: { production?: boolean }): Express {
   auth.default.authenticationMiddleware = () => (req, res, next) => next()
-  return appSetup(allRoutes(standardRouter(new MockUserService())), production)
+  return appSetup(allRoutes(standardRouter(new MockUserService()), new MockCommunityApiService()), production)
```
lawrence-forooghian, line 41: So we're going for dependency injection and explicit mocks created by subclasses? In the hmpps-risk-assessment-ui codebase [they instead just use Jest mocking](https://github.com/ministryofjustice/hmpps-risk-assessment-ui/blob/6659d42687fa92e892b605c3ac6ad856195d0b38/app/assessmentsList/get.controller.test.js#L50) to mock out methods on the services themselves. I quite like dependency injection but I wonder whether we could still use Jest mocking instead of needing to create subclasses, might be more "standard" looking for other JS devs.

lawrence-forooghian (commented):
Looks good, thanks for getting the thinking started about this stuff! A note re commits - I would have found it easier to review if the refactoring (renames) had been split out into a separate commit, would have been easier to see what's new functionality.

server/routes/index.ts
tomsmyers: my thinking here was just, "let's see if we can create a type for these 'controller' objects that will encourage consistency". i wouldn't say we're downcasting, just saying that `IntegrationSamplesRoutes` implements `RouteProvider`

helm_deploy/hmpps-interventions-ui/templates/_envs.tpl
```diff
@@ -7,23 +7,35 @@ env:
...
+      secretKeyRef:
+        name: hmpps-auth
+        key: interventions-login-client-secret
 
   - name: SESSION_SECRET
     valueFrom:
       secretKeyRef:
         name: session
-        key: ui-session-secret
+        key: interventions-ui-session-secret
```
tomsmyers, line 38: are these secret names better for you @sldblog ?

sldblog: Really sorry @tomsmyers, my mistake, I wasn't clear before

I meant the `name` instead of the `key`, from the use case of finding out which secret is used by which service. If we intend to share `session` between UI and service then all is well.

```
$ k get secrets --show-labels=true --namespace=hmpps-interventions-dev
NAME                                                TYPE                                  DATA   AGE     LABELS
circleci-token-24d6f                                kubernetes.io/service-account-token   3      9d      <none>
default-token-gk759                                 kubernetes.io/service-account-token   3      12d     <none>
hmpps-auth                                          Opaque                                6      83m     <none>
session                                             Opaque                                1      82m     <none>
...[snip helm stuff]...
```

`key: ui-session-secret` is a great name -- sorry again, I wasn't looking at `key` before.

Gweaton (approved):
Let's get this merged then we can refine once we start needing to actually do Delius integration work

PR #12, opened 2020-11-27
Set up Pact for contract testing

lawrence-forooghian:

This sets up [Pact](https://pact.io/), a contract testing framework. We're going to try this out as a way of defining the contract with the interventions service, and as a way to provide mocks for testing our client class.

It creates a very barebones client class, just enough for us to test.

To allow us to start building features for the frontend.

### DIFF
--- a/server/services/interventionsService.test.ts
+++ b/server/services/interventionsService.test.ts
@@ -1,0 +1,41 @@
+import { pactWith } from 'jest-pact'
+import { Matchers } from '@pact-foundation/pact'
+
+import InterventionsService from './interventionsService'
+import config from '../config'
+
+pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, provider => {
+  let interventionsService: InterventionsService
+
+  beforeEach(() => {
+    const testConfig = { ...config.apis.interventionsService, url: provider.mockService.baseUrl }
+    interventionsService = new InterventionsService(testConfig)
+  })
+
+  describe('getReferral', () => {
+    beforeEach(async () => {
+      // This is just an example for getting Pact set up; we’ll properly design this endpoint later
+      await provider.addInteraction({
+        state: 'There is an existing referral with ID of 1',
+        uponReceiving: 'a request for that referral',
+        withRequest: {
+          method: 'GET',
+          path: '/referrals/1',
+          headers: { Accept: 'application/json' },
+        },
+        willRespondWith: {
+          status: 200,
+          body: Matchers.like({
+            id: '1',
+          }),
+          headers: { 'Content-Type': 'application/json' },
+        },
+      })
+    })
+
+    it('returns a referral for the given ID', async () => {
+      const referral = await interventionsService.getReferral('1')
+      expect(referral.id).toBe('1')
+    })
+  })
+})
--- a/server/services/interventionsService.ts
+++ b/server/services/interventionsService.ts
@@ -1,0 +1,23 @@
+import RestClient from '../data/restClient'
+import logger from '../../log'
+import { ApiConfig } from '../config'
+
+interface Referral {
+  id: string
+}
+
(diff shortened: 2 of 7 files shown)


server/services/interventionsService.test.ts
```diff
@@ -0,0 +1,50 @@
+import { pactWith } from 'jest-pact'
+import InterventionsService from './interventionsService'
+import config from '../config'
+
+pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, provider => {
+  let interventionsService: InterventionsService
+
+  beforeEach(() => {
+    const testConfig = { ...config.apis.interventionsService, url: provider.mockService.baseUrl }
```
tomsmyers, line 9: do we need the extra url param here. the url is inside the config structure - can we just change it there?

server/services/interventionsService.test.ts
```diff
@@ -0,0 +1,50 @@
...
+          path: '/referrals/1',
+          headers: { Accept: 'application/json' },
+        },
+        willRespondWith: {
+          status: 200,
+          body: {
+            id: '1',
+            desiredOutcomes: [
+              {
+                id: '1',
```
tomsmyers, line 30: in the example code i read, these were like matcher objects, which looked to basically check the structure of the returned data, not nessecarily the values. is this how this works too?

server/services/interventionsService.ts
```diff
@@ -0,0 +1,29 @@
+import RestClient from '../data/restClient'
+import logger from '../../log'
+import { ApiConfig } from '../config'
+
+interface Referral {
+  id: string
```
tomsmyers, line 6: i know there is a comment above about this just being example code, but it's a good opportunity to think about what sort of structure these objects will have. will these IDs be uuids, or an incrementing index, or something else? we could make a type alias 

```
type Uuid = string
```

or call the `id` field `uuid`

```interface Referral { uuid: string } ```

to clarify?

File: server/services/interventionsService.ts [hunk 4]
```diff
@@ -0,0 +1,29 @@
...
+  desiredOutcomes: DesiredOutcome[]
+}
+
+interface DesiredOutcome {
+  id: string
+  text: string
```
tomsmyers, line 12: would prefer more descriptive field name - maybe `description`?

tomsmyers: also, do these need IDs? i guess all these questions are not super relevant for this PR. we can discuss in a different forum.

server/services/interventionsService.test.ts
tomsmyers: sorry - i read this wrong, looks good as is.

tomsmyers (commented):
looks good! would be good to generate the pacts as build artifacts until we get a broker set up.

server/services/interventionsService.test.ts
Gweaton: I've converted this to use `Matcher.like` so the value isn't as important, which will be more useful once the contract artifacts are generated.

server/services/interventionsService.ts
Gweaton: Have removed desired outcomes for now.

server/services/interventionsService.ts
Gweaton: We were thinking this might be more of a back-end concern, as we'll probably just use this id to make POST requests, so we were thinking for our purposes a `string` type would probably be enough, but maybe we can chat about this once we know a bit more about how our `Referral`s are going to look and be used?

lawrence-forooghian: Yup, ditto what George said - the exact structure of an ID seems like a backend implementation detail to me.

tomsmyers: 👍🏼

tomsmyers (approved):

PR #18, opened 2020-12-01
fix integration tests locally

tomsmyers:

fixes the integration tests locally.

enable us to validate changes by running the integration tests. and next get these tests running in circleCI


lawrence-forooghian (approved):
The tests are passing locally for me so I'm happy with this 👍

README.md
```diff
@@ -21,34 +21,25 @@ npm run start:dev
...
-## Integration tests
+### Integration Test
 
-For local running, start a test db, redis, and wiremock instance by:
+The integration tests require a different docker-compose stack and a different application configuration. Run each of the following commands in its own shell. 
 
 `docker-compose -f docker-compose-test.yml up`
 
-Then run the server in test mode by:
+`npm start:test`
```
lawrence-forooghian, line 38: @Gweaton realised that this should be `npm run start:test`

lawrence-forooghian (changes requested):

README.md
tomsmyers: done

Gweaton (dismissed):
I think it would be helpful in future for reviewing if you could add a bit more of a description in the commit message (or PR descripton) as to why they were failing and how this fix solves the problem, as I'm not exactly sure which part of the commit fixes the tests (without digging a lot deeper). e.g I assume it's the changing of the `client_secret`, but it could also be the upgrade of the redis image, the new test `.env`, the changes to the `package.json` or all four!

Thanks for sorting this out.

lawrence-forooghian (changes requested):
I don't think that the "address review comment - fix typo in readme" is a useful contribution to the commit history - could you please instead fixup your original commit? This is in line with our [ADR](https://github.com/ministryofjustice/hmpps-architecture-as-code/blob/4e497571f1a82eb07cfb7457e2e1825be52968e1/decisions/interventions/0003-use-git-commits-to-explain-our-work.md#L88-L89):

> If we make changes to a pull request, we will re-structure our commits to make sure that they continue to tell a useful story.

lawrence-forooghian:
> I think it would be helpful in future for reviewing if you could add a bit more of a description in the commit message (or PR descripton) as to why they were failing and how this fix solves the problem, as I'm not exactly sure which part of the commit fixes the tests (without digging a lot deeper)!

Yeah, at least a reference to the commit which broke the tests (cacde25) would be super helpful.

lawrence-forooghian (approved):

Gweaton (approved):
Thanks for the changes!

lawrence-forooghian (approved):

PR #19, opened 2020-12-01
Add integration test

sldblog:

Adds an integration test job to CI.

So we have as quick feedback as possible.

## Where should the reviewer start?

Please review commit-by-commit.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,46 +3,21 @@ version: 2.1
 orbs:
   hmpps: ministryofjustice/hmpps@1.1.3
   snyk: snyk/snyk@0.0.12
+  node: circleci/node@4.1.0
 
 jobs:
   build:
     executor: hmpps/node
     steps:
       - checkout
-      - run:
-          name: Update npm
-          command: 'sudo npm install -g npm@latest'
-      - restore_cache:
-          key: dependency-cache-{{ checksum "package-lock.json" }}
-      - run:
-          name: Install Dependencies
-          command: npm ci
-      - save_cache:
-          key: dependency-cache-{{ checksum "package-lock.json" }}
-          paths:
-            - node_modules
-            - ~/.cache
-      - run:
-          command: npm run test
+      - node/install-npm
+      - node/install-packages
+      - run: npm run test
       - store_artifacts:
           path: 'pact/pacts'
-      - run:
-          command: |
-            npm run build
-            DATE=$(date '+%Y-%m-%d')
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -3,46 +3,52 @@ version: 2.1
...
+executors:
+  integration:
+    docker:
+      - image: circleci/node:10-buster-browsers
+      - image: rodolpheche/wiremock:2.27.2-alpine
+        command: ["--port", "9091"]
+      - image: bitnami/redis:6.0
+        environment:
+          ALLOW_EMPTY_PASSWORD: "yes"
+      - image: quay.io/hmpps/hmpps-auth:latest
```
tomsmyers, line 17: this isn't needed - it's covered by wiremock

sldblog: removed in 275fe55

.circleci/config.yml
```diff
@@ -3,46 +3,21 @@ version: 2.1
...
       - store_artifacts:
           path: 'pact/pacts'
-      - run:
-          command: |
-            npm run build
-            DATE=$(date '+%Y-%m-%d')
-            export BUILD_NUMBER=${DATE}.${CIRCLE_BUILD_NUM}
-            export GIT_REF="$CIRCLE_SHA1"
-            npm run record-build-info
-      - run: # Run linter after build because the integration test code depends on compiled typescript...
```
lawrence-forooghian, line 36: Could we keep this comment in? I don't actually know _what_ it means but seems like somebody thought it was worth pointing out...

.circleci/config.yml
```diff
@@ -14,11 +14,6 @@ executors:
       - image: bitnami/redis:6.0
         environment:
           ALLOW_EMPTY_PASSWORD: "yes"
-      - image: quay.io/hmpps/hmpps-auth:latest
```
lawrence-forooghian: Can we make this a fixup of the previous commit? I think in general that a change introduced and removed in the same PR should just be wiped from the commit history.

lawrence-forooghian (changes requested):

.circleci/config.yml
sldblog: Done in c45f996: I opted to copy the entire `run lint` step into the integration_test job, following the comment

.circleci/config.yml
sldblog: done in c45f996

lawrence-forooghian (approved):

PR #21, opened 2020-12-02
Enable UI deployment

sldblog:

Amends the helm template and Docker build to allow Kubernetes deployments.

To re-enable deployment to the Cloud Platform.

🙏 Please review commit-by-commit.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@ WORKDIR /app
 RUN apt-get update && \
     apt-get upgrade -y
 
-# Build stage 2.
+# Stage: build assets
 FROM base as build
 ARG BUILD_NUMBER
 ARG GIT_REF
@@ -33,6 +33,8 @@ RUN CYPRESS_INSTALL_BINARY=0 npm ci --no-audit && npm run build  && \
     export BUILD_NUMBER=${BUILD_NUMBER} && \
     export GIT_REF=${GIT_REF} && \
     npm run record-build-info
+
+RUN npm prune --no-audit --production
 
 # Build stage 3.
 FROM base
@@ -52,8 +54,6 @@ COPY --from=build --chown=appuser:appgroup \
 
 COPY --from=build --chown=appuser:appgroup \
         /app/node_modules ./node_modules
-
-RUN npm prune --production
 
 EXPOSE 3000
 ENV NODE_ENV='production'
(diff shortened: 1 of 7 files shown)


helm_deploy/hmpps-interventions-ui/templates/_envs.tpl
```diff
@@ -11,42 +11,42 @@ env:
     valueFrom:
       secretKeyRef:
         name: hmpps-auth
-        key: interventions-ui-client-id
+        key: interventions-ui-client-id.txt
```
tomsmyers, line 14: this is annoying - something must have gone wrong when i uploaded the secrets.

sldblog: I can also change it on the namespace side if you'd like?

Dockerfile
```diff
@@ -53,8 +55,6 @@ COPY --from=build --chown=appuser:appgroup \
 COPY --from=build --chown=appuser:appgroup \
         /app/node_modules ./node_modules
 
-RUN npm prune --production
```
tomsmyers, line 56: what's the reason for moving this ? doesn't this mean that the node_modules on the final image contains the dev dependencies?

Dockerfile
```diff
@@ -34,6 +34,8 @@ RUN CYPRESS_INSTALL_BINARY=0 npm ci --no-audit && npm run build  && \
     export GIT_REF=${GIT_REF} && \
     npm run record-build-info
 
+RUN npm prune --no-audit --production
```
tomsmyers, line 39: what's the reason for moving this here?

sldblog: Sorry, I've added the narrative about this in 9d366b4ab1cb1e4928cf597da9785a70f111ffba:

> Prune production dependencies in the build stage
>
> Removing the packages in the final stage will not result in size savings
> as COPY layer will be retained: we need to remove the unneeded packages
> in the `build` phase so the `COPY --from=build` commands copy less.

sldblog: The difference in size between current `main` and this branch's Docker build:
```
$ docker images | grep 'ui-'
ui-before  latest                  6b974cc88063        2 seconds ago       425MB
ui-after   latest                  250473b4e6a7        2 minutes ago       226MB
```

tomsmyers (approved):

PR #22, opened 2020-12-02
Enable noImplicitAny TypeScript compiler option

lawrence-forooghian:

Enables the `noImplicitAny` TypeScript compiler option and fixes the resulting build failures.

To improve type safety and codebase understandability. See commit message for more details.


server/data/hmppsAuthClient.test.ts
```diff
@@ -14,8 +14,16 @@ jest.mock('redis', () => ({
   set: jest.fn().mockImplementation((key, value, command, ttl, callback) => callback(null, null)),
 }))
 
+interface MockRedis {
+  on: jest.Mock
+  get: jest.Mock
+  set: jest.Mock
+}
+
+const mockRedis = (redis as unknown) as MockRedis
```
tomsmyers, line 23: how come we need the intermediate cast to `unknown` here?

lawrence-forooghian: Without it you get this error message from the compiler:

```
    server/data/hmppsAuthClient.test.ts:23:19 - error TS2352: Conversion of type 'typeof import("/Users/lawrence/code/interventions/hmpps-interventions-ui/node_modules/@types/redis/index")' to type 'MockRedis' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
      Type 'typeof import("/Users/lawrence/code/interventions/hmpps-interventions-ui/node_modules/@types/redis/index")' is missing the following properties from type 'MockRedis': on, get, set
```

I think it's because there's no way for the compiler to know that we've dynamically added the `on` / `get` / `set` methods in the `jest.mock` call, but I'm not sure exactly why that means the `unknown` is needed.

lawrence-forooghian: (This whole mock is a bit weird-feeling though, but I couldn't find a better way of doing it using the stuff that Jest / ts-jest gives you. Probably a limitation of my JS / TS knowledge though so might be able to revisit it once I'm a bit more clued up.)

server/middleware/authorisationMiddleware.test.ts
```diff
@@ -8,7 +8,7 @@ function createToken(authorities: string[]) {
     user_name: 'USER1',
     scope: ['read', 'write'],
     auth_source: 'nomis',
-    ...authorities,
+    authorities,
```
tomsmyers, line 11: this is a behaviour change i think ? but the previous code looks broken ?

server/data/hmppsAuthClient.test.ts
tomsmyers: 👍 thanks for the explanation!

tomsmyers (dismissed):
awesome

server/middleware/authorisationMiddleware.test.ts
lawrence-forooghian: When added types to the params `createResWithToken`, I also changed the structure of the arguments, so the `authorities` param went from being the passed-in object to being the `authorities` property value of the passed-in object. So the corresponding change was needed here.

lawrence-forooghian:
Closing this because Circle is refusing to build it, will reopen

PR #23, opened 2020-12-02
Fetch real API token in interventionsService

lawrence-forooghian:

When making calls to the interventions service, it now fetches an API client token from HMPPS auth service, and sends that token in the request.

We should have done this in 5607f25 and 0863f08, but we were just focused on trying to get Pact working and didn’t really think about tokens. The pattern is copied from `communityApiService`.

To allow us to authenticate with the real interventions service.

### DIFF
--- a/server/services/interventionsService.ts
+++ b/server/services/interventionsService.ts
@@ -1,28 +1,36 @@
 import RestClient from '../data/restClient'
 import logger from '../../log'
 import { ApiConfig } from '../config'
+import HmppsAuthClient from '../data/hmppsAuthClient'
 
 export interface Referral {
   id: string
 }
 
 export default class InterventionsService {
-  constructor(private readonly config: ApiConfig) {}
+  constructor(private readonly config: ApiConfig, private readonly hmppsAuthClient: HmppsAuthClient) {}
 
-  private restClient(token: string): RestClient {
+  private async createRestClient(): Promise<RestClient> {
+    const token = await this.hmppsAuthClient.getApiClientToken()
+
     return new RestClient('Interventions Service API Client', this.config, token)
   }
 
   async getReferral(id: string): Promise<Referral> {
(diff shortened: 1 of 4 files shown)


server/services/interventionsService.ts
```diff
@@ -1,28 +1,36 @@
...
 export interface Referral {
   id: string
 }
 
 export default class InterventionsService {
-  constructor(private readonly config: ApiConfig) {}
+  constructor(private readonly config: ApiConfig, private readonly hmppsAuthClient: HmppsAuthClient) {}
+
+  private async createRestClient(): Promise<RestClient> {
+    const token = await this.hmppsAuthClient.getApiClientToken()
```
tomsmyers, line 14: i like that this makes calling this method simpler and less verbose, but are there any implications we can think of of having the token retrieved here vs by the caller?

the only one i can think of is if the client is re-used over a period of time and the token expires, there is no way to refresh it. but i think it's unlikely to happen.

lawrence-forooghian: Ah, you mean like if we have a method which made multiple API calls in sequence, all using the same client, there's a danger that the token would expire after the first API call or something?

If so, it sounds possible, yeah. Reckon I should request a token each time I make an API call?

tomsmyers: thinking about it more you have the same issue if you pass the token into the constructor. the only way round it is to get a fresh token for every request, making it a parameter in the client's get/post etc methods. let's leave it for now and address in a different change?

tomsmyers (dismissed):

lawrence-forooghian:
Hmm, the tests for this are hanging on CI. I'll look into it, might be something I did.

lawrence-forooghian:
Fixed now 👍

tomsmyers (approved):

PR #29, opened 2020-12-03
update the referrals URLs to match the implementation in the API

tomsmyers:

- changes the URLs in the interventions service to match the URL structure discussed here https://mojdt.slack.com/archives/C01DYKJUKDX/p1606925013029800
- makes the id parameter more representative of the real IDs (in fact, passing plain integers to these endpoints will result in 400 responses - is this something we want/need to test in the UI layer?)

ensure the api calls succeed on the real interventions UI when it's deployed.

### DIFF
--- a/server/services/interventionsService.test.ts
+++ b/server/services/interventionsService.test.ts
@@ -16,17 +16,17 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
     beforeEach(async () => {
       // This is just an example for getting Pact set up; we’ll properly design this endpoint later
       await provider.addInteraction({
-        state: 'There is an existing referral with ID of 1',
+        state: 'There is an existing referral with ID of ac386c25-52c8-41fa-9213-fcf42e24b0b5',
         uponReceiving: 'a request for that referral',
         withRequest: {
           method: 'GET',
-          path: '/referrals/1',
+          path: '/draft-referral/ac386c25-52c8-41fa-9213-fcf42e24b0b5',
           headers: { Accept: 'application/json' },
         },
         willRespondWith: {
           status: 200,
           body: Matchers.like({
-            id: '1',
+            id: 'ac386c25-52c8-41fa-9213-fcf42e24b0b5',
           }),
           headers: { 'Content-Type': 'application/json' },
         },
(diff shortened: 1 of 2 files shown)


server/services/interventionsService.test.ts
```diff
@@ -16,26 +16,27 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
     beforeEach(async () => {
       // This is just an example for getting Pact set up; we’ll properly design this endpoint later
       await provider.addInteraction({
-        state: 'There is an existing referral with ID of 1',
+        state: 'There is an existing referral with known id',
```
lawrence-forooghian, line 19:
```suggestion
        state: 'There is an existing referral with ID of ac386c25-52c8-41fa-9213-fcf42e24b0b5`,
```
won't it be useful for you to know the exact ID that needs to be available when you need to set up this state in the backend Pact tests?

server/services/interventionsService.test.ts
```diff
@@ -16,26 +16,27 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
...
-          path: '/referrals/1',
+          path: '/draft-referral/ac386c25-52c8-41fa-9213-fcf42e24b0b5',
           headers: { Accept: 'application/json' },
         },
         willRespondWith: {
           status: 200,
           body: Matchers.like({
-            id: '1',
+            id: 'ac386c25-52c8-41fa-9213-fcf42e24b0b5',
+            created: '2020-12-02 17:20:14',
```
lawrence-forooghian, line 30: The addition of `created` doesn't seem related to the rest of the changes here, should it be here? Also what format is this datetime? It's not a valid ISO 8601 value.

lawrence-forooghian (changes requested):
As mentioned in our commits RFC please could you make sure that the information which is in the pull request message (especially the motivaition) is also in the commit messages?

server/services/interventionsService.test.ts
tomsmyers: will remove and add in a seperate PR. good point on ISO formatting. it's the default serialization of java `Date`. i sort of figured it would be ISO formatting and did not notice it wasn't.

lawrence-forooghian:
@tomsmyers is this ready for re-review?

lawrence-forooghian (approved):

PR #33, opened 2020-12-03
Fix ESLint issue which forbade declaring enums

lawrence-forooghian:

Replaces a piece of ESLint config which prevented us from declaring TypeScript enums. See commit message.

To allow us to declare TypeScript enums, which seem lovely.


tomsmyers (approved):
"Google it and blindly implement the suggested fix kind of fix." - my favorite kind

PR #34, opened 2020-12-04
Add task list to referral form

lawrence-forooghian:

Populates the referral form with a task list which currently shows a single fixed state - the user has half filled the referral form, and the next thing for them to do is to fill in the desired completion date for the accommodation intervention.

It proposes some architecture that will allow us to (hopefully) easily turn this into a dynamic form as we start to work with an actual referral. See commit message for more details.

To provide a starting point for building out the rest of the referral form UI, and to propose an architecture for future UI work.

## Screenshot

Screenshot_2020-12-04 HMPPS Interventions - GOV UK


tomsmyers (approved):
lgtm

PR #37, opened 2020-12-04
IC-641: Upload pact files to the pact broker after tests

sldblog:

Upload pact files to the pact broker after tests.

The [documentation](https://docs.pact.io/pact_broker/tags/) recommends that

> Tag with the branch name when you publish pacts or verification results, and tag with the environment name when you deploy.

This PR implements for the former.

image

So that we can validate them at a later stage (from the provider side).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,6 +17,9 @@ executors:
 
 jobs:
   build:
+    environment:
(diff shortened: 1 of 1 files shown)


tomsmyers (approved):
nice!

sldblog:
🙇‍♂️ 🎩

PR #40, opened 2020-12-07
Add interventions service API call to set / update completion deadline

lawrence-forooghian:

Proposes the API contract for how we would update a draft referral. It adds a `completionDeadline` property and shows how we'd update this from the frontend. I'm seeking feedback on whether a `PATCH` or a `PUT` is more appropriate - see commit message for thinking.

To provide us with the API call that we need in order to allow a user to fill out the referral form.

### DIFF
--- a/server/data/restClient.ts
+++ b/server/data/restClient.ts
@@ -98,6 +98,33 @@ export default class RestClient {
     }
   }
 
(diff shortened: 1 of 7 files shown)


tomsmyers (approved):
lgtm

PR #42, opened 2020-12-09
IC-614: User can set the completion deadline on a draft referral

lawrence-forooghian:

Adds a page `/draft-referrals/:id/completion-deadline` which allows the user to set the completion deadline on a draft referral. It's the first time that we've built a UI to allow the user to set / update data on the backend, so we’re trying out some patterns to see what works. See commit messages for more details.

To allow the user to set a completion deadline on a referral.

## Screenshots

Screenshot_2020-12-10 HMPPS Interventions - GOV UK
Screenshot_2020-12-10 HMPPS Interventions - GOV UK(1)

### DIFF
--- a/server/routes/index.ts
+++ b/server/routes/index.ts
@@ -31,6 +31,8 @@ export default function routes(router: Router, services: Services): Router {
   get('/referrals/start', (req, res) => referralsController.startReferral(req, res))
   post('/referrals', (req, res) => referralsController.createReferral(req, res))
   get('/referrals/:id/form', (req, res) => referralsController.viewReferralForm(req, res))
+  get('/referrals/:id/completion-deadline', (req, res) => referralsController.viewCompletionDeadline(req, res))
+  post('/referrals/:id/completion-deadline', (req, res) => referralsController.updateCompletionDeadline(req, res))
 
   return router
 }
--- a/server/routes/referrals/completionDeadlineForm.test.ts
+++ b/server/routes/referrals/completionDeadlineForm.test.ts
@@ -1,0 +1,144 @@
+import { Request } from 'express'
+import CompletionDeadlineForm from './completionDeadlineForm'
+
+describe('CompletionDeadlineForm', () => {
+  describe('createForm', () => {
+    it('returns a form and sanitizes the request body', async () => {
+      const req = {
+        body: {
+          'completion-deadline-year': '2021',
+          'completion-deadline-month': '09',
+          'completion-deadline-day': '12',
+        },
+      } as Request
+
+      await CompletionDeadlineForm.createForm(req)
+
+      expect(req.body['completion-deadline-year']).toBe(2021)
+      expect(req.body['completion-deadline-month']).toBe(9)
+      expect(req.body['completion-deadline-day']).toBe(12)
+    })
+
+    describe('with invalid data', () => {
+      it('does not modify the invalid fields', async () => {
+        const req = {
+          body: {
+            'completion-deadline-year': '2021',
+            'completion-deadline-month': '09',
+            'completion-deadline-day': 'hello',
+          },
+        } as Request
+
+        await CompletionDeadlineForm.createForm(req)
+
+        expect(req.body['completion-deadline-year']).toBe(2021)
+        expect(req.body['completion-deadline-month']).toBe(9)
+        expect(req.body['completion-deadline-day']).toBe('hello')
+      })
+    })
+  })
+
+  describe('errors', () => {
+    it('returns null with valid data', async () => {
+      const req = {
+        body: {
+          'completion-deadline-year': '2021',
+          'completion-deadline-month': '09',
+          'completion-deadline-day': '12',
+        },
+      } as Request
+
+      const form = await CompletionDeadlineForm.createForm(req)
+
+      expect(form.errors).toBeNull()
+    })
+
+    it('returns an error when a field is empty', async () => {
+      const req = {
+        body: {
+          'completion-deadline-year': '',
+          'completion-deadline-month': '09',
+          'completion-deadline-day': '12',
+        },
+      } as Request
+
+      const form = await CompletionDeadlineForm.createForm(req)
+
+      expect(form.errors).toEqual({
+        firstErroredField: 'day',
+        erroredFields: ['day', 'month', 'year'],
+        message: 'The date by which the service needs to be completed must be a real date',
+      })
+    })
+
+    it('returns an error when a field is just whitespace', async () => {
+      const req = {
+        body: {
+          'completion-deadline-year': '2021',
+          'completion-deadline-month': '     ',
+          'completion-deadline-day': '12',
+        },
+      } as Request
+
+      const form = await CompletionDeadlineForm.createForm(req)
+
+      expect(form.errors).toEqual({
+        firstErroredField: 'day',
+        erroredFields: ['day', 'month', 'year'],
+        message: 'The date by which the service needs to be completed must be a real date',
+      })
+    })
+
+    it('returns an error when a field is non-numeric', async () => {
+      const req = {
+        body: {
+          'completion-deadline-year': '2021',
+          'completion-deadline-month': '09',
+          'completion-deadline-day': 'hello',
+        },
+      } as Request
+
+      const form = await CompletionDeadlineForm.createForm(req)
+
+      expect(form.errors).toEqual({
+        firstErroredField: 'day',
+        erroredFields: ['day', 'month', 'year'],
+        message: 'The date by which the service needs to be completed must be a real date',
+      })
+    })
+
+    it('returns an error when the date specified does not exist', async () => {
+      const req = {
+        body: {
+          'completion-deadline-year': '2011',
+          'completion-deadline-month': '02',
+          'completion-deadline-day': '31',
+        },
+      } as Request
+
+      const form = await CompletionDeadlineForm.createForm(req)
+
+      expect(form.errors).toEqual({
+        firstErroredField: 'day',
+        erroredFields: ['day', 'month', 'year'],
+        message: 'The date by which the service needs to be completed must be a real date',
+      })
+    })
+  })
+
+  describe('paramsForUpdate', () => {
+    it('returns an object with the completionDeadline key and an ISO-formatted date', async () => {
+      const req = {
+        body: {
+          'completion-deadline-year': '2021',
+          'completion-deadline-month': '09',
+          'completion-deadline-day': '12',
+        },
+      } as Request
+
+      const form = await CompletionDeadlineForm.createForm(req)
+
+      expect(form.paramsForUpdate).toEqual({ completionDeadline: '2021-09-12' })
+    })
+  })
+})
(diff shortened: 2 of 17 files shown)


server/routes/index.ts
```diff
@@ -31,6 +31,8 @@ export default function routes(router: Router, services: Services): Router {
   get('/referrals/start', (req, res) => referralsController.startReferral(req, res))
   post('/referrals', (req, res) => referralsController.createReferral(req, res))
   get('/referrals/:id/form', (req, res) => referralsController.viewReferralForm(req, res))
+  get('/referrals/:id/completion-deadline', (req, res) => referralsController.viewCompletionDeadline(req, res))
```
tomsmyers, line 34: i think this pattern is working well, do you?

server/routes/referrals/completionDeadlineForm.test.ts
```diff
@@ -0,0 +1,126 @@
...
+
+      await CompletionDeadlineForm.createForm(req)
+
+      expect(req.body['completion-deadline-year']).toBe(2021)
+      expect(req.body['completion-deadline-month']).toBe(9)
+      expect(req.body['completion-deadline-day']).toBe(12)
+    })
+
+    describe('with invalid data', () => {
+      it('does not modify the invalid fields', async () => {
```
tomsmyers, line 23: not a blocker - would be useful to know why (it does not modify invliad fields) for future readers.

tomsmyers: i guess this becomes immediately clear as you read the following test case.

server/routes/referrals/completionDeadlineForm.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  static get validations(): SanitizationChain[] {
+    return [
+      // TODO decide whether we like the fact that express-validator is mutating the request?
+      // e.g. the `bail` before toInt() is so that we don’t end up replaying NaN back to the user
+      // if they enter a blank input
+      body('completion-deadline-day')
+        .notEmpty({ ignore_whitespace: true })
+        .withMessage('The date by which the service needs to be completed must include a day')
```
tomsmyers, line 26: this is good, but i wonder if as we add more and more validation messages it would be useful to contain them in a single file, indexed by description `{ "dateFieldEmtpy": "The date by which the service needs to be completed must include a day" }`. this way we could easily have designers audit the messages to ensure they are user friendly enough.

server/routes/referrals/completionDeadlineView.ts
```diff
@@ -0,0 +1,69 @@
...
+    const errorMessage = this.presenter.errorMessage ? { text: this.presenter.errorMessage } : null
+
+    return {
+      id: 'completion-deadline',
+      namePrefix: 'completion-deadline',
+      fieldset: {
+        legend: {
+          text: this.presenter.title,
+          isPageHeading: true,
+          classes: 'govuk-fieldset__legend--xl',
```
tomsmyers, line 16: what's the reason for having css references in here?

server/routes/referrals/referralsController.test.ts
```diff
@@ -91,3 +95,87 @@ describe('GET /referrals/:id/form', () => {
...
+
+  it('renders a form page', async () => {
+    await request(app)
+      .get('/referrals/1/completion-deadline')
+      .expect(200)
+      .expect(res => {
+        expect(res.text).toContain('What date does the accommodation service need to be completed by?')
+      })
+  })
+  // TODO how do we (or indeed, do we) test what happens when the request has a completion deadline - i.e. that the
```
tomsmyers, line 116: can you override the mock for this test and add a completionDealine to it? then check if the date field is populated?

server/utils/calendarDay.ts
```diff
@@ -0,0 +1,21 @@
+export default class CalendarDay {
+  constructor(readonly day: number, readonly month: number, readonly year: number) {}
+
+  static parseIso8601(val: string): CalendarDay | null {
+    const result = /^(\d+)-(\d+)-(\d+)$/.exec(val)
```
tomsmyers, line 5: this parses random dash seperated numbers like `1234-8765123-129873`. is that ok? why not specficy length of 4-2-2? (or 2 to 4, 1 to 2, 1 to 2).

`^(\d{4})-(\d{2})-(\d{2})$` (or `^(\d{2-4})-(\d{1-2})-(\d{1-2})$`)

tomsmyers: in fact, why not use Date.parse(), then we get validation of incorrect dates like 31st february?

const date = new Date(Date.parse(val))
const year = date.getFullYear()
...

tomsmyers (commented):
a few comments - nothing really a blocker. the only thing i think deserved further discussion now is the validation of correctly formatted, but impossible dates e.g. 31st february.

server/routes/index.ts
lawrence-forooghian: Yep, seems to be.

server/routes/referrals/referralsController.test.ts
lawrence-forooghian: The problem isn't getting the data into the view, it's that the testing tools are too limited to be able to check that the fields are populated correctly - all we can really do is a text match on the raw HTML body, but to check the fields we'd need something that parses the HTML and allows us to interrogate its structure. We could introduce something like that into the tests but we wondered whether it's better to leave it for the Cypress tests?

server/routes/referrals/completionDeadlineView.ts
lawrence-forooghian: This class is a "view", so anything to do with rendering might appear here, including which CSS classes to use on a title.

server/routes/referrals/referralsController.test.ts
tomsmyers: fine with me. sounds overly complex to be parsing the html here.

server/utils/calendarDay.ts
lawrence-forooghian: > is that ok? why not specficy length of 4-2-2

I was being pretty liberal with the parsing - all that I really cared about was being able to split the date into its components so that I can put them into text fields. But I've changed it to use `Date.parse`, not sure whether I prefer it though 😬

server/routes/referrals/completionDeadlineForm.ts
lawrence-forooghian: Good shout - added `server/utils/errorMessages.ts`.

sldblog:
A random thought: this change introduces a new expectation from the service. The service currently tests against the latest consumer pacts tagged as `main`.

So once this PR is merged, in theory, the service unit tests start failing.

Would it be possible to add a version number in front of the git sha so we can say in the provider: "we only test against consumers tagged main on version 1"?

lawrence-forooghian:
@sldblog I think [this Pact documentation](https://docs.pact.io/pact_nirvana/step_5) ("Allow contracts to change without breaking your builds") proposes workflow for addressing the problem you describe. It involves using tags on the contracts generated frontend to represent new "features". The idea then being that the frontend feature branch wouldn't be merged until the backend was passing. I don't really like this because long-lived feature branches are unpleasant to work with.

The alternative would be what you suggest - using tags to represent the provider "version" that we're currently describing. I guess the thing I'm wondering is _when_ we'd bump this version on the frontend. I can see it leading to lots of noisy "bump backend version" PRs and commits?

sldblog:
I guess we can use a manual, semantic tagging approach?

Currently, we use `--tag $CIRCLE_BRANCH`.

According to the docs, we can specify many tags:
```
  -t, [--tag=TAG]                                          # Tag name for consumer version. Can be specified multiple times.
```

So we should be able to say
```diff
             npx pact-broker publish pact/pacts \
               --broker-base-url $PACT_BROKER_BASE_URL \
-              --consumer-app-version $CIRCLE_SHA1 --tag $CIRCLE_BRANCH \
+              --consumer-app-version $CIRCLE_SHA1 --tag $CIRCLE_BRANCH --tag "provider-${provider_version} \
               -u $PACT_BROKER_USERNAME -p $PACT_BROKER_PASSWORD
```

And then filter on the provider side to say "all consumer pacts with tags "main" and "provider-v1"?

lawrence-forooghian:
As an alternative, how about we tag each app version with the SHA of the HEAD of the consumer app? Then the backend can be explicit about which version of the consumer app it currently supports.

(I see that we already pass the consumer app version as the `--consumer-app-version` argument, but I think you need it to be provided as a tag to use it as a [selector](https://docs.pact.io/pact_broker/advanced_topics/consumer_version_selectors) on the provider tests.)

sldblog:
👍 That could work, however, I'd probably limit tags of this sort to `main` only to avoid noise.

I'm wondering which semantic "provider checking against consumer version" or "consumer defining provider version" would be simpler.

Shall we discuss outside this PR?

lawrence-forooghian:
Yep, sounds good.

server/utils/calendarDay.ts
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  static parseIso8601(val: string): CalendarDay | null {
+    const result = /^(\d{4})-(\d{2})-(\d{2})$/.exec(val)
+
+    if (result === null) {
+      return null
+    }
+
+    const year = Number(result[1])
```
tomsmyers, line 15: cosmetic - could be condensed to ```const [year, month, day] = result.map(Number)```? (i think)

server/utils/calendarDay.ts
```diff
@@ -0,0 +1,38 @@
...
+    return [
+      String(this.year).padStart(4, '0'),
+      String(this.month).padStart(2, '0'),
+      String(this.day).padStart(2, '0'),
+    ].join('-')
+  }
+}
+
+function isValidDate(day: number, month: number, year: number) {
+  const date = new Date(Date.UTC(year, month - 1, day))
```
tomsmyers, line 36: could do with comment explaining what's going on here (i assume we are checking for overflows)

server/utils/calendarDay.ts
```diff
@@ -0,0 +1,38 @@
...
+  get iso8601(): string {
+    return [
+      String(this.year).padStart(4, '0'),
+      String(this.month).padStart(2, '0'),
+      String(this.day).padStart(2, '0'),
+    ].join('-')
+  }
+}
+
+function isValidDate(day: number, month: number, year: number) {
```
tomsmyers, line 35: could do with comment explaining what's going on here, for both the +/- 1 and the general idea of checking for overflows.

server/routes/referrals/completionDeadlineForm.test.ts
```diff
@@ -0,0 +1,144 @@
...
+      const form = await CompletionDeadlineForm.createForm(req)
+
+      expect(form.errors).toEqual({
+        firstErroredField: 'day',
+        erroredFields: ['day', 'month', 'year'],
+        message: 'The date by which the service needs to be completed must be a real date',
+      })
+    })
+
+    it('returns an error when the date specified does not exist', async () => {
```
tomsmyers, line 110: 🙏🏼

tomsmyers (approved):
nitpicks only - nice work.

server/utils/calendarDay.ts
lawrence-forooghian: 👍 (needs a `slice` too)

server/utils/calendarDay.ts
lawrence-forooghian: 👍 done

PR #43, opened 2020-12-11
use login tokens for interventions API access

tomsmyers:

instead of using the client_credentials token for the interventions service calls, use the logged in user's token, which provides username details and roles to the downstream service.

 this allows granular role management and user-authenticated HMPPS Auth API calls at the API layer. it also allows the interventions API to associate calls with logged in users.

### DIFF
--- a/server/services/interventionsService.ts
+++ b/server/services/interventionsService.ts
@@ -9,18 +9,16 @@ export interface DraftReferral {
 }
 
 export default class InterventionsService {
-  constructor(private readonly config: ApiConfig, private readonly hmppsAuthClient: HmppsAuthClient) {}
+  constructor(private readonly config: ApiConfig) {}
 
-  private async createRestClient(): Promise<RestClient> {
-    const token = await this.hmppsAuthClient.getApiClientToken()
-
+  private createRestClient(token: string): RestClient {
     return new RestClient('Interventions Service API Client', this.config, token)
   }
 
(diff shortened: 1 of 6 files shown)


sldblog:
Would this allow every user API access to the backend without the UI?

sldblog:
Summarising an "off github" discussion: I misunderstood what this does :)

sldblog (commented):
Summarising an "off github" discussion: I misunderstood what this does :)

LGTM, but would like to defer approval to @Gweaton and @lawrence-forooghian

server/services/interventionsService.ts
```diff
@@ -9,36 +9,34 @@ export interface DraftReferral {
 }
 
 export default class InterventionsService {
-  constructor(private readonly config: ApiConfig, private readonly hmppsAuthClient: HmppsAuthClient) {}
-
-  private async createRestClient(): Promise<RestClient> {
-    const token = await this.hmppsAuthClient.getApiClientToken()
+  constructor(private readonly config: ApiConfig) {}
 
+  private async createRestClient(token: string): Promise<RestClient> {
```
lawrence-forooghian, line 14: I think it's a bit confusing for this to remain `async` - can we remove that?

lawrence-forooghian (changes requested):

lawrence-forooghian (approved):

PR #45, opened 2020-12-15
Use the Fishery package to create factories for test data

lawrence-forooghian:

Integrates the [Fishery](https://github.com/thoughtbot/fishery) package. It allows us to easily create model instances to use in our unit tests. I've changed all of the existing tests that use `DraftReferral` instances, to use the factory.

To avoid having to make changes across the whole test suite every time we add a new property to `DraftReferral`.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -94,7 +94,10 @@
         "functions": "never"
       }
     ],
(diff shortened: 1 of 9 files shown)


tomsmyers (approved):
this is great - thanks

PR #47, opened 2020-12-16
expect 'created' field in draft referral reponses

tomsmyers:

adds a 'created' field to the draft referral object and expected draft referral responses.

this will allow us to order lists of draft referrals by when they were started. these changes coming soon.

### DIFF
--- a/server/services/interventionsService.ts
+++ b/server/services/interventionsService.ts
@@ -4,6 +4,7 @@ import { ApiConfig } from '../config'
 
 export interface DraftReferral {
   id: string
+  createdAt: string
   completionDeadline: string | null
   serviceCategoryId: string | null
   complexityLevelId: string | null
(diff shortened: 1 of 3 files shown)


server/services/interventionsService.ts
```diff
@@ -4,6 +4,7 @@ import { ApiConfig } from '../config'
 
 export interface DraftReferral {
   id: string
+  created: string
```
Gweaton, line 7: Thinking out loud, I wonder if `createdAt` might be a bit clearer than plain "created" (which to me sounds like it could be a boolean value?) Ignore me if you think I'm fussing over nothing!

Gweaton (commented):
Remember to rebase this with `origin/main` as there are some changes to the factory now.

server/services/interventionsService.ts
tomsmyers: yeh fine with me. i'll make the change

Gweaton: Are you leaving it as `created` on the API response? If so, maybe we should just leave it as it was so they at least match up here, although I don't know how important that is.

tomsmyers: this already exists in the response, so would require changes on the APi too. i agree the naming does sounds like a boolean value, but it's not going to bug me too much

Gweaton (approved):

PR #48, opened 2020-12-16
use real oauth2 tokens in pact requests

tomsmyers:

use real oauth2 tokens in pact requests

allow calling backend API endpoints that require token auth in pact tests


Gweaton (approved):
Sorry, 2089 development team...

PR #49, opened 2020-12-16
Turn on TypeScript’s strict mode

lawrence-forooghian:

Turns on TypeScript’s [`strict` mode](https://www.typescriptlang.org/tsconfig#strict), and fixes the resulting build failures along the way.

To get the most out of using a compiled language, allowing the compiler to get rid of some common classes of programmer error and to get the richest type information possible.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -55,7 +55,8 @@
             "semi": false,
             "arrowParens": "avoid"
           }
-        ]
+        ],
+        "@typescript-eslint/no-non-null-assertion": 0
       }
     }
   ],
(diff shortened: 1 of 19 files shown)


lawrence-forooghian:
Hmm, the non-null assertions are triggering an ESLint rule:

```
Forbidden non-null assertion  @typescript-eslint/no-non-null-assertion
```

Might disable that one and trust us to use them sensibly.

lawrence-forooghian:
> Might disable that one and trust us to use them sensibly.

Disabled it.

tomsmyers (approved):
thanks for this - the null checks especially are great to have enabled.

PR #50, opened 2020-12-16
Treat ESLint warnings as errors

lawrence-forooghian:

Fails the linting step if ESLint encounters any warnings. This effectively prevents us from introducing ESLint warnings.

(And I fix the current ESLint warnings in our codebase.)

So that we notice potential code quality problems as they occur. See commit message for more details.


tomsmyers (approved):
🤙🏼

PR #52, opened 2020-12-17
IC-666 Further information for service provider page

Gweaton:

Adds the "further information page" (see screenshot).

There are a couple of potentially outstanding things to think about here:
- We don't have any validations on the client side as this is an optional page, so we're rendering any server errors in the "error summary" component, which might not be quite right.
- Because of the above, I haven't used our normal pattern of creating a `form` component for this, as I thought it was probably overkill, but if things change, it shouldn't be too hard to write one and slot it in.
- I'm not 100% convinced by my own logic for the different conditionals around when there's user data and a user submits an empty form. The tests seem to pass but as we don't have any integration tests or a local environment that allows us to test this properly, I'd like to revisit this (and really, the whole form) once we've got proper back-end integration running locally.

To allow users to add further information to a referral.

## Screenshot

image

### DIFF
--- a/server/routes/referrals/furtherInformationPresenter.test.ts
+++ b/server/routes/referrals/furtherInformationPresenter.test.ts
@@ -1,0 +1,44 @@
+import serviceCategoryFactory from '../../../testutils/factories/serviceCategory'
+import FurtherInformationPresenter from './furtherInformationPresenter'
+
+describe('FurtherInformationPresenter', () => {
+  const serviceCategory = serviceCategoryFactory.build({ name: 'social inclusion' })
+
+  describe('title', () => {
+    it('returns a title', () => {
+      const presenter = new FurtherInformationPresenter(serviceCategory)
+      expect(presenter.title).toEqual(
+        'Do you have further information for the social inclusion service provider? (optional)'
+      )
+    })
+  })
+
+  describe('hint', () => {
+    it('returns a hint', () => {
+      const presenter = new FurtherInformationPresenter(serviceCategory)
+      expect(presenter.hint).toEqual(
+        'For example, relevant previous offences, previously completed programmes or further reasons for this referral'
+      )
+    })
+  })
+
+  describe('error information', () => {
+    describe('when no errors are passed in', () => {
+      it('returns no errors', () => {
+        const presenter = new FurtherInformationPresenter(serviceCategory)
+
+        expect(presenter.error).toBeNull()
+      })
+    })
+
+    describe('when errors are passed in', () => {
+      it('returns error information', () => {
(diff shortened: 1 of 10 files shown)


server/routes/referrals/furtherInformationPresenter.test.ts
```diff
@@ -0,0 +1,44 @@
+import serviceCategoryFactory from '../../../testutils/factories/serviceCategory'
+import FurtherInformationPresenter from './furtherInformationPresenter'
+
+describe('FurtherInformationPresenter', () => {
+  const serviceCategory = serviceCategoryFactory.build({ name: 'social inclusion' })
+
+  describe('title', () => {
```
lawrence-forooghian, line 9: When doing the additional needs and requirements page, I found it a bit tedious writing a test for each of these bits of text, so I just [stuck them into a `text` object and tested that](https://github.com/ministryofjustice/hmpps-interventions-ui/pull/54/files#diff-cac9f0e685a58d20c619a10f762a4f25fa0330682460662f140d1696fbcc5246R43). I'll see what whoever reviews that PR thinks of it but might be a simpler approach.

server/routes/referrals/furtherInformationPresenter.ts
```diff
@@ -1,17 +1,35 @@
...
+
+    if (this.referral.furtherInformation) {
+      return this.referral.furtherInformation
+    }
+
+    return ''
+  }
+
+  private get hasUserInputOrEmptyFormSubmitted(): boolean {
+    return this.userInputData !== null && this.userInputData['further-information'] !== undefined
```
lawrence-forooghian, line 33: Similar comment to [the one](https://github.com/ministryofjustice/hmpps-interventions-ui/pull/51#discussion_r551248907) I left on #51 - I don't think we want to check for the presence of that particular field. All we care about is - did the user submit any data (i.e. are we displaying this in response to a POST request)? If so, then replay whatever data they entered, if anything.

Also, in #52 I [encapsulated a bunch of this logic in a helper class](https://github.com/ministryofjustice/hmpps-interventions-ui/pull/54/files#diff-5f63ac2d1f262adf16ff47612e788fd4c885d691208a1e509bc10a71548c9dd4R66). Once that's merged it would be good to bring this in line. I'm hoping we can simplify our tests then too.

lawrence-forooghian (changes requested):
Looks good, I'll address the comments that I've left here.

server/routes/referrals/furtherInformationPresenter.ts
lawrence-forooghian: Addressed the issue here, and will update to use helper class once #54 merged.

lawrence-forooghian (approved):

PR #54, opened 2020-12-18
IC-666 Add "service user needs and requirements" page

lawrence-forooghian:

Adds a page for specifying additional service user needs and requirements. For example, whether they need an interpreter.

It's our first page:
- with multiple inputs
- with conditional radio buttons

This page was initially quite painful to build due to a lot of repeated presenter logic and tests in particular. I've extracted some things to helper classes to avoid this boilerplate pain in the future. See commit messages.

To build another page of the referral form, and to hopefully make it easier to build future pages.

### Initial page

Screenshot_2020-12-18 HMPPS Interventions - GOV UK

### Validation errors

Screenshot_2020-12-18 HMPPS Interventions - GOV UK(1)

### DIFF
--- a/server/routes/referrals/referralDataPresenterUtils.test.ts
+++ b/server/routes/referrals/referralDataPresenterUtils.test.ts
@@ -1,0 +1,252 @@
+import ReferralDataPresenterUtils from './referralDataPresenterUtils'
+import draftReferralFactory from '../../../testutils/factories/draftReferral'
+
+describe('ReferralDataPresenterUtils', () => {
+  describe('stringValue', () => {
+    describe('when the referral has a null value for the property', () => {
+      describe('when there is no user input data', () => {
+        it('returns an empty string', () => {
+          const referral = draftReferralFactory.build({ additionalNeedsInformation: null })
+          const utils = new ReferralDataPresenterUtils(referral, null)
+
+          expect(utils.stringValue('additionalNeedsInformation', 'additional-needs-information')).toBe('')
+        })
+      })
+
+      describe('when there is user input data and the user input value for the property is null', () => {
+        it('returns an empty string', () => {
+          const referral = draftReferralFactory.build({ additionalNeedsInformation: null })
+          const userInputData = { 'additional-needs-information': null }
+          const utils = new ReferralDataPresenterUtils(referral, userInputData)
+
+          expect(utils.stringValue('additionalNeedsInformation', 'additional-needs-information')).toBe('')
+        })
+      })
+
+      describe('when there is user input data and the property is absent from the user input data', () => {
+        it('returns an empty string', () => {
+          const referral = draftReferralFactory.build({ additionalNeedsInformation: null })
+          const userInputData = {}
+          const utils = new ReferralDataPresenterUtils(referral, userInputData)
+
+          expect(utils.stringValue('additionalNeedsInformation', 'additional-needs-information')).toBe('')
+        })
+      })
+
+      describe('when there is user input data and the user input value for the property is an empty string', () => {
+        it('returns an empty string', () => {
+          const referral = draftReferralFactory.build({ additionalNeedsInformation: null })
+          const userInputData = { 'additional-needs-information': '' }
+          const utils = new ReferralDataPresenterUtils(referral, userInputData)
+
+          expect(utils.stringValue('additionalNeedsInformation', 'additional-needs-information')).toBe('')
+        })
+      })
+
+      describe('when there is user input data and the user input value for the property is a non-empty string', () => {
+        it('returns the user input value', () => {
+          const referral = draftReferralFactory.build({ additionalNeedsInformation: null })
+          const userInputData = { 'additional-needs-information': 'bar' }
+          const utils = new ReferralDataPresenterUtils(referral, userInputData)
+
+          expect(utils.stringValue('additionalNeedsInformation', 'additional-needs-information')).toBe('bar')
+        })
+      })
+
+      describe('when there is user input data and the user input value for the property is present but neither a string nor null', () => {
+        it('returns the stringified user input value', () => {
+          const referral = draftReferralFactory.build({ additionalNeedsInformation: null })
+          const userInputData = { 'additional-needs-information': 100 }
+          const utils = new ReferralDataPresenterUtils(referral, userInputData)
+
+          expect(utils.stringValue('additionalNeedsInformation', 'additional-needs-information')).toBe('100')
+        })
+      })
+    })
+
+    describe('when the referral has a non-null value for the property', () => {
+      describe('when there is no user input data', () => {
+        it('returns the value from the referral', () => {
+          const referral = draftReferralFactory.build({ additionalNeedsInformation: 'foo' })
+          const utils = new ReferralDataPresenterUtils(referral, null)
+
+          expect(utils.stringValue('additionalNeedsInformation', 'additional-needs-information')).toBe('foo')
+        })
+      })
+
+      describe('when there is user input data and the user input value for the property is null', () => {
+        it('returns an empty string', () => {
+          const referral = draftReferralFactory.build({ additionalNeedsInformation: 'foo' })
+          const userInputData = { 'additional-needs-information': null }
+          const utils = new ReferralDataPresenterUtils(referral, userInputData)
+
+          expect(utils.stringValue('additionalNeedsInformation', 'additional-needs-information')).toBe('')
+        })
+      })
+
+      describe('when there is user input data and the property is absent from the user input data', () => {
+        it('returns an empty string', () => {
+          const referral = draftReferralFactory.build({ additionalNeedsInformation: 'foo' })
+          const userInputData = {}
+          const utils = new ReferralDataPresenterUtils(referral, userInputData)
+
+          expect(utils.stringValue('additionalNeedsInformation', 'additional-needs-information')).toBe('')
+        })
+      })
+
+      describe('when there is user input data and the user input value for the property is an empty string', () => {
+        it('returns an empty string', () => {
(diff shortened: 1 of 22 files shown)


server/services/interventionsService.test.ts
```diff
@@ -279,6 +279,189 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
...
+    it('returns the updated referral when setting additionalNeedsInformation and accessibilityNeeds', async () => {
+      await provider.addInteraction({
+        state: 'a draft referral with ID dfb64747-f658-40e0-a827-87b4b0bdcfed exists',
+        uponReceiving: 'a PATCH request to set additionalNeedsInformation and accessibilityNeeds',
+        withRequest: {
+          method: 'PATCH',
+          path: '/draft-referral/dfb64747-f658-40e0-a827-87b4b0bdcfed',
+          headers: {
+            Accept: 'application/json',
+            'Content-Type': 'application/json',
```
tomsmyers, line 324: total nit pick - but why is this field in quotes but the others are not?

server/services/interventionsService.test.ts
```diff
@@ -279,6 +279,189 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
...
+          },
+          body: {
+            accessibilityNeeds: 'She uses a wheelchair',
+            additionalNeedsInformation: 'Alex is currently sleeping on her aunt’s sofa',
+          },
+        },
+        willRespondWith: {
+          status: 200,
+          body: {
+            id: Matchers.like('dfb64747-f658-40e0-a827-87b4b0bdcfed'),
```
tomsmyers, line 335: i think this response would contain all the other fields on the referral, whether null or otherwise, like 

```{
    "id": "97385a77-0cec-4ee0-a2a5-dcad0ecf7101",
    "created": "2021-01-04T13:08:59.329395Z",
    "completionDeadline": null,
    "serviceCategoryId": null,
    "complexityLevelId": null,
    etc....
}```

File: server/services/interventionsService.test.ts [hunk 3]
```diff
@@ -279,6 +279,189 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
...
+          headers: {
+            Accept: 'application/json',
+            'Content-Type': 'application/json',
+            Authorization: `Bearer ${token}`,
+          },
+          body: { needsInterpreter: true, interpreterLanguage: 'Spanish' },
+        },
+        willRespondWith: {
+          status: 200,
+          body: {
```
tomsmyers, line 370: same for this one - i expect the response from patch to be the updated referral containing all fields, not just the ones set in this request.

server/services/interventionsService.test.ts
```diff
@@ -106,6 +106,38 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
...
+          withRequest: {
+            method: 'GET',
+            path: '/draft-referral/1219a064-709b-4b6c-a11e-10b8cb3966f6',
+            headers: { Accept: 'application/json', Authorization: 'Bearer token' },
+          },
+          willRespondWith: {
+            status: 200,
+            body: Matchers.like({
+              id: '1219a064-709b-4b6c-a11e-10b8cb3966f6',
+              serviceUser: {
```
tomsmyers, line 125: do we want this in the referral - or as an id referancing another entity as is the case with complexity level, service category etc?

server/services/interventionsService.test.ts
lawrence-forooghian: I reckon it's an ESLint or Prettier thing - only quotes them if they need to be quoted (keys with special characters need to be quoted in JS I believe)

server/utils/formUtils.ts
```diff
@@ -0,0 +1,32 @@
+import { body, Result, ValidationChain, ValidationError, validationResult } from 'express-validator'
+import { Request } from 'express'
+
+export default class FormUtils {
+  static yesNoRadioWithConditionalInputValidationChain({
```
tomsmyers, line 5: this name is long and quite hard to read but is probably as good as can be isn't it

server/utils/formUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+    inputName,
+    notSelectedErrorMessage,
+    inputValidator,
+  }: {
+    radioName: string
+    inputName: string
+    notSelectedErrorMessage: string
+    inputValidator: (chain: ValidationChain) => ValidationChain
+  }): ValidationChain[] {
+    return [
```
tomsmyers, line 16: nice

server/routes/referrals/referralDataPresenterUtils.test.ts
```diff
@@ -0,0 +1,252 @@
...
+          const utils = new ReferralDataPresenterUtils(referral, userInputData)
+
+          expect(utils.stringValue('additionalNeedsInformation', 'additional-needs-information')).toBe('')
+        })
+      })
+
+      describe('when there is user input data and the property is absent from the user input data', () => {
+        it('returns an empty string', () => {
+          const referral = draftReferralFactory.build({ additionalNeedsInformation: 'foo' })
+          const userInputData = {}
```
tomsmyers, line 90: this is the first case that made me double check. what's the rationale that this empty object overwrites the field?

server/routes/referrals/referralDataPresenterUtils.test.ts
```diff
@@ -0,0 +1,252 @@
...
+          const utils = new ReferralDataPresenterUtils(referral, userInputData)
+
+          expect(utils.booleanValue('needsInterpreter', 'needs-interpreter')).toBe(true)
+        })
+      })
+
+      describe("when there is user input data and the user input value for the property is present but neither 'yes', 'no' nor null", () => {
+        it('returns null', () => {
+          const referral = draftReferralFactory.build({ needsInterpreter: null })
+          const userInputData = { 'needs-interpreter': true }
```
tomsmyers, line 183: is this also the case for 'true' and 'false'?

tomsmyers: reason for asking is that i would naively expect 'truthy' strings to resolve properly. in the past i've written little helpers like `{ return ['true', 'on', 'yes', '1'].includes(val) }`.

server/routes/referrals/referralDataPresenterUtils.test.ts
```diff
@@ -0,0 +1,252 @@
...
+          const utils = new ReferralDataPresenterUtils(referral, userInputData)
+
+          expect(utils.booleanValue('needsInterpreter', 'needs-interpreter')).toBeNull()
+        })
+      })
+
+      describe('when there is user input data and the property is absent from the user input data', () => {
+        it('returns null', () => {
+          const referral = draftReferralFactory.build({ needsInterpreter: false })
+          const userInputData = {}
```
tomsmyers, line 214: same again, i'm slightly confused by this one.

server/services/interventionsService.test.ts
lawrence-forooghian: You're right, I think it would. But this approach (of only testing for the updated fields) is consistent with the other `PATCH` contract tests in this file. If the response contained additional fields, then the contract tests would still pass. I think that what we're testing for with these contract tests is that the response to a `PATCH` request correctly reflects the data which was submitted in this request.

There's a broader question which is whether we should be asserting anything about these `PATCH` response bodies at all, given that the UI doesn't actually make use of them at the moment.

server/services/interventionsService.test.ts
lawrence-forooghian: I don't know. I just did the simplest thing possible that would get me a first name I could work with. I guess I thought of the service user data as just being part of the "referral form", which is the data contained in the base referral object, but that's a bit of a flimsy reasoning since complexity level and service category are part of the referral form too.

Do you think we should try and figure this out in this PR or leave it until we come to implement service user details properly? I'm conscious that by merging this PR then we'd add a burden for it to be implemented on the backend, but at the same time I'm keen for this frontend functionality not to get blocked by a conversation about API design.

server/utils/formUtils.ts
lawrence-forooghian: I don't have a better name for it, but perhaps could stick it in a `validations` object so it would be accessed as `FormUtils.validations.yesNoRadioWithConditionalInput`?

server/routes/referrals/referralDataPresenterUtils.test.ts
lawrence-forooghian: The idea is that if the presenter is displaying the page in response to the user submitting a form (which would usually mean that there's been a validation failure and the user needs to correct some of the data they entered), then instead of presenting the data from the model, it should present the data which the user submitted in the form.

The specific case you're asking about - the user submitted a form but that particular field is absent in the request body - could occur for example with checkboxes, where if the user unticks all of the checkboxes for a field `foo` then the browser will not submit a `foo` key in the request body. We still want to display an empty list of checkboxes in that case. For the example here (a text value) it's a bit of an edge case and would only happen in a request from a browser in the case of e.g. a user managing to uncheck an already-selected radio button. A bit pathological but I thought worth sticking a test in for it.

server/routes/referrals/referralDataPresenterUtils.test.ts
lawrence-forooghian: Yeah, I think same answer as my previous one.

tomsmyers (approved):

server/services/interventionsService.test.ts
tomsmyers: makes sense. we can leave like this for now and see how we get on. 👍🏼

PR #57, opened 2021-01-04
small tweak to ISO timestamps to ensure pacts are satisfied

tomsmyers:

changes the format of the ISO timestamp returned in the api response

allow backend pact tests to pass


karenmillermoj (approved):
These changes are fine.  I do however prefer the original format, I don't see a need for millis.

tomsmyers:
> These changes are fine. I do however prefer the original format, I don't see a need for millis.

i actually agree - but the effort involved in changing this in spring was too great for me to bother. there's this whole thing about different config for different date types and it was just sucking too much time. as long as the timestamps are iso8601 formatted i'm happy. @lawrence-forooghian ?

lawrence-forooghian:
Not a problem at least at the moment - we don't use that field in the app anyway, I believe.

lawrence-forooghian (approved):

PR #59, opened 2021-01-05
Add service user risk information form page

lawrence-forooghian:

Adds a page for entering additional information about the service user’s risk.

Note that the summary list of risk information is currently just showing static data.

To continue building the referral form.

## Screenshot

Screenshot_2021-01-05 HMPPS Interventions - GOV UK

### DIFF
--- a/server/routes/referrals/referralDataPresenterUtils.ts
+++ b/server/routes/referrals/referralDataPresenterUtils.ts
@@ -54,4 +54,8 @@ export default class ReferralDataPresenterUtils {
       return aIndex - bIndex
     })
   }
+
+  static errorMessage(errors: { field: string; message: string }[] | null, field: string): string | null {
+    return errors?.find(error => error.field === field)?.message ?? null
+  }
 }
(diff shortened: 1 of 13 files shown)


server/routes/referrals/referralDataPresenterUtils.ts
```diff
@@ -54,4 +54,8 @@ export default class ReferralDataPresenterUtils {
       return aIndex - bIndex
     })
   }
+
+  static errorMessage(errors: { field: string; message: string }[] | null, field: string): string | null {
+    return errors?.find(error => error.field === field)?.message ?? null
```
tomsmyers, line 59: does this assume that the list of errors has a top-level 'field' parameter?

lawrence-forooghian: Yup - that's what this bit of the method signature states:

```
errors: { field: string; message: string }[] | null
```

tomsmyers: right- i was confused that this object was the api response, but i see it's an object we create from the api response.

lawrence-forooghian: Yeah, it'd be either an object resulting from a frontend validation error, _or_ something we create from an API error response.

tomsmyers (approved):
lgtm

PR #60, opened 2021-01-05
add serviceCategoryId to newly created draft referrals to enable better testing in dev environment

tomsmyers:

sets some static data on newly created draft referrals.

currently we will not have a way to specify the service category or service user crn until the discover/find work has commenced. this change means every newly created referral has a service category selected which allows us to use the referral form to fill in the rest of the detials.

### DIFF
--- a/server/services/interventionsService.test.ts
+++ b/server/services/interventionsService.test.ts
@@ -141,7 +141,7 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
   })
 
   describe('createDraftReferral', () => {
-    beforeEach(async () => {
+    it('returns a newly created draft referral', async () => {
       await provider.addInteraction({
         // We're not sure what is an appropriate state here
         state: 'a draft referral can be created',
@@ -165,9 +165,7 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
           },
         },
       })
-    })
 
-    it('returns a referral', async () => {
       const referral = await interventionsService.createDraftReferral(token)
       expect(referral.id).toBe('dfb64747-f658-40e0-a827-87b4b0bdcfed')
     })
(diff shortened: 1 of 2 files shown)


Gweaton (approved):

server/services/interventionsService.ts
```diff
@@ -60,13 +60,17 @@ export default class InterventionsService {
...
+  async createDraftReferral(token: string, initialFields?: Partial<DraftReferral>): Promise<DraftReferral> {
     const restClient = await this.createRestClient(token)
 
-    return (await restClient.post({
+    const referral = (await restClient.post({
       path: `/draft-referral`,
       headers: { Accept: 'application/json' },
     })) as DraftReferral
+
+    return initialFields === undefined || initialFields === {}
```
lawrence-forooghian, line 71: I reckon that if is just a temporary thing then we can just stick this in the controller action - I don't think we need this functionality in the service, and if we do then we should probably test it.

server/services/interventionsService.test.ts
```diff
@@ -527,6 +525,40 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
       expect(referral.id).toBe('dfb64747-f658-40e0-a827-87b4b0bdcfed')
       expect(referral.additionalRiskInformation).toEqual('A danger to the elderly')
     })
+
+    it('returns the updated referral when setting the service category ID', async () => {
```
lawrence-forooghian, line 529: Should we have one that tests we can set the first name, too?

lawrence-forooghian (changes requested):

server/services/interventionsService.ts
tomsmyers: i think the content is temporary, but i can see this being used in the future when we are starting the referral journey with a known crn and service etc.

what do you think ?

i agree on the testing part though. i'll add something.

server/services/interventionsService.test.ts
tomsmyers: isn't there already a test for updating SU info?

lawrence-forooghian: Nope - we're just able to receive it, not set it. I just wanted the property there on the model so that I could use it in presenters.

server/services/interventionsService.ts
lawrence-forooghian: I suppose, but I kinda find it cleaner to stick with the current model of one `InterventionsService` method = one endpoint. Cos Pact tests are designed for testing a single request / response pair so we could no longer test this method with a Pact test. And it would be nice if we could keep all the tests for this class as Pact tests.

server/services/interventionsService.test.ts
lawrence-forooghian: (I've just been cheekily setting that property in the code when testing things)

server/services/interventionsService.ts
lawrence-forooghian: (Also, if we wanted to allow passing initial data, could we actually make that part of the `POST` body?)

tomsmyers: i actually initially wrote this that way. but i decided that changing the API was unnessecary since it's possible with a POST then PATCH. furthermore it makes error handling much harder. e.g. if you send the POST with invalid initial data does the referral actually get created?

lawrence-forooghian: Makes sense. I reckon the same applies here, then - there's no way of communicating to the caller of this method whether the referral got created and hence whether they need to try and recover by trying the PATCH again or by creating another referral.

tomsmyers: on the testing - i agree it makes the pact test not fully test the methods. but at the same time, limiting the logic we can have in the service layer in this way isn't ideal. i can write a normal unit test for this functionality aside from the contract tests. having the ability for a service level method to make multiple calls to achieve an 'action' like 'getOrCreate' for example, or 'createAndSet' (as we have here), or whatever, is good IMO.

server/services/interventionsService.test.ts
tomsmyers: ah i see. for now i might just remove the SU bit from the initial data, and address in another PR after george has done the work on getting SU data from ndelius. then maybe the initial data is the CRN, not SU name.

lawrence-forooghian (approved):

PR #61, opened 2021-01-06
Add usingRarDays and maximumRarDays to referral

lawrence-forooghian:

Adds the fields needed for the “Are you using RAR days for the … service?” question, with contract tests.

To allow the backend to add these fields which will be needed for an upcoming feature.


tomsmyers (commented):
LGTM - leaving open for karen to look at too

tomsmyers (approved):

PR #62, opened 2021-01-06
IC-673: Add the “Are you using RAR days?” question page

lawrence-forooghian:

Adds a page allowing the probation practitioner to state whether they are using RAR days for the intervention, and what is the maximum number of days that should be used.

The Jira story also uses the terminology "enforceable days", but says that we should go with what's in the prototype for now so that's what I've done.

To continue building out the referral form.

## Screenshot

image

### DIFF
--- a/server/routes/referrals/rarDaysForm.test.ts
+++ b/server/routes/referrals/rarDaysForm.test.ts
@@ -1,0 +1,152 @@
+import { Request } from 'express'
+import RarDaysForm from './rarDaysForm'
+import serviceCategoryFactory from '../../../testutils/factories/serviceCategory'
+
+describe(RarDaysForm, () => {
+  const serviceCategory = serviceCategoryFactory.build({ name: 'accommodation' })
+
+  describe('isValid', () => {
+    describe('when there are no errors in the form', () => {
+      it('returns true', async () => {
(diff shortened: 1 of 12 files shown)


server/routes/referrals/rarDaysForm.test.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    describe('when choosing no for using-rar-days', () => {
+      it('gives no error', async () => {
+        const form = await RarDaysForm.createForm({ body: { 'using-rar-days': 'no' } } as Request, serviceCategory)
+
+        expect(form.errors).toBeNull()
+      })
+    })
+
+    describe('when choosing yes for using-rar-days and providing a valid value for maximum-rar-days', () => {
```
Gweaton, line 45: This `describe` might need to be cut down (states "providing a valid value for max rar days" but then there are some different states, e.g. blank value in nested `describe`s)

Gweaton (commented):
Looks good, just one small thing.

server/routes/referrals/rarDaysForm.test.ts
lawrence-forooghian: Good spot, thanks. Fixed it.

Gweaton (approved):

PR #65, opened 2021-01-07
Make referral form more realistic and useful

lawrence-forooghian:

- adds links on the referral form for all the pages we've built so far
- removes links to pages we've not yet built
- removes the second service category from the referral form

To surface the work that we've done so far and to make the form look closer to the finished thing.


Gweaton (approved):
Nice one, good to have this finally linked up, I've had it in the back of my mind for too long.

PR #66, opened 2021-01-11
Handle validation errors from the interventions service

lawrence-forooghian:

This updates the frontend to handle the field-level validation errors returned by the backend as of https://github.com/ministryofjustice/hmpps-interventions-service/pull/53.

To be able to display granular error information to the user.

### DIFF
--- a/server/routes/referrals/completionDeadlinePresenter.ts
+++ b/server/routes/referrals/completionDeadlinePresenter.ts
@@ -1,6 +1,7 @@
 import { DraftReferral, ServiceCategory } from '../../services/interventionsService'
-import { CompletionDeadlineErrors } from './completionDeadlineForm'
 import CalendarDay from '../../utils/calendarDay'
+import { FormValidationError } from '../../utils/formValidationError'
+import ReferralDataPresenterUtils from './referralDataPresenterUtils'
 
 export default class CompletionDeadlinePresenter {
   readonly day: string
@@ -11,9 +12,13 @@ export default class CompletionDeadlinePresenter {
 
   readonly errorMessage: string | null
 
-  readonly erroredFields: ('day' | 'month' | 'year')[]
+  readonly hasMonthError: boolean
 
-  readonly errorSummary: CompletionDeadlineErrorSummaryPresenter | null
+  readonly hasDayError: boolean
+
+  readonly hasYearError: boolean
+
+  readonly errorSummary: { field: string; message: string }[] | null
 
   readonly title = `What date does the ${this.serviceCategory.name} service need to be completed by?`
 
@@ -22,7 +27,7 @@ export default class CompletionDeadlinePresenter {
   constructor(
     private readonly referral: DraftReferral,
     private readonly serviceCategory: ServiceCategory,
-    errors: CompletionDeadlineErrors | null = null,
+    error: FormValidationError | null = null,
     userInputData: Record<string, unknown> | null = null
   ) {
     if (!userInputData) {
@@ -45,20 +50,14 @@ export default class CompletionDeadlinePresenter {
       this.year = String(userInputData['completion-deadline-year'] || '')
     }
 
-    if (!errors) {
-      this.errorSummary = null
-      this.errorMessage = null
-      this.erroredFields = []
-    } else {
-      this.errorSummary = {
-        errors: [{ message: errors.message, linkedField: errors.firstErroredField }],
-      }
-      this.errorMessage = errors.message
-      this.erroredFields = errors.erroredFields
-    }
-  }
-}
+    this.errorSummary = ReferralDataPresenterUtils.errorSummary(error)
+    this.errorMessage =
+      ReferralDataPresenterUtils.errorMessage(error, 'completion-deadline-day') ??
+      ReferralDataPresenterUtils.errorMessage(error, 'completion-deadline-month') ??
+      ReferralDataPresenterUtils.errorMessage(error, 'completion-deadline-year')
 
-export interface CompletionDeadlineErrorSummaryPresenter {
-  errors: { message: string; linkedField: 'day' | 'month' | 'year' }[]
+    this.hasDayError = ReferralDataPresenterUtils.hasError(error, 'completion-deadline-day')
+    this.hasMonthError = ReferralDataPresenterUtils.hasError(error, 'completion-deadline-month')
+    this.hasYearError = ReferralDataPresenterUtils.hasError(error, 'completion-deadline-year')
+  }
 }
--- a/server/routes/referrals/referralsController.ts
+++ b/server/routes/referrals/referralsController.ts
@@ -1,6 +1,7 @@
 import { Request, Response } from 'express'
 import InterventionsService from '../../services/interventionsService'
 import { FormValidationError } from '../../utils/formValidationError'
+import createFormValidationErrorOrRethrow from '../../utils/interventionsFormError'
 import ReferralFormPresenter from './referralFormPresenter'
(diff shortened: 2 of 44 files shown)


server/routes/referrals/completionDeadlinePresenter.ts
```diff
@@ -45,20 +50,14 @@ export default class CompletionDeadlinePresenter {
...
-}
+    this.errorSummary = ReferralDataPresenterUtils.errorSummary(error)
+    this.errorMessage =
+      ReferralDataPresenterUtils.errorMessage(error, 'completion-deadline-day') ??
+      ReferralDataPresenterUtils.errorMessage(error, 'completion-deadline-month') ??
+      ReferralDataPresenterUtils.errorMessage(error, 'completion-deadline-year')
 
-export interface CompletionDeadlineErrorSummaryPresenter {
-  errors: { message: string; linkedField: 'day' | 'month' | 'year' }[]
+    this.hasDayError = !!error?.errors.find(subError => subError.formFields.includes('completion-deadline-day'))
```
Gweaton, line 59: Not a huge deal but would it be possible to wrap this in a nicely named function to make it quicker to parse and reduce repetition in the next three lines e.g. `hasError(error: FormValidationError | null, fieldName: string): boolean`

server/services/interventionsService.ts
```diff
@@ -51,42 +59,73 @@ export default class InterventionsService {
     return new RestClient('Interventions Service API Client', this.config, token)
   }
 
+  private convertError(error: unknown): InterventionsServiceError {
```
Gweaton, line 62: how about `sanitiseError` as we're returning a sanitised error?

server/utils/interventionsFormError.ts
```diff
@@ -0,0 +1,64 @@
+import { InterventionsServiceError, InterventionsServiceValidationError } from '../services/interventionsService'
+import interventionsServiceErrorMessages from './interventionsServiceErrorMessages'
+import { FormValidationError } from './formValidationError'
+
+type InterventionsFormError = { type: 'general'; message: string } | ({ type: 'validation' } & FormValidationError)
+
+// Add new date fields here.
```
Gweaton, line 7: Not sure what this comment is for?

server/routes/referrals/referralsController.ts
```diff
@@ -22,10 +22,20 @@ import RarDaysView from './rarDaysView'
 import RarDaysPresenter from './rarDaysPresenter'
 import RarDaysForm from './rarDaysForm'
 import { FormValidationError } from '../../utils/formValidationError'
+import createInterventionsFormError from '../../utils/interventionsFormError'
 
 export default class ReferralsController {
   constructor(private readonly interventionsService: InterventionsService) {}
 
+  private validationErrorOrThrow(error: InterventionsServiceError) {
```
Gweaton, line 30: Does this not need a return type?

server/utils/interventionsFormError.ts
```diff
@@ -0,0 +1,64 @@
+import { InterventionsServiceError, InterventionsServiceValidationError } from '../services/interventionsService'
+import interventionsServiceErrorMessages from './interventionsServiceErrorMessages'
+import { FormValidationError } from './formValidationError'
+
+type InterventionsFormError = { type: 'general'; message: string } | ({ type: 'validation' } & FormValidationError)
```
Gweaton, line 5: Let me know if this doesn't make sense, but I wonder if it's worth splitting this into two separate types of errors (validation and general) rather than adding a "type". The `validationErrorOrReturn` function introduced in the next commit feels a little bit to generalised (and a bit of a smell that it's doing too much).

server/routes/referrals/referralsController.ts
```diff
@@ -22,10 +22,20 @@ import RarDaysView from './rarDaysView'
...
+    const formError = createInterventionsFormError(error)
+    if (formError.type === 'validation') {
```
Gweaton, line 32: See my previous comment re: type checking inside the function.

Gweaton (commented):
I like the general approach here, just have a couple of pedantic naming comments and one re: single responsibility/type checking, but let me know if I've misunderstood.

server/services/interventionsService.ts
lawrence-forooghian: I'm hesitant to call it that for a couple of reasons:

1. I don't really understand what the word "sanitise" means in the context of doing something with an error, nor what the existing `sanitiseError` method is quite meant to do. I don't really want to use a piece of vocabulary that I don't fully understand - I don't know whether what I'm doing is "sanitising" in the sense that the existing method understands that word.
2. This `convertError` method assumes that its input is already a `SanitisedError`, so I'm not sure whether you can "sanitise" something that's already "sanitised".

server/utils/interventionsFormError.ts
lawrence-forooghian: It was meant to be a piece of instruction for people wanting to know where to add further date fields in the future, but it does look a bit like a TODO. I'll get rid of it.

server/routes/referrals/referralsController.ts
lawrence-forooghian: I don't think so - TypeScript does type inference for return types. But it might be helpful for readers here, I suppose. I've added one.

server/routes/referrals/completionDeadlinePresenter.ts
lawrence-forooghian: That's a good shout - added it.

server/utils/interventionsFormError.ts
lawrence-forooghian: Would you be able to expand a little on what you think the code might look like? The `type` exists so that, ultimately, in `ReferralsController#convertError` we're able to determine whether we have a validation error (in which case we return it) or some other kind of error (in which case we throw it). Do you think there's an alternative way of structuring things that means we can lose the `type`?

server/services/interventionsService.ts
Gweaton: Yeah, I guess "sanitise" might not be the right thing - I think my issue with `convertError` was, from what to what?  I wonder if there's a name that conveys that we're converting a form error, for example, to one of our well-crafted error types. I suppose I used the word `convert`, so I can see why you named it like you did! But unfortunately I can't think of an alternative this early in the day, sorry...

server/utils/interventionsFormError.ts
Gweaton: Having spent a bit too long thinking about this, the only alternative I've come up with would be to call this something like `interventionsFormValidationError` that basically does the `if` statement here, then in the call site (`validationErrorOrThrow`) you'd have:

```
if (interventionsServiceError.validationErrors) {
  return interventionsFormValidationError(interventionsServiceError)
} else {
  throw interventionsServiceError.message
}
```

I don't know if that's any better, it's basically doing the same thing but feels like _slightly_ less type checking? Feel free to ignore this though if you think it's not really an improvement.

server/services/interventionsService.ts
lawrence-forooghian: Gonna lose the transformation aspect of the name and just call it `createServiceError` 😀

server/utils/interventionsFormError.ts
lawrence-forooghian: Yeah, I think I like this - it still gives us what we need and takes the re-throw logic out of the controller. I've pushed a change, is that what you had in mind?

Gweaton: Yeah I think I prefer this, nicely done and thanks!

Gweaton (approved):
Thanks for the changes 👍

PR #67, opened 2021-01-11
Add contract tests for sending a draft referral and fetching a sent referral

lawrence-forooghian:

This adds an endpoint `POST /referral`, which takes the ID of a draft
referral and uses it to create a referral. This is just my idea of how
this would work; putting it up for review to get some other thoughts.

To allow us to build a UI for submitting a draft referral.

### DIFF
--- a/server/services/interventionsService.test.ts
+++ b/server/services/interventionsService.test.ts
@@ -862,6 +862,7 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
   const sentReferral: SentReferral = {
     id: '81d754aa-d868-4347-9c0f-50690773014e',
     createdAt: '2021-01-11T10:32:12.382884Z',
+    referenceNumber: 'HDJ2123F',
     referral: {
       completionDeadline: '2021-04-01',
       serviceProviderId: '674b47a0-39bf-4514-82ae-61885b9c0cb4',
--- a/server/services/interventionsService.test.ts
+++ b/server/services/interventionsService.test.ts
@@ -914,4 +914,27 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
       )
     })
   })
+
+  describe('getSentReferral', () => {
+    it('returns a referral for the given ID', async () => {
+      await provider.addInteraction({
+        state: 'There is an existing sent referral with ID of 81d754aa-d868-4347-9c0f-50690773014e',
+        uponReceiving: 'a request for the sent referral with ID of 81d754aa-d868-4347-9c0f-50690773014e',
+        withRequest: {
+          method: 'GET',
+          path: '/sent-referral/81d754aa-d868-4347-9c0f-50690773014e',
+          headers: { Accept: 'application/json', Authorization: `Bearer ${token}` },
+        },
+        willRespondWith: {
+          status: 200,
+          body: Matchers.like(sentReferral),
+          headers: { 'Content-Type': 'application/json' },
+        },
+      })
+
+      expect(await interventionsService.getSentReferral(token, '81d754aa-d868-4347-9c0f-50690773014e')).toEqual(
+        sentReferral
+      )
+    })
+  })
 })
--- a/server/services/interventionsService.test.ts
+++ b/server/services/interventionsService.test.ts
@@ -1,7 +1,7 @@
 import { pactWith } from 'jest-pact'
 import { Matchers } from '@pact-foundation/pact'
 
-import InterventionsService from './interventionsService'
+import InterventionsService, { SentReferral } from './interventionsService'
 import config from '../config'
 import oauth2TokenFactory from '../../testutils/factories/oauth2Token'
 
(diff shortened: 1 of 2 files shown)


lawrence-forooghian:
I've just added another commit, which adds a `referenceNumber` to the referral too.

server/services/interventionsService.test.ts
```diff
@@ -698,4 +698,40 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
...
+          path: '/referral',
+          headers: { Accept: 'application/json', Authorization: `Bearer ${token}` },
+          body: { draftReferralId: 'dfb64747-f658-40e0-a827-87b4b0bdcfed' },
+        },
+        willRespondWith: {
+          status: 201,
+          body: Matchers.like({
+            id: '81d754aa-d868-4347-9c0f-50690773014e',
+            created: '2021-01-11T10:32:12.382884Z',
+            referenceNumber: 'HDJ2123F',
```
lawrence-forooghian, line 719: Perhaps I should add examples of all the draft referral fields here to make it clear that the data is copied from the draft referral to the referral?

tomsmyers:
thanks for kicking the conversation off on this.

i think we can do this one of two ways:
- as is presented here; 'create' a referral from a draft referral id
- 'send' or 'submit' a draft referral `draft-referral/{id}/send`

personally i find the latter more intuitive. i think it comes down to whether or not we want to this of this as created a new entity, or changing the state of the draft referral.

lawrence-forooghian:
I remember that we originally had a `Referral` entity, and then we changed it to `DraftReferral` because we thought that a referral and a draft referral were two different things. So I assumed that a referral comes into existence at the point that a draft referral is submitted. 

> i think it comes down to whether or not we want to this of this as created a new entity, or changing the state of the draft referral.
Does it make sense to keep calling it a draft referral after it's been submitted?

lawrence-forooghian:
(I also guess I thought that creating a draft referral is more in line with the verb + resource pattern of REST.)

tomsmyers:
> I remember that we originally had a `Referral` entity, and then we changed it to `DraftReferral` because we thought that a referral and a draft referral were two different things. So I assumed that a referral comes into existence at the point that a draft referral is submitted.
> 
> > i think it comes down to whether or not we want to this of this as created a new entity, or changing the state of the draft referral.
> 
> Does it make sense to keep calling it a draft referral after it's been submitted?

yeh this is true. i would expect the result of a say, `POST /draft-referral/{id}/send` is something like `200 {"sentReferralId": "1234"}`

lawrence-forooghian:
I'm gonna update this in line with what we talked about this morning:

- `POST /draft-referral/:id/send`, returns a 201 with a `SentReferral` body
- and I'm gonna add the `GET /sent-referral/:id` endpoint

lawrence-forooghian:
@ministryofjustice/hmpps-interventions this is ready for review now, please could you take another look?

server/services/interventionsService.test.ts
tomsmyers: yeh works for me

tomsmyers: or we can just not return anything in the response, and use the location header. and validate the response in a different test.

server/services/interventionsService.ts
```diff
@@ -50,6 +50,12 @@ export interface ServiceProvider {
   name: string
 }
 
+export interface SentReferral {
+  id: string
+  // At some point we’ll want to add in all the fields of a DraftReferral here
```
tomsmyers, line 55: can we use https://www.typescriptlang.org/docs/handbook/advanced-types.html#intersection-types for this purpose?

tomsmyers (approved):
lgtm, couple non blocking comments

server/services/interventionsService.test.ts
lawrence-forooghian: It's a little easier for us to use the actual entity in the response, because after sending the referral I'm going to want to redirect to a different page like `/referrals/${sentReferralId}/confirmation` - this way I can just grab the `sentReferralId` from the response body instead of having to fetch the sent referral to find out its ID.

server/services/interventionsService.ts
lawrence-forooghian: Yeah, I didn't address it here because also need to figure out how to make everything non-nullable, possibly one of the [utility types](https://www.typescriptlang.org/docs/handbook/utility-types.html) would help but I haven't looked at them in detail. I can take a look into it tomorrow but didn't want to sink much time into it yet.

server/services/interventionsService.test.ts
lawrence-forooghian: Ah sorry, missed what you were talking about here - thought you were talking about the behaviour of the API, you were just talking about the test.

lawrence-forooghian: I've updated the tests to contain a full example of a sent referral.

server/services/interventionsService.ts
lawrence-forooghian: I've done this now.

PR #68, opened 2021-01-11
Add contract tests for service provider entity

lawrence-forooghian:

Adds an endpoint allowing us to fetch a new `ServiceProvider` entity, and adds a `serviceProviderId` property linking a draft referral to a service provider.

To allow us to display the name of the service provider on the referral submission confirmation page.

### DIFF
--- a/server/services/interventionsService.test.ts
+++ b/server/services/interventionsService.test.ts
@@ -72,6 +72,36 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
       })
     })
 
+    describe('for a referral that has had a service provider selected', () => {
+      beforeEach(async () => {
+        await provider.addInteraction({
+          state:
+            'There is an existing draft referral with ID of d496e4a7-7cc1-44ea-ba67-c295084f1962, and it has had a service provider selected',
+          uponReceiving: 'a request for that referral',
+          withRequest: {
+            method: 'GET',
+            path: '/draft-referral/d496e4a7-7cc1-44ea-ba67-c295084f1962',
+            headers: { Accept: 'application/json', Authorization: `Bearer ${token}` },
+          },
+          willRespondWith: {
+            status: 200,
+            body: Matchers.like({
+              id: 'd496e4a7-7cc1-44ea-ba67-c295084f1962',
+              serviceProviderId: '674b47a0-39bf-4514-82ae-61885b9c0cb4',
+            }),
+            headers: { 'Content-Type': 'application/json' },
+          },
+        })
+      })
+
+      it('returns a referral for the given ID, with the serviceProviderId field populated', async () => {
+        const referral = await interventionsService.getDraftReferral(token, 'd496e4a7-7cc1-44ea-ba67-c295084f1962')
+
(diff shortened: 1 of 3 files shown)


server/services/interventionsService.test.ts
```diff
@@ -72,6 +72,36 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
       })
     })
 
+    describe('for a referral that has had a service provider selected', () => {
+      beforeEach(async () => {
+        await provider.addInteraction({
+          state:
+            'There is an existing draft referral with ID of d496e4a7-7cc1-44ea-ba67-c295084f1962, and it has had a service provider selected',
```
sldblog, line 79: > and it has had a service provider selected

I'm still confused about when this selection happens: before draft referrals are created or during... Not a blocker, though!

server/services/interventionsService.test.ts
```diff
@@ -72,6 +72,36 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
...
+          withRequest: {
+            method: 'GET',
+            path: '/draft-referral/d496e4a7-7cc1-44ea-ba67-c295084f1962',
+            headers: { Accept: 'application/json', Authorization: `Bearer ${token}` },
+          },
+          willRespondWith: {
+            status: 200,
+            body: Matchers.like({
+              id: 'd496e4a7-7cc1-44ea-ba67-c295084f1962',
+              serviceProviderId: '674b47a0-39bf-4514-82ae-61885b9c0cb4',
```
sldblog, line 90: _thinking out loud, not a blocker_
This could be `selectedServiceProviderId` to reinforce that we elect one of them to be the "recipient" of the referral. The context of provider/referral might be obvious as it is, though.

sldblog (approved):
👍 LGTM -- no blockers

Gweaton (approved):

server/services/interventionsService.test.ts
Gweaton: I think it's when we've gone through the "Find" process and an intervention has been selected, but I could be wrong about that...

tomsmyers: yeh the service itself (and the service user) will be selected right from the start.

lawrence-forooghian: Yeah, good point. When we first started writing these contract tests, it was with the original user journey in mind, where the draft referral is created and then the user Find-s an intervention (i.e. service provider and service category). Now that Find is done before the user starts the referral form, we _could_ make the service category ID and service provider ID non-nullable on a draft referral, but I've kept it nullable to be consistent with [the corresponding test for `serviceCategoryId`](https://github.com/ministryofjustice/hmpps-interventions-ui/blob/672a957cf3aecd976639682c11de0c82ce2672a5/server/services/interventionsService.test.ts#L44-L73).

server/services/interventionsService.test.ts
lawrence-forooghian: I think that if there were multiple different ways in which a service provider could be related to a referral, then it would make sense to disambiguate them, but I'm struggling to think of what these might be.

PR #69, opened 2021-01-12
change the user id in the 'getDraftReferralsForUser' pact test

tomsmyers:

change the user id in the 'getDraftReferralsForUser' pact test

the reason for doing this is the previous user id is used in tests that create new draft referrals, which means that we cannot predictably test the number of entries returned when querying `GET /draft-referrals`

_Required._

### DIFF
--- a/server/services/interventionsService.test.ts
+++ b/server/services/interventionsService.test.ts
@@ -734,12 +734,12 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
   describe('getDraftReferralsForUser', () => {
     it('returns a list of draft referrals for a given userID', async () => {
       await provider.addInteraction({
-        state: 'a referral for user with ID 2500128586 exists',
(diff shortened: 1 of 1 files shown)


server/services/interventionsService.test.ts
```diff
@@ -734,12 +734,12 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
   describe('getDraftReferralsForUser', () => {
     it('returns a list of draft referrals for a given userID', async () => {
       await provider.addInteraction({
-        state: 'a referral for user with ID 2500128586 exists',
+        state: 'a referral for user with ID 8751622134 exists',
```
sldblog, line 737: What do you think about electing named users for these cases? I.e. `for a user with a single referral`?

tomsmyers: 👍 can do

sldblog (approved):
👍

PR #76, opened 2021-01-18
IC-627: Add integration test for creating and submitting a referral

lawrence-forooghian:

Adds an integration test, using Cypress and Wiremock, that simulates a user creating and submitting a referral.

To increase our confidence that the forms / presenters / controllers are glued together properly.

## Video (super fast)

https://user-images.githubusercontent.com/53756884/104944024-8118cd80-59ae-11eb-90da-5ab467cc0a79.mp4

### DIFF
--- a/integration_tests/support/commands.js
+++ b/integration_tests/support/commands.js
@@ -2,3 +2,7 @@ Cypress.Commands.add('login', () => {
   cy.request(`/`)
   cy.task('getLoginUrl').then(cy.visit)
 })
+
+Cypress.Commands.add('withinFieldsetThatContains', (text, action) => {
+  cy.contains(text).parent('fieldset').within(action)
+})
--- a/integration_tests/integration/referral.spec.js
+++ b/integration_tests/integration/referral.spec.js
@@ -1,0 +1,140 @@
+import draftReferralFactory from '../../testutils/factories/draftReferral'
+import sentReferralFactory from '../../testutils/factories/sentReferral'
+import serviceCategoryFactory from '../../testutils/factories/serviceCategory'
+import serviceProviderFactory from '../../testutils/factories/serviceProvider'
+
+describe('Referral form', () => {
+  beforeEach(() => {
+    cy.task('reset')
+    cy.task('stubLogin')
+    cy.task('stubAuthUser')
+  })
+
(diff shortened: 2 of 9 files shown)


integration_tests/support/commands.js
```diff
@@ -2,3 +2,7 @@ Cypress.Commands.add('login', () => {
   cy.request(`/`)
   cy.task('getLoginUrl').then(cy.visit)
 })
+
+Cypress.Commands.add('withinFieldsetThatContains', (text, action) => {
```
Gweaton, line 6: 👍

Gweaton (approved):
This is great, nice work. Do we think it'll be possible to switch over to the real `interventions-service` in future, and/or do we trust that behaviour is covered by Pact tests?

lawrence-forooghian:
> Do we think it'll be possible to switch over to the real `interventions-service` in future, and/or do we trust that behaviour is covered by Pact tests?

I imagine it would be possible, once all of the corresponding functionality is built in the backend. But I think that for most of the initial work we're doing on the UI, where the backend hasn't been built yet, we'll have to stick to using mocks.

Also, although mocks aren't as realistic as using the real service, I do like how they're so predictable to work with - no having to try and reset the service between tests, or trying to write tests that ignore the side-effects of previous tests. So I'm not sure yet whether we should use the real service even if we could. Do you have any thoughts about this?

I think that the Pact tests do correctly test all of the backend behaviour. But the missing link is that there's no guarantee that the mocks that I've set up in these integration tests are in sync with the expectations of the Pact tests, so there's the potential for a gap to form there. I don't have a good idea of how to improve that yet. I don't know if there's something more that we could be sharing between the two tests (possibly use the factories in the Pact tests...?)

lawrence-forooghian:
I think it might be useful to think of Cypress as an item in a toolbox instead of being used to define the position of a test in the "testing pyramid". That is to say, we use Cypress when we want to do things that we can't easily do using a Jest test. That might be something like checking that a particular error message is displayed correctly and that particular fields are highlighted, or it might be simulating the actions of a user throughout a particular journey. Some of these things can be considered "integration" tests and some "end to end" tests, and that might inform whether we use mocking or not for a particular test. So I don't think it necessarily has to be "Cypress implies use a mock" or "Cypress implies use the real service".

PR #78, opened 2021-01-19
Watch files when running test server

lawrence-forooghian:

Runs the test server using `npm run start:dev`, so that it restarts automatically when the code changes.

It’s nice to not need to restart the server in order for changes to be picked up in Cypress tests.


lawrence-forooghian:
Hmm, I need to check how this works when you have two different commands simultaneously running (dev server and test server) both trying to write to `dist` etc, let's not merge this yet.

lawrence-forooghian:
I'll close this for now so it's not cluttering up our inbox.

PR #79, opened 2021-01-20
IC-946, IC-947: Redirect users to different start pages depending on whether they're a probation practitioner or a service provider

tomsmyers:

- Adds a new placeholder "receive referrals dashboard" page for service providers
- Makes the `/` route redirect to either the referral start page or the receive dashboard depending on whether the logged-in user is a probation practitioner or a service provider

To prove that we can distinguish between different types of user.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,9 +21,10 @@ services:
       test: ['CMD', 'curl', '-f', 'http://localhost:8090/auth/health']
     environment:
       - SERVER_PORT=8090
-      - SPRING_PROFILES_ACTIVE=dev
+      - SPRING_PROFILES_ACTIVE=dev,delius
       - SPRING_JPA_PROPERTIES_HIBERNATE_SHOW_SQL=false
       - SPRING_H2_CONSOLE_SETTINGS_WEBALLOWOTHERS=true
+      - DELIUS_ENDPOINT_URL=http://community-api:8080
 
   community-api:
     image: quay.io/hmpps/community-api:latest
--- a/server/routes/index.ts
+++ b/server/routes/index.ts
@@ -29,8 +29,15 @@ export default function routes(router: Router, services: Services): Router {
   const referralsController = new ReferralsController(services.interventionsService, services.communityApiService)
 
   get('/', (req, res, next) => {
-    res.render('pages/index')
+    const { authSource } = res.locals.user
+    if (authSource === 'delius') {
+      res.redirect('/referrals/start')
+    } else {
+      res.redirect('/receive/dashboard')
+    }
   })
+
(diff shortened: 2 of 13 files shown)


docker-compose.yml
```diff
@@ -21,8 +21,9 @@ services:
       test: ['CMD', 'curl', '-f', 'http://localhost:8090/auth/health']
     environment:
       - SERVER_PORT=8090
-      - SPRING_PROFILES_ACTIVE=dev
+      - SPRING_PROFILES_ACTIVE=dev,delius
       - SPRING_JPA_PROPERTIES_HIBERNATE_SHOW_SQL=false
+      - DELIUS_ENDPOINT_URL=http://community-api:8080
```
Gweaton, line 27: Would you mind adding a bit more detail as to why this has been added to the commit message e.g. "to configure the service look up the user from the locally running instance of the community api"?

server/routes/index.ts
```diff
@@ -29,9 +29,16 @@ export default function routes(router: Router, services: Services): Router {
   const referralsController = new ReferralsController(services.interventionsService, services.communityApiService)
 
   get('/', (req, res, next) => {
-    res.render('pages/index')
+    const { authSource } = res.locals.user
+    if (authSource === 'delius') {
```
Gweaton, line 33: Optional, but I wonder if this behaviour is something that might be better extracted to a middleware function (like `authorizationMiddleware` to keep this logic out of the routes file?

Gweaton (commented):
Looks good, just a couple of minor things.

server/routes/index.ts
lawrence-forooghian: Good question, I'm not sure where's the best place for it. I guess that `authorizationMiddleware` acts on all requests and is designed to block the user from performing certain actions if they shouldn't be allowed to. Whereas all that this code is doing is just deciding where `/` takes a given user, which I guess I thought of as "routing". But not sure. Perhaps we could leave it here for now.

docker-compose.yml
lawrence-forooghian: 👍 I felt dirty changing somebody else's commit message, but I powered through the guilt and did it

server/routes/index.ts
Gweaton: Yeah, we can leave it here for now. Sorry, if it wasn't clear, I meant using `authorizationMiddleware` as an example, rather than putting this behaviour in there, we'd probably create a new middleware. I guess in future we could use something like that to stop service providers being able to access the "make a referral" & "monitor" flows, but it's not something we've really thought about yet (and not necessarily relevant to this logic that directs a user to the correct homepage)

PR #82, opened 2021-01-21
IC-972: Add service user to referral upon starting the referral form

Gweaton:

- Adds the service user to a referral after entering a CRN and looking up the service user in nDelius.
- Adds a `SerializeDeliusServiceUser`(better names welcome) to convert the response from the Community API into a format the Interventions Service expects.
- Updates the Pacts tests to accept a whole service user object.

This is potentially a temporary way of doing things, as we will probably find the service user earlier in the process, but this allows us to use the referral form with a Servce User selected on a referral, and will allow us to view and confirm Service User details.

## Next steps

Render service user details on their own page.


server/services/communityApiService.ts
```diff
@@ -14,16 +14,38 @@ export interface DeliusUser {
...
+  disabilities: string
 }
 
 interface DeliusRole {
   name: string
 }
 
+interface OtherIds {
+  crn: string
+  nomsNumber: string
```
tomsmyers, line 35: heads up we have decided to de-scope these other ids for MVP. we're only storing CRN.

server/services/communityApiService.ts
```diff
@@ -14,16 +14,38 @@ export interface DeliusUser {
...
+  mostRecentPrisonerNumber: string
+}
+
+interface OffenderProfile {
+  offenderLanguages: OffenderLanguages
+}
+
+interface OffenderLanguages {
```
tomsmyers, line 43: similarly here, we only require primary language at this stage. we capture need for interpreter later in the form and other languages is not used anywhere else AFAIK.

tomsmyers: ah i see here we're just replicating the data structure used in community api. that's cool. ignore these comments.

server/services/interventionsService.test.ts
```diff
@@ -991,3 +1051,41 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
     })
   })
 })
+
```
tomsmyers, line 1054: do these new contracts already pass or is there any work required on backend for these?

tomsmyers: comment on wrong part of file- but i'm talking about the contract test above

server/services/interventionsService.test.ts
```diff
@@ -991,3 +1051,41 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
...
+        },
+      },
+      title: 'Mr',
+      firstName: 'Alex',
+      surname: 'River',
+      dateOfBirth: '01/01/1980',
+      gender: 'Male',
+      ethnicity: 'British',
+      religionOrBelief: 'Agnostic',
+      disabilities: 'Autism',
```
tomsmyers, line 1079: question on disabilities.- i thought these came from delius as a list of objects with start and end dates in?

server/services/communityApiService.ts
Gweaton: Yeah, I'm following the structure for the object in Delius with this object, I left off some other fields, so could get rid of these if we're not going to use them (same with needs interpreter below)

server/services/interventionsService.test.ts
Gweaton: ah yeah, you're right, will fix that.

server/services/communityApiService.ts
Gweaton: Have removed these fields we're not using at the moment.

server/services/interventionsService.ts
```diff
@@ -95,6 +104,33 @@ export default class InterventionsService {
     return sanitisedError
   }
 
+  serializeDeliusServiceUser(deliusServiceUser: DeliusServiceUser | null): ServiceUser {
+    if (!deliusServiceUser) {
+      return {} as ServiceUser
+    }
+
+    const currentDisabilities = deliusServiceUser.disabilities
```
tomsmyers, line 112: just a note that we discussed including "ended" disabilities with a postfix such as `(expired)`, but having read this i do think it's probably better to just not include them at all.

tomsmyers (approved):

server/services/interventionsService.ts
```diff
@@ -103,6 +104,33 @@ export default class InterventionsService {
...
+      .filter(disability => {
+        const today = new Date().toString()
+        return disability.endDate === '' || Date.parse(disability.endDate) >= Date.parse(today)
+      })
+      .map(disability => disability.disabilityType.description)
```
lawrence-forooghian, line 117: Why do we want to do this formatting instead of storing it as structured data (i.e. a list) in the interventions service? It seems like we're narrowing the range of things we could do with this data later.

server/services/interventionsService.test.ts
```diff
@@ -1050,3 +1051,66 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
...
+          endDate: '2020-01-22',
+          notes: 'Some notes',
+          startDate: '2020-01-22',
+        },
+      ],
+    } as DeliusServiceUser
+
+    const serviceUser = interventionsService.serializeDeliusServiceUser(deliusServiceUser)
+
+    expect(serviceUser.crn).toEqual('X123456')
```
lawrence-forooghian, line 1108: Perhaps we could test the whole object instead so that we don't miss out on testing any fields? e.g. `dateOfBirth` is missing here.

server/services/interventionsService.ts
```diff
@@ -103,6 +104,33 @@ export default class InterventionsService {
...
+      .join(', ')
+
+    return {
+      crn: deliusServiceUser.otherIds.crn,
+      title: deliusServiceUser.title,
+      firstName: deliusServiceUser.firstName,
+      lastName: deliusServiceUser.surname,
+      dateOfBirth: deliusServiceUser.dateOfBirth,
```
lawrence-forooghian, line 125: I imagine that the interventions service would want to enforce a format here - i.e. ISO date format like we use for e.g. completion deadline. So looks like some translation might be required from the Delius data.

server/routes/referrals/referralsController.test.ts
```diff
@@ -50,10 +64,31 @@ describe('POST /referrals/start', () => {
   describe('when searching for a CRN found in Delius', () => {
     beforeEach(() => {
       communityApiService.getServiceUserByCRN.mockResolvedValue({
```
lawrence-forooghian, line 66: Does the return value here matter if we're mocking out `serializeDeliusServiceUser` anyway? And could we use `deliusServiceUser` factory to create this object?

server/services/communityApiService.ts
```diff
@@ -14,16 +14,42 @@ export interface DeliusUser {
 }
 
 export interface DeliusServiceUser {
-  offenderId: string
+  otherIds: OtherIds
+  offenderProfile: OffenderProfile
```
lawrence-forooghian, line 18: I guess this was true before, but as this data structure gets larger we're making more and more assumptions about the data received in the response. At some point we'll need to validate these assumptions instead of just doing a type assertion, else we might end up passing malformed data to the interventions service. This can be covered within the scope of [IC-620](https://dsdmoj.atlassian.net/browse/IC-620); can we put a TODO for that?

server/services/interventionsService.test.ts
Gweaton: Yeah, I mainly tested the ones that involve the most wrangling on our side, and left most of the ones that have the same key, but I'll add the rest in.

server/services/interventionsService.ts
Gweaton: At the moment, we only want to display it on the user details page, and the shape of the data might be different once we switch to using the Probation Offender Search - how would you feel about us getting this in as is and making tweaks in future when we need to use this data for other purposes (e.g. Discover?)

server/services/interventionsService.ts
Gweaton: Good point, I'll convert to ISO here like completion deadline.

server/services/interventionsService.ts
lawrence-forooghian: Yeah, would it be much more work to just store it as a list of strings for now and then format them in a presenter? That's what the presenters are for.

I think that without seeing examples of real data, I don't know whether just sticking the disabilities from Delius together as a comma-separated list would make sense - e.g. are they all single nouns or might they be whole sentences / narrative things?

server/routes/referrals/referralsController.test.ts
Gweaton: You're right. I think it's quite nice to have the whole object in the test as a form of documentation, but I'll use the factory for it (I don't think it's got much use otherwise)

server/services/interventionsService.ts
Gweaton: Yeah, I initially did put this in, but this morning had the thought that actually it's probably not relevant to the service provider if it's not a current disability.

server/services/interventionsService.ts
tomsmyers: in terms of the format of the data i have no idea what it will look like. 

in terms of what we do with it i think this is purely informational for the SP, so as long as we can display it i don't really care how it's stored. there's a bit more overhead storing lists in the db, but it's not a big deal. pre-processing the data and storing it as a simple string is my preference though.

lawrence-forooghian: > I don't know whether just sticking the disabilities from Delius together as a comma-separated list would make sense - e.g. are they all single nouns or might they be whole sentences / narrative things?

Any thoughts on this @tomsmyers? The point being that keeping the data structured gives us the flexibility to stick them into an `<ul>` or whatever later on.

tomsmyers: i'll make the backend changes now accept a json list for this field. sounds like it makes everything more flexible.

server/services/communityApiService.ts
Gweaton: I've begrudgingly written my first TODO (you know how I feel about comments in code 🤐)

server/services/interventionsService.ts
Gweaton: Turns out we can actually use the `CalendarDay` functionality here as-is, as I got the format back from Delius wrong, it's actually e.g. `1980-01-01`

server/services/interventionsService.ts
Gweaton: I've changed this to an array of strings, I hope that's not going to be too much work to change @tomsmyers ? I'll format it nicely in the presenter when we display the page.

server/services/communityApiService.ts
Gweaton: Ugh, I've just tried testing this out and everything needs to be nullable here...

server/services/interventionsService.ts
```diff
@@ -95,6 +105,34 @@ export default class InterventionsService {
...
+    }
+
+    const currentDisabilities = deliusServiceUser.disabilities
+      .filter(disability => {
+        const today = new Date().toString()
+        return disability.endDate === '' || Date.parse(disability.endDate) >= Date.parse(today)
+      })
+      .map(disability => disability.disabilityType.description)
+
+    const iso8601DateOfBirth = CalendarDay.parseIso8601(deliusServiceUser.dateOfBirth)?.iso8601 || null
```
lawrence-forooghian, line 120: Why are we parsing an ISO 8601 string and then converting the result back to ISO 8601? Is it to validate that it's correctly formatted?

Gweaton: Yeah, I wasn't sure if we could rely on it being in the correct format, but maybe we can...

lawrence-forooghian: I think it's sensible to check. Our ultimate goal is to be checking everything, so why not start here!

lawrence-forooghian (approved):

PR #85, opened 2021-01-25
fix 'no draft referrals' text on start referrals page

tomsmyers:

updates the name of a presenter field in the start referral page which has been refactored

show the message when there are no draft referrals to display


sldblog (commented):
_pondering_
Should these situations generate some kind of warning/build failure?

Gweaton:
We could possibly use something like snapshot testing for this, e.g.

image

It would mean we'd probably have to test a variety of scenarios though e.g. both no draft referrals and draft referrals. Possibly one for dev team chatting?

Gweaton (approved):

PR #86, opened 2021-01-25
IC-972 Add Service User Details Page

Gweaton:

Adds service user details page to the form. Clicking "save and continue" takes you back to the form. 

Local data is currently very limited, so I'd like to check it out on Dev too. It'd be good to add some data to the local community API image.

To demonstrate that we can display user details in the referral form and integrate with the community API.

## Screenshot

image

### DIFF
--- a/server/routes/index.ts
+++ b/server/routes/index.ts
@@ -45,6 +45,8 @@ export default function routes(router: Router, services: Services): Router {
   get('/referrals/start', (req, res) => referralsController.startReferral(req, res))
   post('/referrals/start', (req, res) => referralsController.createReferral(req, res))
   get('/referrals/:id/form', (req, res) => referralsController.viewReferralForm(req, res))
+  get('/referrals/:id/service-user-details', (req, res) => referralsController.viewServiceUserDetails(req, res))
+  post('/referrals/:id/service-user-details', (req, res) => referralsController.confirmServiceUserDetails(req, res))
   get('/referrals/:id/complexity-level', (req, res) => referralsController.viewComplexityLevel(req, res))
   post('/referrals/:id/complexity-level', (req, res) => referralsController.updateComplexityLevel(req, res))
   get('/referrals/:id/completion-deadline', (req, res) => referralsController.viewCompletionDeadline(req, res))
--- a/server/routes/referrals/needsAndRequirementsPresenter.ts
+++ b/server/routes/referrals/needsAndRequirementsPresenter.ts
@@ -1,5 +1,6 @@
 import { DraftReferral } from '../../services/interventionsService'
 import { FormValidationError } from '../../utils/formValidationError'
+import { SummaryListItem } from '../../utils/summaryList'
 import ReferralDataPresenterUtils from './referralDataPresenterUtils'
 
 export default class NeedsAndRequirementsPresenter {
@@ -9,7 +10,7 @@ export default class NeedsAndRequirementsPresenter {
     private readonly userInputData: Record<string, unknown> | null = null
   ) {}
 
-  readonly summary: { key: string; lines: string[]; isList: boolean }[] = [
+  readonly summary: SummaryListItem[] = [
     { key: 'Needs', lines: ['Accommodation', 'Social inclusion'], isList: true },
     { key: 'Gender', lines: ['Male'], isList: false },
     // TODO IC-746 populate with service user data once we have it
--- a/server/routes/referrals/needsAndRequirementsView.ts
+++ b/server/routes/referrals/needsAndRequirementsView.ts
@@ -7,28 +7,7 @@ export default class NeedsAndRequirementsView {
   private readonly errorSummaryArgs = ViewUtils.govukErrorSummaryArgs(this.presenter.errorSummary)
 
   private get summaryListArgs() {
-    return {
-      rows: this.presenter.summary.map(item => {
-        return {
-          key: {
-            text: item.key,
-          },
-          value: (() => {
-            if (item.isList) {
-              const html = `<ul class="govuk-list">${item.lines
-                .map(line => `<li>${ViewUtils.escape(line)}</li>`)
-                .join('\n')}</ul>`
-              return { html }
-            }
-            if (item.lines.length > 1) {
-              const html = item.lines.map(line => `<p class="govuk-body">${ViewUtils.escape(line)}</p>`).join('\n')
-              return { html }
-            }
-            return { text: item.lines[0] || '' }
-          })(),
-        }
-      }),
-    }
+    return ViewUtils.summaryListArgs(this.presenter.summary)
   }
 
   private get additionalNeedsInformationTextareaArgs() {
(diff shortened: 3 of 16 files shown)


tomsmyers:
can we make the start referral button go directly to this page? i wonder if we should only create the referral once the PP has confirmed the SU details. that way if there's a mistake in the crn, the referral isn't created and the PP can go 'back' in the browser and enter another.

Gweaton:
> can we make the start referral button go directly to this page? i wonder if we should only create the referral once the PP has confirmed the SU details. that way if there's a mistake in the crn, the referral isn't created and the PP can go 'back' in the browser and enter another.

I had a chat with Ming and Marcin yesterday, and there was talk of possibly combining the first three steps of the referral form into one section (e.g. confirm service user details, confirm/check risk and needs). As part of walking skeleton/MVP we discussed the option of just requiring PPs to delete bad referrals at the end of the journey, or maybe they could delete drafts this way.

server/routes/referrals/needsAndRequirementsView.ts
```diff
@@ -7,28 +7,7 @@ export default class NeedsAndRequirementsView {
...
-            if (item.lines.length > 1) {
-              const html = item.lines.map(line => `<p class="govuk-body">${ViewUtils.escape(line)}</p>`).join('\n')
-              return { html }
-            }
-            return { text: item.lines[0] || '' }
-          })(),
-        }
-      }),
-    }
+    return ViewUtils.summaryListArgs(this.presenter.summary)
```
lawrence-forooghian, line 10: Nice!

server/utils/viewUtils.ts
```diff
@@ -27,4 +27,43 @@ export default class ViewUtils {
...
+    return {
+      rows: summaryListItems.map(item => {
+        return {
+          key: {
+            text: item.key,
+          },
+          value: (() => {
+            if (item.isList) {
+              const html = `<ul class="govuk-list">${item.lines
+                .map(line => `<li>${ViewUtils.escape(line)}</li>`)
```
lawrence-forooghian, line 41: Would be good to test that this escaping works.

server/utils/viewUtils.ts
```diff
@@ -27,4 +27,43 @@ export default class ViewUtils {
...
+                .join('\n')}</ul>`
+              return { html }
+            }
+            if (item.lines.length > 1) {
+              const html = item.lines.map(line => `<p class="govuk-body">${ViewUtils.escape(line)}</p>`).join('\n')
```
lawrence-forooghian, line 46: Would be good to test that you get `<p>` tags when there are multiple lines and it's not a list.

server/routes/referrals/needsAndRequirementsPresenter.ts
```diff
@@ -1,5 +1,6 @@
 import { DraftReferral } from '../../services/interventionsService'
 import { FormValidationError } from '../../utils/formValidationError'
+import { SummaryListItem } from '../../utils/viewUtils'
```
lawrence-forooghian, line 3: I think that the presenters should be written without reference to any view concepts / files. In general the presenter doesn't have any knowledge of the view. So I think best to put this in an independent file that's not part of something view-y, imported here and in `viewUtils`, like `summaryListPresenter.ts` or something.

server/routes/referrals/serviceUserDetailsPresenter.ts
```diff
@@ -0,0 +1,21 @@
...
+import { SummaryListItem } from '../../utils/viewUtils'
+
+export default class ServiceUserDetailsPresenter {
+  constructor(private readonly serviceUser: ServiceUser) {}
+
+  readonly title = `${this.serviceUser.firstName || 'Service user'}'s information`
+
+  readonly summary: SummaryListItem[] = [
+    { key: 'CRN', lines: [this.serviceUser.crn], isList: false },
+    { key: 'Title', lines: [this.serviceUser.title || ''], isList: false },
```
lawrence-forooghian, line 11: I think that, although in this case, the outcome will be the same, it's a bit clearer to use the [nullish coalescing operator `??`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) instead of `||`, since `||` tests truthiness (e.g. is the string non-empty) but I think we just care whether it's `null` or not.

server/routes/referrals/serviceUserDetailsPresenter.test.ts
```diff
@@ -0,0 +1,80 @@
...
+    title: 'Mr',
+    firstName: 'Alex',
+    lastName: 'River',
+    dateOfBirth: '1980-01-01',
+    gender: 'Male',
+    ethnicity: 'British',
+    preferredLanguage: 'English',
+    religionOrBelief: 'Agnostic',
+    disabilities: ['Autism spectrum condition', 'sciatica'],
+  } as ServiceUser
```
lawrence-forooghian, line 16: Why do we need the type assertion?

server/routes/index.ts
```diff
@@ -45,6 +45,8 @@ export default function routes(router: Router, services: Services): Router {
   get('/referrals/start', (req, res) => referralsController.startReferral(req, res))
   post('/referrals/start', (req, res) => referralsController.createReferral(req, res))
   get('/referrals/:id/form', (req, res) => referralsController.viewReferralForm(req, res))
+  get('/referrals/:id/service-user-details', (req, res) => referralsController.viewServiceUserDetails(req, res))
```
lawrence-forooghian, line 48: Can we add this page to our referral integration test?

Gweaton: Have done this - i wasn't sure what level of detail we wanted to go into here, am testing all the service user details page, let me know if you have a better suggestion.

server/routes/referrals/serviceUserDetailsPresenter.test.ts
Gweaton: Oops, left this in by accident, it's a good way of getting vs code to tell you the fields you're missing in a nicer way than the usual errors.

server/routes/referrals/serviceUserDetailsPresenter.ts
Gweaton: Yeah, I keep seeing you using this operator but hadn't taken enough time to look into it fully, have changed it.

server/routes/referrals/needsAndRequirementsPresenter.ts
Gweaton: I've extracted into a util called `summaryList`, I thought that might be better than a fully-fledged presenter, as we seem to have specific patterns for those in this project.

server/utils/viewUtils.ts
Gweaton: Have added this to the current test

server/utils/viewUtils.ts
Gweaton: Have added a new test for this

server/routes/index.ts
lawrence-forooghian: Yep, I think it looks good, I'm not too sure about level of detail either when it comes to this sort of test!

lawrence-forooghian (approved):

PR #91, opened 2021-01-27
IC-909: Make intervention details section required before being able to view check answers page

Gweaton:

- Requires PP to fill in the minimum intervention details before being able to submit their referral to a service provider.
- Displays "Not started" or "Completed" badge by these sections, removing "in progress" for now.
- Removes 'multi' form section for referrals, as we're only showing one, so we don't need a sub-heading for each referral type selected.

To ensure we have the minimum intervention details before submitting a referral to the back-end, so a user doesn't submit their referral too early in the process and get stuck with an unfriendly error message.

### Before completing intervention details section

image

### Intervention details section completed

image

### DIFF
--- a/integration_tests/integration/referral.spec.js
+++ b/integration_tests/integration/referral.spec.js
@@ -60,7 +60,24 @@ describe('Referral form', () => {
       },
     })
 
-    const sentReferral = sentReferralFactory.fromFields(draftReferral).build()
+    const completedDraftReferral = {
+      ...draftReferral,
+      completionDeadline: '2021-04-01',
+      complexityLevelId: 'd0db50b0-4a50-4fc7-a006-9c97530e38b2',
+      furtherInformation: 'Some information about the service user',
+      desiredOutcomesIds: ['3415a6f2-38ef-4613-bb95-33355deff17e', '5352cfb6-c9ee-468c-b539-434a3e9b506e'],
+      additionalNeedsInformation: 'Alex is currently sleeping on her aunt’s sofa',
+      accessibilityNeeds: 'She uses a wheelchair',
+      needsInterpreter: true,
+      interpreterLanguage: 'Spanish',
+      hasAdditionalResponsibilities: true,
+      whenUnavailable: 'She works Mondays 9am - midday',
+      additionalRiskInformation: 'A danger to the elderly',
+      usingRarDays: true,
+      maximumRarDays: 10,
+    }
+
+    const sentReferral = sentReferralFactory.fromFields(completedDraftReferral).build()
 
     cy.stubGetServiceUserByCRN('X320741', deliusServiceUser)
     cy.stubCreateDraftReferral(draftReferral)
@@ -155,6 +172,9 @@ describe('Referral form', () => {
     cy.location('pathname').should('equal', `/referrals/${draftReferral.id}/further-information`)
     cy.get('h1').contains('Do you have further information for the accommodation service provider? (optional)')
     cy.get('textarea').type('Some information about Geoffrey')
+
+    // stub completed draft referral to mark section as completed
+    cy.stubGetDraftReferral(draftReferral.id, completedDraftReferral)
 
     cy.contains('Save and continue').click()
 
--- a/server/routes/referrals/referralFormPresenter.test.ts
+++ b/server/routes/referrals/referralFormPresenter.test.ts
@@ -11,7 +11,7 @@ describe('ReferralFormPresenter', () => {
           .serviceCategorySelected(serviceCategory.id)
           .completionDeadlineSet()
           .build()
-        const presenter = new ReferralFormPresenter(referral)
+        const presenter = new ReferralFormPresenter(referral, 'social inclusion')
 
         const expected = [
           {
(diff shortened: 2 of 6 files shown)


server/routes/referrals/referralFormPresenter.test.ts
```diff
@@ -4,85 +4,184 @@ import serviceCategoryFactory from '../../../testutils/factories/serviceCategory
...
-          number: '6',
-          status: ReferralFormStatus.CannotStartYet,
-          tasks: [{ title: 'Check your answers', url: 'check-answers' }],
-        },
-      ]
+        expect(presenter.sections).toEqual(expected)
+      })
+    })
+
+    describe('when the intervention details section has all required fields filled in', () => {
```
lawrence-forooghian, line 84: Might this draft referral be a better example of the state described here if it _only_ has the required fields filled in, and the rest (e.g. `interpreterLanguage`) `null`?

server/routes/referrals/referralsController.ts
```diff
@@ -130,7 +130,11 @@ export default class ReferralsController {
   async viewReferralForm(req: Request, res: Response): Promise<void> {
     const referral = await this.interventionsService.getDraftReferral(res.locals.user.token, req.params.id)
 
-    const presenter = new ReferralFormPresenter(referral)
+    const serviceCategoryName = referral.serviceCategoryId
+      ? (await this.interventionsService.getServiceCategory(res.locals.user.token, referral.serviceCategoryId)).name
+      : 'Accommodation'
```
lawrence-forooghian, line 135: Is there a real-life situation where `serviceCategoryId` wouldn't have been set on a draft referral at this stage (I think not)? If so, it might be clearer to throw an error here instead of an arbitrary name.

server/routes/referrals/referralFormPresenter.test.ts
```diff
@@ -108,7 +108,7 @@ describe('ReferralFormPresenter', () => {
           usingRarDays: true,
           maximumRarDays: 10,
         })
-        const presenter = new ReferralFormPresenter(referral)
+        const presenter = new ReferralFormPresenter(referral, 'Accommodation')
```
lawrence-forooghian, line 105: The rest of the application assumes that a `ServiceCategory`'s `name` is always in lower case - i.e. that it's suitable to drop into a sentence. I think we could do with verifying this in a contract test, actually.

server/routes/referrals/referralFormPresenter.ts
```diff
@@ -1,7 +1,7 @@
 import { DraftReferral } from '../../services/interventionsService'
 
 export default class ReferralFormPresenter {
-  constructor(private readonly referral: DraftReferral) {}
+  constructor(private readonly referral: DraftReferral, private readonly serviceCategory: string) {}
```
lawrence-forooghian, line 4: `serviceCategoryName`?

integration_tests/integration/referral.spec.js
```diff
@@ -156,6 +173,9 @@ describe('Referral form', () => {
     cy.get('h1').contains('Do you have further information for the accommodation service provider? (optional)')
     cy.get('textarea').type('Some information about Geoffrey')
 
+    // stub completed draft referral to mark section as completed
+    cy.stubGetDraftReferral(draftReferral.id, completedDraftReferral)
```
lawrence-forooghian, line 177: > This feels like a bit of a hack

I think it's fine!

lawrence-forooghian (commented):
Looks good, a few small comments 👍

server/routes/referrals/referralFormPresenter.test.ts
Gweaton: Good point. Have updated.

server/routes/referrals/referralsController.ts
Gweaton: Yep, good point, I think I did this out of laziness...

server/routes/referrals/referralFormPresenter.test.ts
Gweaton: Any suggestions for how to write a better contract test than the one we have? the current one here https://github.com/ministryofjustice/hmpps-interventions-ui/blob/e394d6a6bf2fea27c31c38bfe931cde11500db9d/server/services/interventionsService.test.ts#L862 kind of tests it, but not all that explicitly.

server/routes/referrals/referralFormPresenter.ts
Gweaton: 👍

server/routes/referrals/referralFormPresenter.test.ts
lawrence-forooghian: By "optional", I mean the fields that are nullable on the `ReferralFields` type, so that excludes e.g. `additionalNeedsInformation`. Also, I think that `usingRarDays` should be `false` and `maximumRarDays` `null` here.

server/routes/referrals/referralFormPresenter.test.ts
lawrence-forooghian: We could use a [regular expression matcher](https://docs.pact.io/getting_started/matching/#regular-expressions) on the `name` field in that test, to check that it starts with a capital letter (`/^[A-Z].*/`)?

lawrence-forooghian (commented):
Thanks for making the changes - just one thing that I wasn't very clear about (the optional fields thing).

server/routes/referrals/referralFormPresenter.test.ts
lawrence-forooghian: Sorry - I understand what's going on now we've spoken - just `usingRarDays` needs changing I think.

server/routes/referrals/referralFormPresenter.test.ts
Gweaton: Didn't know that was a thing, have added it - thanks!

server/routes/referrals/referralFormPresenter.test.ts
Gweaton: As discussed, we're leaving out the other "required" fields like `additionalNeedsInformation` as they live outside this intervention details section. Have updated `usingRarDays` though!

lawrence-forooghian (approved):

PR #93, opened 2021-02-01
IC-1012: Link intervention to draftReferral when creating

Gweaton:

Links an intervention to a draft referral on creation.

The `interventionId` will be hardcoded until we start work on the 'Find' journey, which
will pass the selected `interventionId` to the `#createDraftReferral`
method.

At the moment, we don't actually need to update the response, as the
'Refer' journey doesn't need anything from the intervention. We're
currently getting all the information we need from the service category.

This will allow us to store things like a `location` or its ID on the
intervention when we need to display it in future.

### DIFF
--- a/server/routes/referrals/referralsController.test.ts
+++ b/server/routes/referrals/referralsController.test.ts
@@ -67,19 +67,26 @@ describe('GET /referrals/start', () => {
 })
 
 describe('POST /referrals/start', () => {
-  describe('when searching for a CRN found in Delius', () => {
+  describe('when searching for a CRN found in Delius and an intervention has been selected', () => {
     beforeEach(() => {
       communityApiService.getServiceUserByCRN.mockResolvedValue(deliusServiceUser.build())
     })
 
     it('creates a referral on the interventions service and redirects to the referral form', async () => {
+      const currentlyHardcodedInterventionId = '98a42c61-c30f-4beb-8062-04033c376e2d'
+      const serviceUserCRN = 'X123456'
+
       await request(app)
         .post('/referrals/start')
-        .send({ 'service-user-crn': 'X123456' })
+        .send({ 'service-user-crn': serviceUserCRN })
         .expect(303)
         .expect('Location', '/referrals/1/form')
 
-      expect(interventionsService.createDraftReferral).toHaveBeenCalledTimes(1)
+      expect(interventionsService.createDraftReferral).toHaveBeenCalledWith(
+        'token',
+        serviceUserCRN,
+        currentlyHardcodedInterventionId
+      )
     })
 
     it('updates the newly-created referral on the interventions service with the found service user', async () => {
--- a/server/routes/referrals/referralsController.ts
+++ b/server/routes/referrals/referralsController.ts
@@ -59,6 +59,7 @@ export default class ReferralsController {
     let serviceUser: DeliusServiceUser | null = null
 
     const crn = req.body['service-user-crn']
+    const hardcodedInterventionId = '98a42c61-c30f-4beb-8062-04033c376e2d'
 
     if (form.isValid) {
       try {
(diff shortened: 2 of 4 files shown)


server/routes/referrals/referralsController.test.ts
```diff
@@ -67,19 +67,26 @@ describe('GET /referrals/start', () => {
...
 
 describe('POST /referrals/start', () => {
-  describe('when searching for a CRN found in Delius', () => {
+  describe('when searching for a CRN found in Delius and an intervention has been selected', () => {
     beforeEach(() => {
       communityApiService.getServiceUserByCRN.mockResolvedValue(deliusServiceUser.build())
     })
 
     it('creates a referral on the interventions service and redirects to the referral form', async () => {
+      const currentlyHardcodedInterventionId = '24b8cd48-ccb2-48e0-a842-f5673c73e38f'
```
tomsmyers, line 76: please can you use `98a42c61-c30f-4beb-8062-04033c376e2d` for this for now? it avoids us having to change it in the API test data

server/services/interventionsService.test.ts
```diff
@@ -191,22 +191,27 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
...
+        state: 'an intervention has been selected and a draft referral can be created',
         uponReceiving: 'a POST request to create a draft referral',
         withRequest: {
           method: 'POST',
           path: '/draft-referral',
           headers: { Accept: 'application/json', Authorization: `Bearer ${token}` },
-          body: { serviceUserCrn: 'X862134' },
+          body: {
+            serviceUserCrn: serviceUserCRN,
+            interventionId,
```
tomsmyers, line 206: this should also be a named field e.g. `interventionId: interventionId`

tomsmyers (commented):
couple small things

server/services/interventionsService.test.ts
Gweaton: This is JS property shorthand for the above as the key and variable name are the same (the linter complains otherwise, but I can rename the variable if you find it clearer, or we can turn off that linting rule?) I'll change it for the one above too if you're happy with it being shorthand (doesn't complain as `CRN` vs Crn`)

server/routes/referrals/referralsController.test.ts
Gweaton: Sure

server/services/interventionsService.test.ts
Gweaton: Have updated this to use property shorthand in both cases, but lemme know if you'd rather we didn't.

tomsmyers: oh cool, awesome thanks

server/routes/referrals/referralsController.ts
```diff
@@ -59,6 +59,7 @@ export default class ReferralsController {
     let serviceUser: DeliusServiceUser | null = null
 
     const crn = req.body['service-user-crn']
+    const hardcodedInterventionId = '98a42c61-c30f-4beb-8062-04033c376e2d'
```
tomsmyers, line 62: sorry that this is the second time i've commented a similar thing but

i've added a sample intervention into dev with id `3165ded2-e754-49d5-9c93-e8e063381c56`. so that's the one to use when making the real call. 

for the contract tests and running localy, the existing `98a42c61...` is the one.

i wanted them to be different to ensure there is a clear seperation of the data between local/dev. but it means this hardcoded value is gonne be wrong either for local or dev. i think it should be correct in dev. then we can always change it ourselves when running locally. 

what do you think?

Gweaton: hmm, I'd rather we used the same for dev and contract tests, so we can just get things up and running without remembering we'll need to change things locally, as I can imagine that becoming a bit of a pain. When you say you want there to be a separation of data between local and dev, could you see it becoming a problem if we were to re-use it across both?

tomsmyers: i think this issue only arises from the fact it's hard coded. in general the data will be totally seperate (apart from things that are common across all environments like service category ids). since it's an usual case i'm happy to change the id of the intervention in dev to match the local one.

tomsmyers: done

tomsmyers (approved):

PR #95, opened 2021-02-02
IC-1036 Move service provider to top level field on referral

Gweaton:

- Moves the `ServiceProvider` to the top level field on a referral.
- Removes `getServiceProvider` function, as we no longer need a Service Provider on its own.
- (Possibly controversial) removes `ServiceProviderFactory` for now, as we weren't using it anywhere and I don't feel it added much value, although we may want to add it back when we add more fields, but this won't be a large piece of work...

To reduce the number of requests to the InterventionsService - we no longer need to request a `ServiceProvider` by its `id`, as we can access it on the referral itself.

### DIFF
--- a/server/services/interventionsService.ts
+++ b/server/services/interventionsService.ts
@@ -202,15 +202,6 @@ export default class InterventionsService {
     })) as DraftReferral[]
   }
 
-  async getServiceProvider(token: string, id: string): Promise<ServiceProvider> {
-    const restClient = this.createRestClient(token)
-
-    return (await restClient.get({
-      path: `/service-provider/${id}`,
-      headers: { Accept: 'application/json' },
-    })) as ServiceProvider
-  }
-
   async sendDraftReferral(token: string, id: string): Promise<SentReferral> {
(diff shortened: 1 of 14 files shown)


server/services/interventionsService.ts
```diff
@@ -202,15 +202,6 @@ export default class InterventionsService {
     })) as DraftReferral[]
   }
 
-  async getServiceProvider(token: string, id: string): Promise<ServiceProvider> {
```
lawrence-forooghian, line 206: If you search for `stubGetServiceProvider`, there are some other stubs that need to be removed for this call, I think.

lawrence-forooghian (commented):
👍 code looks good to me, but I'll defer to a backend person (@tomsmyers?) on whether we're happy to make this change in the API.

server/services/interventionsService.ts
Gweaton: Ah, well remembered. Those are gone now

tomsmyers (approved):

PR #99, opened 2021-02-04
IC-1096 Add SU contact details to referral

Gweaton:

Adds service user's mobile and email to the referral when fetching the service user from the Community API. It currently doesn't display it anywhere.

Extracts Delius Service User serialization function into own class, as it was getting quite complicated in there.

So we can store the SU details on a referral and play it back to the Service Provider.


Gweaton:
We shouldn't be doing this, sorry! We should be fetching them on the receive side so they're up to date...

PR #101, opened 2021-02-05
IC-1097: Display Service User contact details on SP intervention view

Gweaton:

Adds a placeholder GOVUK Notification banner (needs futher design work) to the SP's interventions view, with data fetched from the Community API. This is currently hardcoded to Aadland Bertrand locally, which looks a bit weird on some pages, as the response from the community api doesn't match the mocked service user on our referral.

A couple of notes/questions:
- What happens if the community API is down and a Service Provider is trying to view this screen? Should we store a snapshot on the contact details on referral creation and fall back to this if that's the case, or would that confuse things?
- If we want to display this on another screen, it'd be good to move it into a partial, but I've gone for the simplest option for now as the Critical Outcomes approach.
- Again, the local instance of the community api has made testing this locally very difficult, as the data is a bit lacking (everything seems to come back empty, as you see in the screenshot below) - I'm going to have a chat with Matt R and see if he can help us update this.

To ensure we can show an up-to-date record of the Service User's contact details to the Service Provider.

## Screenshot (shows data discrepancy between local community API and mocks, but there's an example with Aadland Bertrand as the service user too)

image

### DIFF
--- a/server/routes/serviceProviderReferrals/showReferralPresenter.ts
+++ b/server/routes/serviceProviderReferrals/showReferralPresenter.ts
@@ -1,5 +1,6 @@
-import { DeliusUser } from '../../services/communityApiService'
+import { DeliusServiceUser, DeliusUser } from '../../services/communityApiService'
 import { SentReferral, ServiceCategory } from '../../services/interventionsService'
+import CalendarDay from '../../utils/calendarDay'
 import { SummaryListItem } from '../../utils/summaryList'
 import utils from '../../utils/utils'
 import ReferralDataPresenterUtils from '../referrals/referralDataPresenterUtils'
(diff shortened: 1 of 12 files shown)


server/routes/serviceProviderReferrals/showReferralPresenter.ts
```diff
@@ -21,4 +23,49 @@ export default class ShowReferralPresenter {
...
+  }
+
+  private get serviceUserEmail(): string {
+    const { emailAddresses } = this.serviceUser.contactDetails
+
+    if (emailAddresses && emailAddresses.length > 0) {
+      return emailAddresses[0]
+    }
+
+    return 'Email address not found'
```
tomsmyers, line 56: why the difference in 'not found' behaviour for email? the other two use `notFoundMessage`

tomsmyers:
it's good we get these contact details without any extra API calls. regarding what we do if community API is down - in all honesty, i don't think our service will function at all in this case. PP's won't even be able to sign in. so i don't think we need a fallback for this.

tomsmyers (approved):
LGTM

PR #102, opened 2021-02-08
IC-1057 - session progress table

isratc:
**Jira card URL:**
https://dsdmoj.atlassian.net/browse/IC-1052
## Changes in this PR:
Added in the table and tag component, And styled it as per design
## Screenshots of UI changes:
Screenshot 2021-02-08 at 12 00 01

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6052,9 +6052,9 @@
       }
     },
     "govuk-frontend": {
-      "version": "3.9.1",
-      "resolved": "https://registry.npmjs.org/govuk-frontend/-/govuk-frontend-3.9.1.tgz",
-      "integrity": "sha512-ouOoDUj0QwDA4uCHIBkGCFMpORuTRcSuDscOrz7V1PBcOecntLglxJAZAuNm+j2sPo7anoScHU0ZSeE2QIoeAg=="
+      "version": "3.10.2",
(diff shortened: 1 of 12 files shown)


package-lock.json
```diff
@@ -1,8 +1,15953 @@
 {
   "name": "hmpps-interventions-ui",
   "version": "0.0.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
Gweaton, line 4: Do you think you could remove this change (the whole `package-lock`) from the commit? It probably did this by itself when we ran `npm install` earlier. What do you get when you run `npm -v` in the project?

I think this is something we should make standard on the project, or find a workaround for, as I just tried to use the latest verson and it updated `package.lock`. Probably best to fix that in another PR though.

assets/sass/components/_badges-tags.scss
```diff
@@ -0,0 +1,12 @@
+a.govuk-link {
+  font-weight: 400 !important;
```
Gweaton, line 2: I don't think you need the `!important` here - it seems to work without (at least, for me, locally) - would you mind trying without? It's best to try and avoid using `!important` where possible as it can lead to some confusing behaviour when trying to add new styles etc.

PR #103, opened 2021-02-08
add username claim to token used for testing

tomsmyers:

add username claim to token used for testing

ensures that the username claim is available to the backend when running pact tests.

### DIFF
--- a/testutils/factories/oauth2Token.ts
+++ b/testutils/factories/oauth2Token.ts
@@ -4,6 +4,7 @@ import { Factory } from 'fishery'
 interface TokenParams {
   authSource: string
   userID: string
+  username: string
   roles: string[]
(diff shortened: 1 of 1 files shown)


Gweaton (commented):
Most annoying thing ever, but is it a username, or the user's name? If the first, I'd suggest all lowercase, but if it's the second I'd say camel case (like you've done).

tomsmyers:
> Most annoying thing ever, but is it a username, or the user's name? If the first, I'd suggest all lowercase, but if it's the second I'd say camel case (like you've done).

hmm. well it is the former. but the claim is named `user_name` in the real tokens ...

Gweaton (approved):

rgforsyth (approved):

PR #104, opened 2021-02-08
IC-1057 - Session progress table

Gweaton:

**Jira card URL:**
https://dsdmoj.atlassian.net/browse/IC-1052

## Changes in this PR:
Added in the table and tag component, and styled it as per design

## Screenshots of UI changes:
Screenshot 2021-02-08 at 12 00 01


assets/sass/components/_badges-tags.scss
```diff
@@ -0,0 +1,12 @@
+a.govuk-link {
+  font-weight: 400;
```
lawrence-forooghian, line 2: Perhaps I'm missing something, but does this change the weight of all of the links in the app? Is that the intention?

isratc: Yes it was meant to be wrapped around the table tag we had a bit of a git mashup yesterday forgot to add in

assets/sass/components/_badges-tags.scss
```diff
@@ -1,8 +1,7 @@
-a.govuk-link {
-  font-weight: 400;
-}
-
 .govuk-table {
+  a.govuk-link {
+    font-weight: 400;
```
Gweaton, line 3: Hey, in your other PR I saw you put: `font-weight: normal; ` - I didn't know it was a thing, but I think it's nicer than using the magic number of `400` - do you think you'd adding it here too? It's also nice to have consistency across the styles.

isratc: sure, but it's truly not a big deal

isratc (approved):

PR #107, opened 2021-02-08
IC-1067: Add contract for `GET /interventions`

lawrence-forooghian:
## What does this pull request do?

Adds the `Intervention` type and a contract for the `GET /interventions` endpoint proposed in ministryofjustice/hmpps-interventions-service#87.

## What is the intent behind these changes?

To allow us to display a list of interventions.

PR #108, opened 2021-02-08
IC-1067: Show a list of all interventions

lawrence-forooghian:

Adds a page for displaying a list of all of the interventions.

To allow a probation practitioner to see all of the available interventions, and to provide a starting point for us to build the search functionality.

## Why is this a draft?

1. It's based on top of #107, so that needs to be merged first.
2. I have an outstanding question about where the "Region" should come from (see inline comment).

## Screenshot

Screenshot_2021-02-08 HMPPS Interventions - GOV UK

### DIFF
--- a/server/routes/findInterventions/searchResultsPresenter.test.ts
+++ b/server/routes/findInterventions/searchResultsPresenter.test.ts
@@ -1,0 +1,145 @@
+import { DeepPartial } from 'fishery'
+import SearchResultsPresenter from './searchResultsPresenter'
+import interventionFactory from '../../../testutils/factories/intervention'
+import eligibilityFactory from '../../../testutils/factories/eligibility'
+import { Intervention } from '../../services/interventionsService'
+
+describe(SearchResultsPresenter, () => {
+  describe('results', () => {
+    it('contains as many items as there are interventions', () => {
+      const presenter = new SearchResultsPresenter(interventionFactory.buildList(3))
+
+      expect(presenter.results.length).toBe(3)
+    })
+
+    function linesForKey(key: string, params: DeepPartial<Intervention>): string[] {
+      const presenter = new SearchResultsPresenter([interventionFactory.build(params)])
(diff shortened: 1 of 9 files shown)


server/routes/findInterventions/searchResultsPresenter.test.ts
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    describe('Type', () => {
+      it('is “Dynamic Framework”', () => {
+        expect(linesForKey('Type', {})).toEqual(['Dynamic Framework'])
+      })
+    })
+
+    describe('Region', () => {
+      it('is “TODO”', () => {
```
lawrence-forooghian, line 31: In the [designs](https://www.figma.com/file/hvnANWQPHEornIY2jFXocu/BETA-Single-role-user-for-dashboard%2Fintervention-progress?node-id=2173%3A28519), this is, for example, "North East" or "Yorkshire and the Humber". I don't think we have anything in the `Intervention` DTO that corresponds to these, which are perhaps the NPS regions?

Gweaton: Ming's let me know that the design was from before we decided to use PCC regions, so I've removed the Region for now - am waiting for final clarification from Rhiannon/Tom but I think we can always add it back in when we know more.

server/routes/findInterventions/searchResultsPresenter.ts
```diff
@@ -0,0 +1,80 @@
...
+  summary: SummaryListItem[]
+}
+
+export default class SearchResultsPresenter {
+  constructor(private readonly interventions: Intervention[]) {}
+
+  readonly results: SearchResultPresenter[] = this.interventions.map(intervention => ({
+    title: intervention.title,
+    href: '#',
+    body: intervention.description,
```
Gweaton, line 18: This was hardcoded to a string description before (I force pushed before leaving a comment, but have updated it, and the mocks).

Gweaton (approved):

PR #109, opened 2021-02-09
IC-1095-safeguarding-issue-form

isratc:
**Jira card URL:**
https://dsdmoj.atlassian.net/browse/IC-1095

##Changes in this PR:
Adding in the template for the radio buttons m text area and save and continue button in the two-third layout and style it as per design
**This has been checked for mobile responsive**

##Screenshots of UI changes:
Screenshot 2021-02-09 at 11 03 53


Gweaton (approved):
This looks good, nice work.

One thing I would say, is in your commit messages, it's helpful to have the main thing the commit introduces, rather than the jira issue name/number, as that's already on the branch name. An example of what that could look like here is:

> Update safeguarding issue form to match latest designs
> 
> - Add radio buttons
> - Add text area
> - Make sure label is not bold
> - Add save and continue button
>
> _Optional: add a link to the designs/jira issue_

This makes it a bit easier to read in a git log, as you can see quickly what each commit did.

PR #111, opened 2021-02-09
WIP Filter component

Gweaton:

Adds a very rough filter spike, essentially a copy and paste of https://github.com/alphagov/digitalmarketplace-govuk-frontend/tree/master/src/digitalmarketplace/components/option-select

I've struggled with this for half a day, it feels like it's almost there but for some reason the "closed" behaviour doesn't work. I'd like to dig into why, but I feel like I've already spent too much time on this.

This code I've copied from the other team is old-style JS (`assets/js/optionSelect.js`), rather than the ES6 we're using, so it would be best to re-write this ourselves, as I haven't had time to dig into how it works too much.

To see how long it would take to build a filter component.

## Screenshot

image

## Conclusion

I feel like this component is a little bit complex to use without spending more time looking into how it works, as there's a lot of client side code going on. It might be simpler to re-style the [MOJ Filter component](https://moj-design-system.herokuapp.com/components/filter), or build a slimmed down version ourselves.

### DIFF
--- a/assets/js/optionSelect.js
+++ b/assets/js/optionSelect.js
@@ -1,0 +1,188 @@
+function OptionSelect($module) {
+  this.$optionSelect = $module
+  this.$options = this.$optionSelect.querySelectorAll("input[type='checkbox']")
+  this.$optionsContainer = this.$optionSelect.querySelector('.js-options-container')
+  this.$optionList = this.$optionsContainer.querySelector('.js-auto-height-inner')
+  this.$allCheckboxes = this.$optionsContainer.querySelectorAll('.govuk-checkboxes__item')
+  this.checkedCheckboxes = []
+}
+
+OptionSelect.prototype.init = function () {
+  // Attach listener to update checked count
+  this.$optionSelect.addEventListener(
+    'change',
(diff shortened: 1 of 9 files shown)


lawrence-forooghian:
> This code I've copied from the other team is old-style JS (assets/js/optionSelect.js), rather than the ES6 we're using

Just a thought on this - the modern JS that we write is only backend code. If we wanted to write modern JS for frontend, wouldn't we need to set up some sort of transpilation to make sure we support a good range of browsers?

Gweaton:
> > This code I've copied from the other team is old-style JS (assets/js/optionSelect.js), rather than the ES6 we're using
> 
> Just a thought on this - the modern JS that we write is only ~backend~ server-side code. If we wanted to write modern JS for ~frontend~ execution in the browser, wouldn't we need to set up some sort of transpilation to make sure we support a good range of browsers?

Yes, very good point. We'll probably need to just write plain JS for this anyway.

Gweaton:
Closing this as we've built our own in https://github.com/ministryofjustice/hmpps-interventions-ui/pull/128

PR #116, opened 2021-02-11
IC-1158 - Add service provider template and layout

Gweaton:

Adds a new template, replacing the standard GOV UK template as our Service Provider sessions pages have a different layout. This will allow us to extend the layout on those pages, and override the various blocks with the components that Israt will be adding content for.

An example template has been added to the `static-pages` index to demonstrate how you'd use these templates.

## Screenshot of example page (ugly text where partials will be added)

image

### DIFF
--- a/server/views/serviceProviderReferrals/partials/layoutAllComponents.njk
+++ b/server/views/serviceProviderReferrals/partials/layoutAllComponents.njk
@@ -1,0 +1,47 @@
+{% extends "../templates/serviceProviderSessionsTemplate.njk" %}
+{% from "govuk/components/phase-banner/macro.njk" import govukPhaseBanner %}
+
+{% block head %}
+  <!--[if !IE 8]><!-->
+  <link href="/assets/stylesheets/application.css?{{ version }}" rel="stylesheet"/>
+  <!--<![endif]-->
+
+  <!--[if lt IE 9]> <link href="/assets/stylesheets/application-ie8.css?{{ version }}" rel="stylesheet"/> <script src="/assets/js/html5shiv-3.7.3.min.js"></script> <![endif]-->
+
+  <script src="/assets/js/jquery.min.js"></script>
+  <script src="https://code.jquery.com/ui/1.12.1/jquery-ui.min.js" integrity="sha256-VazP97ZCwtekAsvgPBSUwPFKdrwD3unUfSGVYrahUqU=" crossorigin="anonymous"></script>
+  <link href="https://code.jquery.com/ui/1.12.1/themes/ui-lightness/jquery-ui.css" rel="stylesheet">
+
+{% endblock %}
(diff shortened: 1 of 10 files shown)


server/views/serviceProviderReferrals/sessions/layoutExample.njk
```diff
@@ -0,0 +1,289 @@
+{% from "govuk/components/table/macro.njk" import govukTable %}
+{% from "govuk/components/tag/macro.njk" import govukTag %}
+
+{% extends "../partials/layoutAllComponents.njk" %}
```
Gweaton, line 4: @isratc here's how you'd add the layout to a view

server/views/serviceProviderReferrals/sessions/layoutExample.njk
```diff
@@ -0,0 +1,289 @@
...
+{% block primaryTitle %}
+    <h1 class="govuk-heading-l">This is a primary title example</h2>
+{% endblock %}
+
+
+{% block pageHeading %}
+  <h2 class="govuk-heading-l">Track progress of each intervention</h2>
+{% endblock %}
+
+{% block mainContent %}
```
Gweaton, line 18: and here's `mainContent` (the original templates just use `content` but now we're adding a `mainContent` section and `sidebar`, as they need to both be part of the main page's content.

server/views/serviceProviderReferrals/templates/serviceProviderSessionsTemplate.njk
```diff
@@ -0,0 +1,68 @@
+{% from "govuk/components/skip-link/macro.njk" import govukSkipLink -%}
```
lawrence-forooghian, line 1: Is this `serviceProviderSessionsTemplate.njk` template a modified copy of an existing one? If so, it would be super-useful to mention where it's copied from in the commit message.

server/views/serviceProviderReferrals/partials/layoutAllComponents.njk
```diff
@@ -0,0 +1,47 @@
+{% extends "../templates/serviceProviderSessionsTemplate.njk" %}
```
lawrence-forooghian, line 1: I don't think I understand why we have two files to create this layout. Would you mind explaining the difference between `partials/layoutAllComponents` and `templates/serviceProviderSessionsTemplate.njk`? Could we not just put all of this into the template file?  I thought that a partial is a component that you drop into a larger page, not something that defines the layout of the entire page.

isratc:
Thanks George, once a refined approach is done, and are added to that /staticpages link. please give me a shout and you can walk me through it.

Gweaton:
Closing this as we've got a rough version of it working for the action plan journey to use as an example.

PR #118, opened 2021-02-12
IC-1058/Review-session–form

isratc:
https://dsdmoj.atlassian.net/browse/IC-1058

## Changes in this PR:

- Create a custom syntax that has change links as a summary list component would not be suitable as we are change linking   "pages to change" rather than a single question.
- you will see that I've added dummy text to direct to you when you add in the variables that pull in the answers from the form
- Two layout
- Submit button
- tested/styled for mobile responsive

## Screenshots of UI changes:
Screenshot 2021-02-12 at 14 44 16

### DIFF
--- a/assets/sass/components/_form-fields.scss
+++ b/assets/sass/components/_form-fields.scss
@@ -1,8 +1,28 @@
 .govuk-form-group {
   label.govuk-label.govuk-label--s {
     font-weight: normal;
-    &.strong {
-      font-weight: bold;
-    }
   }
 }
+.govuk-review__list {
+  .change__right {
+    @include govuk-media-query($from: tablet) {
+      float: right;
+    }
+    @include govuk-media-query($from: mobile) {
+      @include govuk-responsive-margin(6, "bottom");
+      display: block;
+    }
+  }
+  .govuk-grid-column-three-quarters, .govuk-grid-column-two-thirds, .govuk-grid-column-full, 
+  .govuk-grid-column-two-third, .govuk-grid-column-one-third {
+    padding: 0px;
+  }
+  .govuk-summary-list__key {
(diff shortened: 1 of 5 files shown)


assets/sass/components/_form-fields.scss
```diff
@@ -3,3 +3,26 @@
...
   }
 }
+.govuk-review__list {
+  .change__right {
+    @include govuk-media-query($from: tablet) {
+      float:right;
+    }
+    @include govuk-media-query($from: mobile) {
+      @include govuk-responsive-margin(6, "bottom");
+      display:block
```
Gweaton, line 13: Missing `;` here (would be caught by linting)

assets/sass/components/_review-session-form.scss
```diff
@@ -0,0 +1,19 @@
+.govuk-review__list {
+  .right {
+    @include govuk-media-query($until: tablet) {
+      float:right;
```
Gweaton, line 4: See comment on other file re: spaces after colons

assets/sass/components/_badges-tags.scss
```diff
@@ -1,11 +1,11 @@
 .govuk-table {
   a.govuk-link {
-    font-weight: normal;
+    font-weight: normal ;
```
Gweaton, line 3: extra space here before the `;` - my previous comment about adding a space was about adding a space between the property (`font-weight:`) and the value: (`normal`), apologies if that wasn't clear. It looks like you've got this in the other files too. Would you mind updating those too?

So just as an example, you'd have:
`white-space: nowrap;`

Gweaton (approved):
Thanks for taking care of all the styles!

PR #119, opened 2021-02-12
IC-1074: Add intervention details page

Gweaton:

Adds intervention details page, accessible from Find search results.

Currently doesn't include the miscellaneous bullet points present in the design, as we're unsure exactly where they're from.

In doing this, extracts `interventionDetails` into own presenter

Also renames "Sex" to "Gender" in the Find journey.

## Screenshot
image

### DIFF
--- a/server/routes/findInterventions/findInterventionsController.test.ts
+++ b/server/routes/findInterventions/findInterventionsController.test.ts
@@ -49,4 +49,23 @@ describe(FindInterventionsController, () => {
         })
     })
   })
+
+  describe('GET /find-interventions/intervention/:id', () => {
+    it('responds with a 200', async () => {
+      const intervention = interventionFactory.build({
+        title: 'Better solutions (anger management)',
+        serviceCategory: { name: 'thinking and behaviour' },
+      })
+
+      interventionsService.getIntervention.mockResolvedValue(intervention)
+
+      await request(app)
+        .get(`/find-interventions/intervention/${intervention.id}`)
+        .expect(200)
+        .expect(res => {
+          expect(res.text).toContain('Better solutions (anger management)')
+          expect(res.text).toContain('Thinking and behaviour')
+        })
+    })
+  })
 })
--- a/server/routes/findInterventions/interventionDetailsView.ts
+++ b/server/routes/findInterventions/interventionDetailsView.ts
@@ -1,6 +1,7 @@
 import ViewUtils from '../../utils/viewUtils'
 import { SummaryListArgs, SummaryListItem } from '../../utils/summaryList'
 import InterventionDetailsPresenter from './interventionDetailsPresenter'
+import { TabsArgs } from '../../utils/govukFrontendTypes'
 
 export default class InterventionDetailsView {
   constructor(private readonly presenter: InterventionDetailsPresenter) {}
@@ -9,10 +10,28 @@ export default class InterventionDetailsView {
     return { ...ViewUtils.summaryListArgs(items), classes: 'govuk-summary-list--no-border' }
   }
 
+  private tabsArgs(summaryListMacro: (args: unknown) => string): TabsArgs {
+    return {
+      items: this.presenter.tabs.map(tab => {
+        return {
+          label: tab.title,
+          id: tab.id,
+          panel: {
+            html: summaryListMacro(InterventionDetailsView.summaryListArgs(tab.items)),
+          },
+        }
+      }),
+    }
+  }
+
   get renderArgs(): [string, Record<string, unknown>] {
     return [
       'findInterventions/interventionDetails',
-      { presenter: this.presenter, summaryListArgs: InterventionDetailsView.summaryListArgs },
+      {
+        presenter: this.presenter,
+        summaryListArgs: InterventionDetailsView.summaryListArgs,
+        tabsArgs: this.tabsArgs.bind(this),
+      },
     ]
   }
 }
--- a/server/routes/findInterventions/interventionDetailsView.ts
+++ b/server/routes/findInterventions/interventionDetailsView.ts
@@ -1,0 +1,18 @@
+import ViewUtils from '../../utils/viewUtils'
+import { SummaryListArgs, SummaryListItem } from '../../utils/summaryList'
+import InterventionDetailsPresenter from './interventionDetailsPresenter'
+
+export default class InterventionDetailsView {
+  constructor(private readonly presenter: InterventionDetailsPresenter) {}
+
+  static summaryListArgs(items: SummaryListItem[]): SummaryListArgs & { classes: string } {
+    return { ...ViewUtils.summaryListArgs(items), classes: 'govuk-summary-list--no-border' }
+  }
+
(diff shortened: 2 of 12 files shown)


server/routes/findInterventions/searchResultsPresenter.ts
```diff
@@ -1,75 +1,12 @@
...
-import utils from '../../utils/utils'
-
-export interface SearchResultPresenter {
-  title: string
-  href: string
-  body: string
-  summary: SummaryListItem[]
-}
+import { Intervention } from '../../services/interventionsService'
+import InterventionDetailsPresenter, { InterventionDetails } from './InterventionDetailsPresenter'
```
lawrence-forooghian, line 2: The file name (and the test file name) should be `lowerCamelCase`.

server/routes/findInterventions/InterventionDetailsPresenter.ts
```diff
@@ -0,0 +1,82 @@
...
+      },
+      {
+        key: 'Sex',
+        lines: [this.sexDescription(this.intervention.eligibility)],
+        isList: false,
+      },
+    ]
+  }
+
+  private ageGroupDescription(eligibility: Eligibility): string {
```
lawrence-forooghian, line 72: Is there a reason you changed these from `static` methods? I usually use `static` to make it clear at a glance that the behaviour / return value of a method isn't affected by any instance state.

server/routes/findInterventions/InterventionDetailsPresenter.ts
```diff
@@ -0,0 +1,82 @@
+import { Eligibility, Intervention } from '../../services/interventionsService'
+import { SummaryListItem } from '../../utils/summaryList'
+import utils from '../../utils/utils'
+
+export interface InterventionDetails {
```
lawrence-forooghian, line 5: What is the intention of this type? Would it be a problem for `SearchResultsPresenter`'s `results` method to return `InterventionDetailsPresenter[]`?

server/routes/findInterventions/findInterventionsController.test.ts
```diff
@@ -49,4 +49,22 @@ describe(FindInterventionsController, () => {
         })
     })
   })
+
+  describe('GET /find-interventions/intervention/:id', () => {
+    it('responds with a 200', async () => {
+      const intervention = interventionFactory.build({
+        title: 'Better solutions (anger management)',
+        serviceCategory: { name: 'thinking and behaviour' },
```
lawrence-forooghian, line 57: Do we need to specify a service category name if we're not making any assertions that depend on it?

server/routes/findInterventions/interventionDetailsView.ts
```diff
@@ -9,10 +10,28 @@ export default class InterventionDetailsView {
     return { ...ViewUtils.summaryListArgs(items), classes: 'govuk-summary-list--no-border' }
   }
 
+  tabsArgs(summaryListMacro: (args: unknown) => string): TabsArgs {
```
lawrence-forooghian, line 13:
```suggestion
  private tabsArgs(summaryListMacro: (args: unknown) => string): TabsArgs {
```

server/routes/findInterventions/InterventionDetailsPresenter.test.ts
```diff
@@ -38,6 +39,30 @@ describe(InterventionDetailsPresenter, () => {
...
   })
 
+  describe('tabs', () => {
+    const presenter = new InterventionDetailsPresenter(
+      interventionFactory.build({
+        serviceProvider: serviceProviderFactory.build({ name: 'Harmony Living' }),
+      })
+    )
+
+    it('returns an array of service provider fields to pass to the tab args, including a summary list', () => {
```
lawrence-forooghian, line 49: I'm not sure what "to pass to the tab args" means here - are you referring to the `InterventionDetailsView.prototype.tabsArgs` method? If so, I don't think that a presenter would be expected to know about that method (view knows about presenter, not vice-versa).

How about just

```suggestion
    it('returns an array of summary lists, each with a title', () => {
```

server/utils/tabs.ts
```diff
@@ -0,0 +1,12 @@
+export interface TabsArgs {
```
lawrence-forooghian, line 1: This is a TypeScript type for the object that we pass to the `govukTabs` macro - a subset of the "Nunjucks macro options" table [here](https://design-system.service.gov.uk/components/tabs/) - is that right?

I think it's an interesting idea to have strong types for these things - reduces the changes of an accidental typo. We haven't done it in any of other views so far so it would probably be good to create a tech debt for doing it consistently.

I wonder about the naming, though. I think we could be more explicit that these are related to GOV.UK Frontend macros. Perhaps `server/utils/govukFrontendTypes.ts`, and pop everything inside a [`namespace GovukFrontend`](https://www.typescriptlang.org/docs/handbook/namespaces.html)?

lawrence-forooghian (commented):
Nice work - a few comments but looks good.

server/routes/findInterventions/searchResultsPresenter.ts
Gweaton: Oops, don't know how I did this, I blame Fridays.

server/routes/findInterventions/InterventionDetailsPresenter.ts
Gweaton: Yeah I think this is due to the first way I implemented this (badly) then I refactored it and must have forgotten to update this - well spotted.

server/routes/findInterventions/findInterventionsController.test.ts
Gweaton: Another case where I'd like to blame it being a Friday - I meant to add an assertion for this... Have done so

server/routes/findInterventions/InterventionDetailsPresenter.ts
Gweaton: Yeah, something about calling static methods inside instance methods felt a bit off to me, I'm not sure why, as we call stuff like `Thing.new` all the time in Ruby. I'll change this.

server/routes/findInterventions/InterventionDetailsPresenter.test.ts
Gweaton: Yeah, really wasn't sure about this - have taken your suggestion

server/utils/tabs.ts
Gweaton: Yeah, that's right. I've moved this into `server/utils/govukFrontendTypes.ts` for our benefit, so we can re-use it, and will add an issue to the tech debt list.

It looks like namespaces are no longer the "done" thing in modern Typescript ([at least according to the ESLint folks](https://github.com/typescript-eslint/typescript-eslint/blob/v4.6.1/packages/eslint-plugin/docs/rules/no-namespace.md)), so I've left it as a place where we can drop these things as exportable interfaces for now, but let me know if there's anything else you'd prefer.

lawrence-forooghian (approved):

PR #121, opened 2021-02-12
IC-1089: Ensure `desired-outcomes-ids` is always an array

Gweaton:

Ensures `desired-outcomes-ids` are an array, as the backend was being sent a string (rather than an array, which it expected) when a single Desired Outcome checkbox was selected, and didn't know how to handle it.

To allow users to be able to select a single desired outcome without breaking the user journey :trollface:

### DIFF
--- a/server/routes/referrals/desiredOutcomesView.ts
+++ b/server/routes/referrals/desiredOutcomesView.ts
@@ -7,7 +7,7 @@ export default class DesiredOutcomesView {
   get checkboxArgs(): Record<string, unknown> {
     return {
       idPrefix: 'desired-outcomes-ids',
-      name: 'desired-outcomes-ids',
+      name: 'desired-outcomes-ids[]',
       fieldset: {
         legend: {
           text: this.presenter.title,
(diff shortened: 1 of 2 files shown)


server/routes/referrals/desiredOutcomesView.ts
```diff
@@ -7,7 +7,7 @@ export default class DesiredOutcomesView {
   get checkboxArgs(): Record<string, unknown> {
     return {
       idPrefix: 'desired-outcomes-ids',
-      name: 'desired-outcomes-ids',
+      name: 'desired-outcomes-ids[]',
```
lawrence-forooghian, line 10: We should update the [controller tests](https://github.com/ministryofjustice/hmpps-interventions-ui/blob/b96b8fccceb9e22c19fa0a5d9e6478141c47fb13/server/routes/referrals/referralsController.test.ts#L572-L606) too, I think.

lawrence-forooghian (changes requested):

lawrence-forooghian (approved):

PR #126, opened 2021-02-16
IC-1113 filter sent referrals by SP organization

tomsmyers:

passes service provider organization query param to 'get sent referrals' endpoint

only show the cases relevent to the SP staff member in their dashboard.

### DIFF
--- a/server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
+++ b/server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
@@ -13,7 +13,10 @@ export default class ServiceProviderReferralsController {
   ) {}
 
   async showDashboard(req: Request, res: Response): Promise<void> {
-    const referrals = await this.interventionsService.getSentReferrals(res.locals.user.token)
+    const referrals = await this.interventionsService.getSentReferrals(
+      res.locals.user.token,
+      res.locals.user.organizations[0].code
+    )
 
     const dedupedServiceCategoryIds = Array.from(
       new Set(referrals.map(referral => referral.referral.serviceCategoryId))
--- a/mockApis/interventionsService.ts
+++ b/mockApis/interventionsService.ts
@@ -120,7 +120,7 @@ export default class InterventionsServiceMocks {
     return this.wiremock.stubFor({
       request: {
         method: 'GET',
-        urlPattern: `${this.mockPrefix}/sent-referrals`,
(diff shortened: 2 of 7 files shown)


server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
```diff
@@ -13,7 +13,10 @@ export default class ServiceProviderReferralsController {
   ) {}
 
   async showDashboard(req: Request, res: Response): Promise<void> {
-    const referrals = await this.interventionsService.getSentReferrals(res.locals.user.token)
+    const referrals = await this.interventionsService.getSentReferrals(
+      res.locals.user.token,
+      res.locals.user.organizations[0].code
```
tomsmyers, line 18: i think this raises a question about what we do with this list of orgs..

@sldblog are all the questions about primes/subs answered to the point we can make some decisions about org level access now?

lawrence-forooghian:
@tomsmyers should we make this a draft until the data is set up, so we don't accidentally merge it?

server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
sldblog: Currently, we can assume
- one provider per user
- more than one provider: 🚫 invalid state

After onboarding subcontractors (let's not worry about this now, though)
- non-admin users: one organisation per user
- admin users: many organisations per user
- admin users: one _prime_ organisation per user, more than one prime: 🚫 invalid state

Does this help?

sldblog: For now, it would be good to explicitly flop the login if we have more than 1 org than rely on potential ordering chaos 😅

sldblog: However, longer-term:

- each user will have `0..*` contracts assigned to them
- we will need to get the sent referrals only for those specific contract(s)
- intersected by their providers

sldblog: Is it possible to completely defer the scoping to the API based on the user token?

tomsmyers: done. closing this pR

tomsmyers: https://github.com/ministryofjustice/hmpps-interventions-service/pull/107

PR #128, opened 2021-02-16
IC-1063 to 1065, IC-1033: A probation practitioner can filter the list of intervention search results

lawrence-forooghian:

Adds filtering controls to the list of intervention search results, and a summary of the filters chosen.

To allow a probation practitioner to narrow down the search results.

## Screenshot

image

### DIFF
--- a/server/routes/findInterventions/interventionsFilter.ts
+++ b/server/routes/findInterventions/interventionsFilter.ts
@@ -1,0 +1,46 @@
+import { Request } from 'express'
+import { InterventionsFilterParams } from '../../services/interventionsService'
+
+export default class InterventionsFilter {
+  pccRegionIds: string[] | undefined
+
+  gender: ('male' | 'female')[] | undefined
+
+  age: '18-to-25-only'[] | undefined
+
+  static fromRequest(request: Request): InterventionsFilter {
+    const filter = new InterventionsFilter()
+
+    filter.pccRegionIds = request.query['pcc-region-ids'] as string[] | undefined
+    filter.gender = request.query.gender as ('male' | 'female')[] | undefined
+    filter.age = request.query.age as '18-to-25-only'[] | undefined
+
+    return filter
+  }
+
+  get params(): InterventionsFilterParams {
+    const params: InterventionsFilterParams = {}
+
+    if (this.pccRegionIds !== undefined) {
+      params.pccRegionIds = this.pccRegionIds
+    }
+
+    if (this.gender !== undefined) {
+      if (this.gender.includes('male')) {
+        params.allowsMale = true
+      }
+
+      if (this.gender.includes('female')) {
+        params.allowsFemale = true
+      }
+    }
+
+    if (this.age !== undefined) {
+      if (this.age.includes('18-to-25-only')) {
+        params.maximumAge = 25
+      }
+    }
+
+    return params
+  }
+}
--- a/server/routes/findInterventions/searchResultsView.ts
+++ b/server/routes/findInterventions/searchResultsView.ts
@@ -5,6 +5,48 @@ import { SummaryListArgs, SummaryListItem } from '../../utils/summaryList'
 export default class SearchResultsView {
   constructor(private readonly presenter: SearchResultsPresenter) {}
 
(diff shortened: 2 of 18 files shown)


server/routes/findInterventions/interventionsFilter.ts
```diff
@@ -0,0 +1,46 @@
+import { Request } from 'express'
+import { InterventionsFilterParams } from '../../services/interventionsService'
+
+export default class InterventionsFilter {
+  pccRegionIds: string[] | undefined
+
+  gender: ('male' | 'female')[] | undefined
```
Gweaton, line 7: No biggie (and something we can do after this is merged if you think it's a good idea), but could be worth storing these in an exported enum, so we could use them all over the codebase?

server/routes/findInterventions/searchResultsView.ts
```diff
@@ -5,14 +5,62 @@ import { SummaryListArgs, SummaryListItem } from '../../utils/summaryList'
 export default class SearchResultsView {
   constructor(private readonly presenter: SearchResultsPresenter) {}
 
+  private checkboxArgs({
+    name,
+    legend,
+    checkboxes,
+  }: {
+    name: string
```
Gweaton, line 13: Probably one to add to the `govuk-frontend-types` that we introduced in a recent PR if we're keen to go down the route of strong types here, but we've got that covered in [this jira issue](https://dsdmoj.atlassian.net/browse/IC-1173?atlOrigin=eyJpIjoiODgwMDczYzBhZjQ1NGY5YWFhMmNjZjVmODdhYmM1MzUiLCJwIjoiaiJ9)

server/routes/findInterventions/searchSummaryPresenter.test.ts
```diff
@@ -0,0 +1,58 @@
...
+  describe('summary', () => {
+    describe('with an empty filter', () => {
+      it('returns no summary list items', () => {
+        const presenter = new SearchSummaryPresenter(new InterventionsFilter(), [])
+
+        expect(presenter.summary).toEqual([])
+      })
+    })
+
+    function linesForKey(key: string, args: [InterventionsFilter, PCCRegion[]]): string[] {
```
Gweaton, line 16: I know the args types are slightly different here, but might be worth extracting the `linesForKey` function in `interventionDetailsPresenter` for this to some kind of "test helper" as this is very similar (on a quick glance, I could be wrong though...). Again, we can do this later, but if so, let's add a card for it.

Gweaton (approved):
Nice work 👏 

A few comments, but nothing serious, and probably stuff we can leave til after Show & Tell if you agree with any of it.

server/routes/findInterventions/interventionsFilter.ts
lawrence-forooghian: It's not yet clear to me what the enum would represent. It's not "gender", it's just "the options that we display under ‘gender’ on the filter list". This type is quite specific to that particular set of filter options. If it becomes clear that those filter options need to be representable outside of one screen or something, then we can think about extracting them.

server/routes/findInterventions/searchResultsView.ts
lawrence-forooghian: Good point! I'll leave it until we work on that.

server/routes/findInterventions/interventionsFilter.ts
Gweaton: Sorry, when I say "all over the codebase" I think I probably just meant this file (I saw a lot of 'male' and 'female' in the tests, so thought it'd make it more standardised, but it's probably not right to use this in the tests). Probably overkill, and you're right, it doesn't really represent "gender" itself.

server/routes/findInterventions/searchSummaryPresenter.test.ts
lawrence-forooghian: Good shout - [done](https://github.com/ministryofjustice/hmpps-interventions-ui/pull/128/commits/91b6079c96d90732621d3f52c3f6a075ab550798).

PR #131, opened 2021-02-17
IC-1156: Add NPS Region to intervention

Gweaton:

Adds Region line to Intervention details summary in the Find journey. This isn't filterable, just for the PP's information.

To display region to PPs looking for an intervention.

## Screenshot

image

### DIFF
--- a/server/routes/findInterventions/interventionDetailsPresenter.test.ts
+++ b/server/routes/findInterventions/interventionDetailsPresenter.test.ts
@@ -91,6 +91,16 @@ describe(InterventionDetailsPresenter, () => {
       })
     })
 
+    describe('Region', () => {
+      it('is the NPS region name', () => {
+        expect(
+          linesForKey('Region', {
+            npsRegion: { id: 'A', name: 'North East' },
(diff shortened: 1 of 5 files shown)


server/routes/findInterventions/searchResultsPresenter.ts
```diff
@@ -7,7 +7,7 @@ export default class SearchResultsPresenter {
   constructor(
     private readonly interventions: Intervention[],
     private readonly filter: InterventionsFilter,
-    private readonly pccRegions: PCCRegion[]
```
tomsmyers, line 10: should these classes now have both region and location?

Gweaton: Rather than PCC Region and NPS region?

tomsmyers (approved):
nice

Gweaton:
@tomsmyers @lawrence-forooghian Have updated this to use individual `NPSRegion` and `PCCRegion` rather than shared `Region` to keep things separate and avoid confusion.

PR #133, opened 2021-02-18
IC-1141: Add contracts for assigning a referral to a caseworker and representing an assigned referral

lawrence-forooghian:

Adds contracts for assigning a referral to a caseworker and representing an assigned referral. This relates to the API issue IC-1142.

To allow us to build the user interface for assigning a referral to a caseworker.

### DIFF
--- a/server/services/interventionsService.test.ts
+++ b/server/services/interventionsService.test.ts
@@ -1069,6 +1069,41 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
     })
   })
 
+  describe('assignSentReferral', () => {
+    it('returns the sent referral, with the assignedTo property populated', async () => {
+      await provider.addInteraction({
+        state: 'There is an existing sent referral with ID of 400be4c6-1aa4-4f52-ae86-cbd5d23309bf',
+        uponReceiving:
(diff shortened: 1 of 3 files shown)


server/services/interventionsService.test.ts
```diff
@@ -1039,6 +1064,34 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
     })
   })
 
+  describe('assignSentReferral', () => {
```
lawrence-forooghian, line 1067: Are there any validation failures that we might need to handle?

mahalrmoj: We don't expect the request ti result in any validation failures

On another note, we would like the id for the assignedTo user so it can be created/update in the database as a result of the request.

Gweaton: Do we not want a test for if there wasn't no user found with that username/id?

lawrence-forooghian: > On another note, we would like the id for the assignedTo user so it can be created/updated in the database as a result of the request.

Added `userId` to the `AuthUser` type.

lawrence-forooghian: > Do we not want a test for if there's no user found with that username/id?

I don't think we need it - we don't have any other tests of a 404 response in this file.

tomsmyers (approved):

PR #134, opened 2021-02-18
IC-1170: Add Stylelint for SASS linting

Gweaton:

- Updates package-lock version for latest version of NPM (7.5.4 at time of PR)
- Adds Stylelint with a few custom rules to override the defaults (see below)
- Fixes existing linting violations (or adds rules to ignore them) 
- Adds automatic fixing of style errors to pre-commit hook

### Rules

- `"selector-class-pattern": null`

Allows us to use styles with double underscores.

- `"color-named": null`

Allows us to use named colours.

- `"Rule-empty-line-before": "always-multi-line"`

Ensures we have an empty line between each rule, except the first nested rule.

- `Max-nesting-depth: 2`

Set due to current css styles, I don't know what a sensible max nesting depth
is, but this at least allows our current implementation to pass.

- `selector-max-id: 1`

Allows us to apply styles to an ID - this is turned off in Stylelint as a
default, but our current styles from the Typescript template already apply
styles to IDs.

### Possibly controversial rule added:

Alphabetical ordering of styles - I do quite like this, as it keeps things neat and consistent, but can imagine others might prefer not to bother with this. However, as it's fixed automatically from now on, it shouldn't be a problem.

- To enforce consistent SASS style throughout codebase.
- To save time on code review by ensuring code style violations are all caught automatically.

### DIFF
--- a/doc/architecture/decisions/0010-use-stylelint-for-linting-styles.md
+++ b/doc/architecture/decisions/0010-use-stylelint-for-linting-styles.md
@@ -1,0 +1,31 @@
+# 10. Use Stylelint for linting styles
+
+Date: 2021-02-18
+
+## Status
+
+Accepted
+
+## Context
+
+We want to enforce consistency in our code, and catch as many errors
+automatically as we are able to. Linting the code is good practice to achieve
+these aims. [Stylelint](https://stylelint.io/) is one of the more popular CSS
+linters with support for SASS, and is easily configurable for our purposes.
+
+## Decision
+
+We will check SASS syntax using Stylelint.
+
+We will use the recommended configuration for plugins where possible.
+
+We will use Stylelint to automatically fix linting errors in a pre-commit hook.
+
+## Consequences
+
+Stylelint enables us to agree on and enforce a code style without having to
+keep it in the heads of developers, meaning we shouldn't have to discuss
+individual code style violations as they come up.
+
+It will save time in code review, as any style issues will have been caught
+before the PR is opened.
--- a/package-lock.json
+++ b/package-lock.json
@@ -86,6 +86,8 @@
         "nodemon": "^2.0.5",
         "prettier": "^2.1.2",
         "sass": "^1.27.0",
+        "stylelint": "^13.10.0",
+        "stylelint-config-sass-guidelines": "^8.0.0",
         "supertest": "^4.0.2",
         "ts-jest": "^26.4.1",
         "typescript": "^4.0.3"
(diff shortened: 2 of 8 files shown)


package-lock.json
```diff
@@ -1,8 +1,15952 @@
 {
```
lawrence-forooghian, line 1: I'm finding this diff quite difficult to parse. What was the command that you ran to generate this change? Might be worth popping in the commit message.

package.json
```diff
@@ -22,7 +22,7 @@
     "int-test-ui": "cypress open",
     "record-build-info": "node ./bin/record-build-info",
     "security_audit": "npx audit-ci --config audit-ci.json",
-    "lint": "eslint . --cache --max-warnings 0",
+    "lint": "eslint . --cache --max-warnings 0 && stylelint **/*.scss",
```
lawrence-forooghian, line 25: Will Stylelint know to exclude the files in `node_modules`?

doc/architecture/decisions/0010-use-stylelint-for-linting-styles.md
```diff
@@ -0,0 +1,31 @@
+# 10. Use Stylelint for linting styles
```
lawrence-forooghian, line 1: Awesome

.circleci/config.yml
```diff
@@ -26,7 +26,6 @@ jobs:
       tag: 14-stretch-browsers
     steps:
       - checkout
-      - node/install-npm
```
lawrence-forooghian, line 29: What motivated you to make this change?

lawrence-forooghian (commented):
Awesome, great idea to add this. A few comments.

package.json
Gweaton: Apparently, yes - according to [this](https://stylelint.io/user-guide/configure#ignorefiles):

> stylelint ignores the node_modules directory by default. However, this is overridden if ignoreFiles is set.

package-lock.json
Gweaton: Have added to the commit message, but it was running `npm install` with the latest stable version of npm - I'm not sure why it changed so much, but it seems to do so every time I run `npm install` (and did on one of Israt's branches too) so thought it'd be good to update here at least. I kept it in its own commit (separate from the one that adds the Stylelint package) to make it a bit clearer that it's separate.

.circleci/config.yml
Gweaton: The build was failing on CI due to issues installing Cypress. I thought this was due to having multiple versions of npm or Cypress installed but I'm not sure if this is actually the correct fix - was hoping someone might be able to give me a hand with this today.

Gweaton: Have removed this to spend some more time on it today.

tomsmyers (approved):

PR #135, opened 2021-02-19
Do not log the user out on authentication errors.

tomsmyers:

Do not log the user out on authentication errors.

I find this behaviour really annoying and confusing. When a page
returns a 403 or 401 response, the user is redirected to /logout,
and ends up back at the login page. It's much better to tell the user
why they can't view the page.

We can tidy up this behaviour later, but for now just removing the
redirect it miles better IMO.

In the short term - make it easier to see what's going on when you hit 403s.


lawrence-forooghian (approved):
💯

lawrence-forooghian (approved):

PR #142, opened 2021-02-23
IC-1237: Allow NPS Region to be null

Gweaton:

Allows NPS Region to be `null` on an Intervention, only adding it to the summary list if it's present.

We're not 100% sure whether an NPS region is required, as it's not currently filterable. Until we figure this out, it's safer to allow it to be null and not break the user journey.

## Screenshot

image


Gweaton:
@tomsmyers I'm assuming we'll need to update the API for this change.

tomsmyers (approved):
nice - thanks

PR #140, opened 2021-02-22
A service provider user can assign a referral to a caseworker

lawrence-forooghian:
## What does this pull request do?

This allows a service provider user:

- to enter the email address of another service provider user ("caseworker") on the referral details page
- to view the details of the caseworker that they have selected
- to confirm this choice and assign the referral to that caseworker

The assignee is then shown on:

- the service provider dashboard (their username)
- the referral details page (their full name)

## What is the intent behind these changes?

It's the first step in the delivery of an intervention.

PR #143, opened 2021-02-23
UNDO: Pass whole SentReferral to showReferralPresenter

Gweaton:

A combination of
- Revert "Don't pass whole sentReferral to referralPresenter"
  This reverts commit b81a92e3da70f79f602dc02a306188f7d6402abd.
- Rename `referral` variable to be `sentReferral` so we're not calling`referral.referral` all over the place.

As it turns out, we're going to be using the `AssignedTo` top level field on the `sentReferral` in https://github.com/ministryofjustice/hmpps-interventions-ui/pull/140, so it makes sense to undo what I did before....


tomsmyers (approved):
lgtm

PR #144, opened 2021-02-23
IC-1232: Add View column to SP dashboard

Gweaton:

Adds View column to SP dashboard and removes clickable referral number that takes user to the SP referral
details page.

To make the CTA clearer to users

## Screenshot

image


tomsmyers (approved):
nice

PR #146, opened 2021-02-24
IC-1246: Use correct AuthUser response object when assigning a referral

Gweaton:

Uses the correct `AuthUser` from HMPPS Auth's '/api/authuser' endpoint when assigning a referral to a caseworker.

If the user doesn't exist, we render an ugly error page for now - we're going to be handling the case better in IC-1180.

This object has a different structure to the usual User model we were attempting to use before, meaning the journey silently never returned a name for the user.

What also made this tricky was that this the 'api/authuser' endpoints we consume return an array of AuthUsers, rather than a single one. If this were to return multiple users, we'd probably want to handle this better.

### DIFF
--- a/server/data/hmppsAuthClient.ts
+++ b/server/data/hmppsAuthClient.ts
@@ -1,4 +1,4 @@
-import superagent from 'superagent'
+import superagent, { Response } from 'superagent'
 import querystring from 'querystring'
 import redis from 'redis'
 import { promisify } from 'util'
@@ -44,12 +44,24 @@ function getApiClientTokenFromHmppsAuth(username?: string): Promise<superagent.R
     .timeout(timeoutSpec)
 }
 
-export interface User {
+interface User {
   name: string
   userId: string
   username: string
   authSource: string
   active: boolean
+}
+
+export interface AuthUser {
+  userId: string
+  username: string
+  email: string
+  firstName: string
+  lastName: string
+  locked: boolean
+  enabled: boolean
+  verified: boolean
+  lastLoggedIn: string
 }
 
 interface UserRole {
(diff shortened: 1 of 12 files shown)


server/data/hmppsAuthClient.ts
```diff
@@ -71,14 +83,23 @@ export default class HmppsAuthClient {
...
+    return authUsers[0]
   }
 
-  getUserByUsername(token: string, username: string): Promise<User> {
+  async getUserByUsername(token: string, username: string): Promise<AuthUser> {
     logger.info(`Getting user detail by username: calling HMPPS Auth`)
-    return this.restClient(token).get({ path: `/api/authuser/${username}` }) as Promise<User>
+    const authUsers = (await this.restClient(token).get({
+      path: `/api/authuser/${username}`,
+    })) as AuthUser[]
```
tomsmyers, line 100: this endpoint returns a single value, not a list.

server/data/hmppsAuthClient.ts
```diff
@@ -71,14 +83,23 @@ export default class HmppsAuthClient {
...
-    logger.info(`Getting user detail by email address: calling HMPPS Auth`)
-    return this.restClient(token).get({ path: `/api/authuser`, query: { email: emailAddress } }) as Promise<User>
+  async getUserByEmailAddress(token: string, emailAddress: string): Promise<AuthUser> {
+    logger.info(`Getting auth user detail by email address: calling HMPPS Auth`)
+    const authUsers = (await this.restClient(token).get({
+      path: `/api/authuser`,
+      query: { email: emailAddress },
+    })) as AuthUser[]
+
+    return authUsers[0]
```
tomsmyers, line 93: what happens when the API returns an empty list ?

tomsmyers: when this occurs the response code is 204 success. so it won't throw an exception.

server/data/hmppsAuthClient.ts
Gweaton: Oops, well spotted. Have updated this

server/data/hmppsAuthClient.ts
Gweaton: We've fixed this now

tomsmyers (approved):

PR #152, opened 2021-02-25
IC-1180 validate user email address for assigning referrals to SP users

tomsmyers:

validate user email address for assigning referrals to SP users

ensure referrals cannot be assigned to non-existent or mistyped email addresses

### DIFF
--- a/server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
+++ b/server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
@@ -1,7 +1,8 @@
 import { Request, Response } from 'express'
+import querystring from 'querystring'
 import CommunityApiService from '../../services/communityApiService'
 import InterventionsService from '../../services/interventionsService'
-import HmppsAuthClient from '../../data/hmppsAuthClient'
+import HmppsAuthClient, { AuthUser } from '../../data/hmppsAuthClient'
 import CheckAssignmentPresenter from './checkAssignmentPresenter'
 import CheckAssignmentView from './checkAssignmentView'
 import DashboardPresenter from './dashboardPresenter'
@@ -10,6 +11,8 @@ import ShowReferralPresenter from './showReferralPresenter'
 import ShowReferralView from './showReferralView'
 import AssignmentConfirmationView from './assignmentConfirmationView'
 import AssignmentConfirmationPresenter from './assignmentConfirmationPresenter'
+import { FormValidationError } from '../../utils/formValidationError'
+import errorMessages from '../../utils/errorMessages'
 
 export default class ServiceProviderReferralsController {
   constructor(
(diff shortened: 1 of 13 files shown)


Gweaton (commented):
So far so good - I'll have another look tomorrow if you're still working on it tonight, as my brain is too fried to see if I can determine any alternatives for the `checkAssignment` implementation (it's probably fine!)

package.json
```diff
@@ -119,6 +119,7 @@
     "nunjucks": "^3.2.3",
     "passport": "^0.4.1",
     "passport-oauth2": "^1.5.0",
+    "qs": "^6.7.0",
```
Gweaton, line 122: Does `qs` give us something that doesn't come with the Node `Query string` package (already part of `package.lock`)?

I've just tried it locally, and using `QueryString.stringify` seems to produce the same result - maybe we don't need to worry about importing a new package for this

server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
```diff
@@ -47,32 +50,71 @@ export default class ServiceProviderReferralsController {
...
     const view = new CheckAssignmentView(presenter)
 
-    res.render(...view.renderArgs)
+    return res.render(...view.renderArgs)
   }
 
   async assignReferral(req: Request, res: Response): Promise<void> {
-    // TODO IC-1180 - Validation: presence
     const { email } = req.body
+    if (email === undefined || email === null || email === '') {
```
Gweaton, line 113: I know we have a lot of `if` statements like this in the codebase, but would `if(!email) {}` cover all this? Happy to leave it if we want to be more explicit, though as it's sometimes unclear what it's gonna be.

Gweaton (commented):
Nice work, just one question about the `qs` package (and one other non-blocking thing)

package.json
tomsmyers: good point - i wasn't aware node had this built in. differences are there but unimportant for us:

https://stackoverflow.com/questions/29136374/what-the-difference-between-qs-and-querystring

we already have this lib pulled in as a transitive dep. so which one we use is here nor there. based on the SO answer above, since qs provides more features, and we already have it anyway, i'm happy with using it here.

server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
tomsmyers: yeh checking for truthy is fine here. like you say there are lots of blocks like this though. i'd probably rather address them all at once in a refactor PR.

package.json
Gweaton: Just copying this from our slack conv: I'd rather we used Querystring as it's a dependency we don't have to worry too much about updating on its own, as it's part of Node, and we haven't needed the additional things that `qs` provides up til now.

But I'll leave it up to you - one thing, if you do decide to stick with `qs`, would you mind updating the commit message to add a short description as to why we're using it?

Gweaton (approved):
Thanks for making the changes 🏅

PR #159, opened 2021-03-01
Fix: `await` response of assigning referral

Gweaton:

Adds `await` keyword to async function that was left off in error.

This was causing assignments to fail locally, as we weren't waiting for
a response before redirecting.

I've tried to get the compiler to complain when this is an issue when
running the build, using Typescript-ESLint's `no-floating-promises`
setting, but haven't been able to get it working yet.


tomsmyers (approved):
awesome - are you planning on adding the eslint config in another PR ?

PR #160, opened 2021-03-01
Catch floating awaits

Gweaton:

Adds `no-floating-awaits` rule to ESLint rules, to ensure we don't miss any async behaviour we need to `await`.

Also fixes/ignores rule violations throughout the codebase.

This came about as we missed an `await` when assigning a referral to a caseworker and it was very difficult to debug, so should warn us if we're trying to do this in future.


tomsmyers (approved):
awesome, thanks.

PR #163, opened 2021-03-02
IC-1281 filter disabled and unverified auth users from 'getSPByEmailAddress'.

tomsmyers:

in HMPPS Auth there are 'inactive' users who have had their account disabled.
there are also users who have unverified email addresses, i.e. auth can't be sure
the email they have provided is actually theirs. in both cases, we should assume
that these user accounts are not able to use our service properly and the users
should filtered out.

do not allow referral assignments to inactive or unverified auth accounts.

### DIFF
--- a/server/data/hmppsAuthClient.test.ts
+++ b/server/data/hmppsAuthClient.test.ts
@@ -55,17 +55,28 @@ describe('hmppsAuthClient', () => {
 
   describe('getSPUserByEmailAddress', () => {
     describe('when a matching user is found with the requested email address', () => {
-      it('should return the first matching user from the API response', async () => {
+      it('should return the first active and verified user from the API response', async () => {
         const response = [
+          {
+            userId: 'D89C09A9-1FAC-476D-91F8-E0EDCA10ECAF',
+            username: 'disable user account',
+            email: 'user@example.com',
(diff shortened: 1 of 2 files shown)


server/data/hmppsAuthClient.test.ts
```diff
@@ -63,9 +63,9 @@ describe('hmppsAuthClient', () => {
             email: 'user@example.com',
             firstName: 'Auth',
             lastName: 'User',
-            locked: true,
-            enabled: false,
-            verified: false,
+            locked: false,
```
Gweaton, line 66: I think it'd be a good idea now we have this filtering behaviour to add a couple more users to this response (possibly a locked/unverified one first) so we can check we're selecting the right one. Then we could update the `it` to say something like "should return the first matching **active** user from the API response"

Gweaton (commented):
Just one thing. Also, I'm assuming all our dev users are all active/verified etc, but if not, would be good to catch before demoing at show & tell!

Gweaton (approved):

server/data/hmppsAuthClient.test.ts
tomsmyers: done

PR #164, opened 2021-03-03
IC-1216: Add skeleton action plan activities page

Gweaton:

Adds a blank action plan "add activities" page with just the titles. This is currently linked from the Show Referral page, but will actually come from an SP progress page (still to build, but probably as a separate PR).

To provide a skeleton so we can build the "add number of sessions page" in parallel.

### Screenshot

image

### DIFF
--- a/integration_tests/integration/serviceProviderReferrals.spec.js
+++ b/integration_tests/integration/serviceProviderReferrals.spec.js
@@ -3,6 +3,7 @@ import serviceCategoryFactory from '../../testutils/factories/serviceCategory'
 import deliusUserFactory from '../../testutils/factories/deliusUser'
 import deliusServiceUserFactory from '../../testutils/factories/deliusServiceUser'
 import hmppsAuthUserFactory from '../../testutils/factories/hmppsAuthUser'
+import draftActionPlanFactory from '../../testutils/factories/draftActionPlan'
 
 describe('Service provider referrals dashboard', () => {
   beforeEach(() => {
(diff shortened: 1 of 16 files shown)


server/routes/serviceProviderReferrals/dashboardPresenter.ts
```diff
@@ -33,7 +33,13 @@ export default class DashboardPresenter {
         },
         { text: utils.convertToProperCase(serviceCategory.name), sortValue: null, href: null },
         { text: referral.assignedTo?.username ?? '', sortValue: null, href: null },
-        { text: 'View', sortValue: null, href: `/service-provider/referrals/${referral.id}` },
+        {
+          text: 'View',
+          sortValue: null,
+          href: referral.assignedTo?.username
```
tomsmyers, line 39: this seems to be a weird place to initialite the add activities page to me

Gweaton:
Will squash https://github.com/ministryofjustice/hmpps-interventions-ui/pull/164/commits/5c9d042d3d690a653c7d98e014398698984fec94 into https://github.com/ministryofjustice/hmpps-interventions-ui/pull/164/commits/da7b71e0d2ea0ff5789540f4dde5c83f9c68d19d if we think it's a good approach to go down.

server/routes/serviceProviderReferrals/actionPlanFormPresenter.test.ts
```diff
@@ -1,8 +1,8 @@
-import AddActionPlanActivitiesPresenter from './addActionPlanActivitiesPresenter'
```
lawrence-forooghian: I'm a bit confused about the `addActionPlanActivities` -> `actionPlanForm` rename. Could you explain a little why you think it'll be useful to use the same presenter / view for multiple pages in the journey? We didn't do this for the referral journey because the pages didn't really have much in common. Curious to know why we think this journey is different.

In particular, does it make sense to have an `ActionPlanFormView` that renders a more specifically-named `serviceProviderReferrals/actionPlan/addActivities`?

Let's talk it through if it's easier that way.

lawrence-forooghian (commented):
All the functionality looks good to me, just a question about the final rename.

server/routes/serviceProviderReferrals/actionPlanFormPresenter.test.ts
Gweaton: This was something @tomsmyers suggested last week, and introduced as part of his PR. I think I quite liked the approach as it's a small form with only 3 or so pages so felt quite nice to keep it self-contained.

I can see your point about the `view` though, I don't think it's quite right - a view should represent a single screen/page in the service. I guess it feels a bit out of sync to have a single `FormPresenter` but individual `view` objects so am happy to undo that commit (hence why I didn't squash it in yet!)

@tomsmyers do you have a strong preference on this, or any other reasons I've missed?

lawrence-forooghian: Yeah, personally I don't understand the benefit - I think that the "one presenter per page" model is easy to understand, and that it makes sense to deviate from it if for example you've got a whole bunch of very similar pages, but I don't think that's the case here.

tomsmyers: no strong preference - just want to avoid repetition where possible.

i think the three pages in question are very similar though. they share everything but the input elements.

Gweaton: Ok, as discussed have removed the last commit on this.

lawrence-forooghian (approved):

PR #167, opened 2021-03-04
IC-1218 create page to enter number of sessions for action plan

tomsmyers:

create page to enter number of sessions for action plan

this is unfinished code - but since george and i are working on the same screens some of this code needs to be merged early.

there is some refactoring in here - sorry it's not split out into seperate commits. i was rushing.

add the 'create action plan' form

### DIFF
--- a/server/routes/index.ts
+++ b/server/routes/index.ts
@@ -53,6 +53,10 @@ export default function routes(router: Router, services: Services): Router {
     }
   })
 
+  get('/service-provider/referrals/:id/action-plan/form/sessions', (req, res) =>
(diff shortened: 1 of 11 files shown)


server/routes/index.ts
```diff
@@ -53,6 +53,10 @@ export default function routes(router: Router, services: Services): Router {
     }
   })
 
+  get('/service-provider/referrals/:id/action-plan/form/sessions', (req, res) =>
```
Gweaton, line 56: We've changed this to be action plan ID rather than referral id - did you forget to stash pop?

PR #169, opened 2021-03-09
IC-1282 Refactor logging to use structured logs throughout.

tomsmyers:

This change adds a logger factory method which can be used to create
logger instances with common attributes. This is used in the restClient
base class to include the client name in every log message.

improve our production logs


Gweaton (approved):
This makes sense to me

PR #170, opened 2021-03-09
IC-1357: Service provider intervention progress page

lawrence-forooghian:

Creates a basic page to display intervention progress to a service provider user, and then moves the create action plan journey from the referral details page to this new page. It also tells the user whether an action plan has already been submitted.

To allow a service provider user to progress an intervention and view the progress of an intervention.

### Action plan not yet submitted

Screenshot_2021-03-09 HMPPS Interventions - GOV UK(1)

### Action plan already submitted

Screenshot_2021-03-09 HMPPS Interventions - GOV UK

### DIFF
--- a/server/routes/serviceProviderReferrals/interventionProgressPresenter.ts
+++ b/server/routes/serviceProviderReferrals/interventionProgressPresenter.ts
@@ -1,0 +1,15 @@
+import { SentReferral, ServiceCategory } from '../../services/interventionsService'
+import utils from '../../utils/utils'
+
+export default class InterventionProgressPresenter {
+  constructor(private readonly referral: SentReferral, private readonly serviceCategory: ServiceCategory) {}
+
+  readonly createActionPlanFormAction = `/service-provider/referrals/${this.referral.id}/action-plan`
+
+  readonly text = {
+    title: utils.convertToTitleCase(this.serviceCategory.name),
+    actionPlanStatus: 'Not submitted',
+  }
+
+  readonly actionPlanStatusColour: 'active' | 'inactive' = 'inactive'
+}
--- a/server/routes/serviceProviderReferrals/interventionProgressPresenter.ts
+++ b/server/routes/serviceProviderReferrals/interventionProgressPresenter.ts
@@ -1,15 +1,23 @@
-import { SentReferral, ServiceCategory } from '../../services/interventionsService'
+import { ActionPlan, SentReferral, ServiceCategory } from '../../services/interventionsService'
 import utils from '../../utils/utils'
 
 export default class InterventionProgressPresenter {
-  constructor(private readonly referral: SentReferral, private readonly serviceCategory: ServiceCategory) {}
+  constructor(
+    private readonly referral: SentReferral,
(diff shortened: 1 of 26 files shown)


server/routes/serviceProviderReferrals/interventionProgressPresenter.ts
```diff
@@ -0,0 +1,23 @@
...
+  ) {}
+
+  readonly createActionPlanFormAction = `/service-provider/referrals/${this.referral.id}/action-plan`
+
+  readonly text = {
+    title: utils.convertToTitleCase(this.serviceCategory.name),
+    actionPlanStatus: this.actionPlanSubmitted ? 'Submitted' : 'Not submitted',
+  }
+
+  readonly actionPlanStatusColour: 'active' | 'inactive' = this.actionPlanSubmitted ? 'active' : 'inactive'
```
tomsmyers, line 18: these don't sound like colours :P

server/services/interventionsService.test.ts
```diff
@@ -1046,6 +1047,34 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
         )
       })
     })
+
+    describe('for a referral that has an action plan', () => {
```
tomsmyers, line 1051: @mahalrmoj heads up that these contracts now exist

server/views/serviceProviderReferrals/interventionProgress.njk
```diff
@@ -0,0 +1,50 @@
...
+
+      {{ govukSummaryList(initialAssessmentSummaryListArgs(govukTag)) }}
+
+      <h2 class="govuk-heading-m">Action plan</h2>
+
+      <p class="govuk-body">
+      Please complete the service user’s action plan and submit it to the probation practitioner for approval.
+      </p>
+
+      {{ govukSummaryList(actionPlanSummaryListArgs(govukTag, csrfToken)) }}
```
tomsmyers, line 31: personally i'm not a huge fan of passing these template vars back into the js but it's not a blocker for me

tomsmyers (approved):
nice work

server/routes/serviceProviderReferrals/interventionProgressPresenter.ts
lawrence-forooghian: Good point - gonna rename it to "style".

server/views/serviceProviderReferrals/interventionProgress.njk
lawrence-forooghian: Me neither, but especially in the case of the Nunjucks macro I think it would take a bigger refactor, because somehow we'd need to I guess import all of those `govuk` macros into our global Nunjucks environment and then pass this environment through to all of the views so that we can pull the macros out of it. Do you think it's worth investing time in that (or do you have an alternative in mind)?

lawrence-forooghian: (The CSRF token is an easier one to fix, we could just pass that one in from the controller)

PR #172, opened 2021-03-10
IC-1289 add contracts for appointment APIs

tomsmyers:

adds contracts for creating, updating, getting, and deleting action plan appointments (sessions)

implement the SP side of viewing appointments/sessions for an assigned referral

### DIFF
--- a/server/services/interventionsService.test.ts
+++ b/server/services/interventionsService.test.ts
@@ -1637,17 +1637,17 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
     const actionPlanAppointments = [
       {
         sessionNumber: 1,
-        appointmentTime: '2021-05-13T12:30:000000Z',
+        appointmentTime: '2021-05-13T13:30:00+01:00',
         durationInMinutes: 120,
       },
       {
         sessionNumber: 2,
-        appointmentTime: '2021-05-20T12:30:000000Z',
+        appointmentTime: '2021-05-20T13:30:00+01:00',
         durationInMinutes: 120,
       },
       {
         sessionNumber: 3,
-        appointmentTime: '2021-05-27T12:30:000000Z',
+        appointmentTime: '2021-05-27T13:30:00+01:00',
         durationInMinutes: 120,
       },
     ]
@@ -1681,7 +1681,7 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
   describe('createActionPlanAppointment', () => {
     const actionPlanAppointment = {
       sessionNumber: 1,
-      appointmentTime: '2021-05-13T12:30:000000Z',
+      appointmentTime: '2021-05-13T13:30:00+01:00',
       durationInMinutes: 120,
     }
 
@@ -1720,7 +1720,7 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
   describe('updateActionPlanAppointment', () => {
     const actionPlanAppointment = {
       sessionNumber: 2,
-      appointmentTime: '2021-05-13T12:30:000000Z',
+      appointmentTime: '2021-05-13T13:30:00+01:00',
       durationInMinutes: 60,
     }
 
--- a/server/services/interventionsService.test.ts
+++ b/server/services/interventionsService.test.ts
@@ -1632,6 +1632,51 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
       )
     })
   })
+
+  describe('getActionPlanAppointments', () => {
+    const actionPlanAppointments = [
+      {
+        sessionNumber: 1,
+        appointmentTime: '2021-05-13T12:30:000000Z',
+        durationInMinutes: 120,
+      },
+      {
+        sessionNumber: 2,
+        appointmentTime: '2021-05-20T12:30:000000Z',
+        durationInMinutes: 120,
+      },
+      {
+        sessionNumber: 3,
+        appointmentTime: '2021-05-27T12:30:000000Z',
+        durationInMinutes: 120,
+      },
+    ]
+
+    beforeEach(async () => {
+      await provider.addInteraction({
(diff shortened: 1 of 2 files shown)


server/services/interventionsService.ts
```diff
@@ -399,4 +405,41 @@ export default class InterventionsService {
...
+      path: `/action-plan/${actionPlanId}/appointment`,
+      headers: { Accept: 'application/json' },
+      data: { ...appointment },
+    })) as ActionPlanAppointment
+  }
+
+  async updateActionPlanAppointment(
+    token: string,
+    actionPlanId: string,
+    appointment: ActionPlanAppointment
```
mahalrmoj, line 433: ActionPlanAppointment contains a sessionNumber. In the case of an update is it applicable?

mahalrmoj: I do favour it as a parameter as it is part of the key to identifying an appointment and thus something that can't change, whereas the body contains elements that can change.

tomsmyers: done

mahalrmoj (approved):

server/services/interventionsService.ts
```diff
@@ -406,11 +406,24 @@ export default class InterventionsService {
...
+  async createActionPlanAppointment(
+    token: string,
+    actionPlanId: string,
+    appointment: ActionPlanAppointment
+  ): Promise<ActionPlanAppointment> {
+    const restClient = this.createRestClient(token)
+    return (await restClient.post({
+      path: `/action-plan/${actionPlanId}/appointment`,
+      headers: { Accept: 'application/json' },
+      data: { ...appointment },
```
lawrence-forooghian, line 431: Why not just 

```suggestion
      data: appointment,
```

server/services/interventionsService.ts
```diff
@@ -426,4 +431,18 @@ export default class InterventionsService {
...
+    token: string,
+    actionPlanId: string,
+    sessionNumber: number,
+    appointmentUpdate: Partial<ActionPlanAppointmentUpdate>
+  ): Promise<ActionPlanAppointment> {
+    const restClient = this.createRestClient(token)
+    return (await restClient.patch({
+      path: `/action-plan/${actionPlanId}/appointment/${sessionNumber}`,
+      headers: { Accept: 'application/json' },
+      data: { ...appointmentUpdate },
```
lawrence-forooghian, line 445: Same question as before about the use of the spread operator here.

server/views/serviceProviderReferrals/showReferral.njk
```diff
@@ -26,6 +26,7 @@
         <p class="govuk-body">This intervention is assigned to <strong>{{ presenter.text.assignedTo }}</strong>.</p>
 
         <form method="post" action="{{ presenter.createActionPlanFormAction }}">
+          <input type="hidden" name="_csrf" value="{{csrfToken}}">
```
lawrence-forooghian, line 29: What does this have to do with the rest of the PR? Can we ditch it? I think it's in George's other draft PR anyway.

server/services/interventionsService.test.ts
```diff
@@ -1552,17 +1552,17 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
     const actionPlanAppointments = [
       {
         sessionNumber: 1,
-        appointmentTime: '2021-05-13T12:30:000000Z',
```
lawrence-forooghian: These are all modifications to lines originally introduced by this PR, right? If so can we please incorporate them into the commits where the lines were introduced?

server/services/interventionsService.test.ts
```diff
@@ -1552,17 +1552,17 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
     const actionPlanAppointments = [
       {
         sessionNumber: 1,
-        appointmentTime: '2021-05-13T12:30:00.000000Z',
+        appointmentTime: '2021-05-13T13:30:00+01:00',
```
lawrence-forooghian, line 1555: I'm a little confused about this change. What are we trying to communicate with this time zone? Does the time zone passed here have some meaning, and if so what is the frontend meant to pass?

mahalrmoj: When implementing the contract tests server side a couple things came to light:

1) appointmentTime is in a period that is subject to BST (based on the local my server) and the serialisation library considers this when doing the serialisation, so whilst the time is materially the same, it serialises to a different string
2) if the milli's is set to zero, the component is omitted from the serialised form 

These changes were necessary to get the contract tests to function from the server side as it appears the verification is based on string values.

lawrence-forooghian: > appointmentTime is in a period that is subject to BST (based on the local my server) and the serialisation library considers this when doing the serialisation, so whilst the time is materially the same, it serialises to a different string

So, in that case it makes sense that the API might return ISO 8601 timestamps with a non-zero UTC offset in the GET appointments response. But in the contract tests I'm also seeing examples of non-zero offsets being supplied to the POST and PATCH endpoints too. And I'd like to understand what you need this offset for, and if that means there's an expectation on the client to calculate this offset?

I'd say that _if_ you do need time zone information, then a clearer way of communicating it is a tuple like (date, wall time, tz identifier), e.g. `('2021-03-01', '10:03:00.000', 'Europe/London')`.

server/services/interventionsService.ts
tomsmyers: you get some type warnings. it's an issue with the implementation of interfaces in typescript and this is the only way to get around it that was sensible

server/views/serviceProviderReferrals/showReferral.njk
tomsmyers: i cherry picked georges commit to fix this to make local testing palatable. does it matter which PR it's included in?

server/services/interventionsService.test.ts
tomsmyers: yup

server/views/serviceProviderReferrals/showReferral.njk
lawrence-forooghian: I think it's no longer necessary now that this is handled on the intervention progress page.

lawrence-forooghian (approved):

PR #173, opened 2021-03-12
IC-1028: Switch off local mocks

lawrence-forooghian:

- Turns off the mocks of the intervention service in local development environment
- Adds documentation about how to interact with the locally-running Interventions and HMPPS Auth services

To use the real interventions service, since at the moment all of the functionality that we require has been built in the real service. This allows us to test things that involve mutations, like adding activities to an action plan.

We might need to turn on some of the mocks again in the future, if there are endpoints that we need that haven't been built yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,12 +32,34 @@ npm run start:dev
 
 Navigate to `http://localhost:3000` and log in:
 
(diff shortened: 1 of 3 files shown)


tomsmyers (approved):
thank you!

PR #174, opened 2021-03-12
IC-1381 Remove Get /Draft-Action-Plan

mahalrmoj:

Removes GET /draft-action-plan/{id} from contract test

Make the URLs consistent and not distinguish by state for GET


tomsmyers (approved):
thanks

tomsmyers:
this needs some changes in the interventions service layer. i'll add them and request review from @lawrence-forooghian before merging.

server/services/interventionsService.ts
```diff
@@ -367,27 +354,27 @@ export default class InterventionsService {
...
     }
   }
 
-  async getDraftActionPlan(token: string, actionPlanId: string): Promise<DraftActionPlan> {
+  async getActionPlanForReferral(token: string, referralId: string): Promise<ActionPlan> {
     const restClient = this.createRestClient(token)
 
     return (await restClient.get({
-      path: `/draft-action-plan/${actionPlanId}`,
+      path: `/action-plan?referralId=${referralId}`,
```
lawrence-forooghian, line 375: Was this addition of `getActionPlanForReferral` here intentional? I don't see any added contract tests for it. Feels like a separate thing.

lawrence-forooghian (changes requested):

server/services/interventionsService.ts
tomsmyers: this is likely a merge mistake i'll remove

server/services/interventionsService.ts
```diff
@@ -126,14 +126,13 @@ export interface InterventionsFilterParams {
   maximumAge?: number
 }
 
-export interface DraftActionPlan extends ActionPlanFields {
+export interface ActionPlan {
```
lawrence-forooghian, line 129: Perhaps you don't consider it part of the scope of this PR, but I've realised that we don't have any contract tests that show a `GET /action-plan` where the fields are populated. Perhaps we could update the `it('returns an existing submitted action plan')` test with examples of these fields?

lawrence-forooghian (commented):
Another thing I just noticed. Also I reckon we can probably squash the two commits together.

server/services/interventionsService.ts
tomsmyers: can we do this in a follow up PR? this is blocking some other bits so i'm keen to get it merged.

lawrence-forooghian: Yeah, do you mind doing that follow up PR?

lawrence-forooghian: But could we squash the commits now though? Happy to approve then

tomsmyers: actually upon inspection there is a test for this


```it('returns an existing submitted action plan', async () => {
      const actionPlanId = '7a165933-d851-48c1-9ab0-ff5b8da12695'

        state: `an action plan exists with ID ${actionPlanId}, and it has been submitted`,
        uponReceiving: `a GET request to view the action plan with ID ${actionPlanId}`,
        withRequest: {
          method: 'GET',
          path: `/action-plan/${actionPlanId}`,
          headers: { Accept: 'application/json', Authorization: `Bearer ${token}` },
        },
        willRespondWith: {
          status: 200,
          body: Matchers.like({
            id: actionPlanId,
            submittedAt: '2021-03-09T15:08:38Z',
          }),
          headers: {
            'Content-Type': 'application/json',
          },
        },
      })```

tomsmyers: i have squashed as requested

tomsmyers: misread your comment - i will change the existing test to chekc for populated fields.

tomsmyers: @mahalrmoj i've updated the contracts in this PR. they should still pass on the backend. would you mind just validating that against this branch? thanks!

lawrence-forooghian (approved):

PR #177, opened 2021-03-17
IC-1388 add 'back' link to referral navigation template to return easily to the SP dashboard

tomsmyers:

IC-1388 add 'back' link to referral navigation template to return easily to the SP dashboard

make it easier to return without using browser back button

### DIFF
--- a/server/views/serviceProviderReferrals/referralNavigationTemplate.njk
+++ b/server/views/serviceProviderReferrals/referralNavigationTemplate.njk
@@ -3,6 +3,7 @@
 {% from "govuk/components/input/macro.njk" import govukInput %}
 {% from "govuk/components/button/macro.njk" import govukButton %}
 {% from "moj/components/sub-navigation/macro.njk" import mojSubNavigation %}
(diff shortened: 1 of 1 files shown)


Gweaton (dismissed):

server/views/serviceProviderReferrals/referralNavigationTemplate.njk
```diff
@@ -12,6 +13,12 @@
   {{ govukNotificationBanner(serviceUserBannerArgs) }}
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
+
+        {{ govukBackLink({
```
Gweaton, line 17: Sorry, approved then realised the indentation isn't quite right (I think) - shoud this not line up with the `<h1>` below? It's really hard to get nunjucks files linted properly... I wonder if there's an easy way to solve that.

tomsmyers: yeh i couldn't find a plugin that worked well. also i don't really even know the correct formatting for this half the time.

Gweaton (approved):

PR #178, opened 2021-03-17
IC-1233: Escape HTML to fix HTML injection vulnerabilities

Gweaton:

Uses `ViewUtils.escape` to escape anywhere we're rendering pure HTML in GOV UK components. Although some of these are probably over the top (e.g. escaping `href`s), it feels safest to escape everything.

I haven't worried about escaping the `staticContentController` views or methods, as these won't be part of the production UI.

To ensure we don't unwillingly render any HTML in our views where we pass pure HTML to a GOV UK component.

### DIFF
--- a/server/routes/probationPractitionerReferrals/dashboardView.ts
+++ b/server/routes/probationPractitionerReferrals/dashboardView.ts
@@ -1,3 +1,4 @@
+import ViewUtils from '../../utils/viewUtils'
 import DashboardPresenter from './dashboardPresenter'
 
 export default class DashboardView {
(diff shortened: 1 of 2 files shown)


tomsmyers:
nice - are these vulns listed anywhere? like is there a warning that this changeset fixes?

Gweaton:
@tomsmyers I don't see any browser warnings in my local environment, this is just something Lawrence picked up on that we hadn't spotted. I wonder if there's a safe way of noticing things like this in future, any ideas?

lawrence-forooghian (approved):

PR #183, opened 2021-03-18
IC-1400: Add contract for updating Sentence ID

Gweaton:

Adds a contract test for updating the referral on the Interventions Service with a Sentence ID.

Users will select the relevant sentence in the Refer journey based on values fetched from the Community API for that Service User's CRN.

The sentence against which the referral is being made must be associated with the referral by way of a sentence identifier from nDelius. This must be recorded for every referral.

### DIFF
--- a/server/services/interventionsService.test.ts
+++ b/server/services/interventionsService.test.ts
@@ -434,6 +434,41 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
       expect(referral.furtherInformation).toBe('Some information about the service user')
     })
 
+    it('returns the updated referral when selecting the relevant sentence', async () => {
+      await provider.addInteraction({
+        state: 'There is an existing draft referral with ID of d496e4a7-7cc1-44ea-ba67-c295084f1962',
+        uponReceiving: 'a PATCH request to update the sentence ID',
+        withRequest: {
+          method: 'PATCH',
+          path: '/draft-referral/d496e4a7-7cc1-44ea-ba67-c295084f1962',
+          headers: {
+            Accept: 'application/json',
+            'Content-Type': 'application/json',
+            Authorization: `Bearer ${token}`,
+          },
+          body: {
+            relevantSentenceId: 2600295124,
+          },
+        },
+        willRespondWith: {
+          status: 200,
+          body: {
+            id: 'd496e4a7-7cc1-44ea-ba67-c295084f1962',
+            relevantSentenceId: 2600295124,
+          },
+          headers: {
+            'Content-Type': 'application/json',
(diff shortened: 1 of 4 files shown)


mahalrmoj (changes requested):
SentenceId in community API is a BigInt/Long and speaking to David we should reflect that our side too

Gweaton:
Have updated this to use a `number` type. When we deserialize the value we get from the Community API to a `DeliusConviction` type, we'll raise an error if it overflows `Number`.

server/services/interventionsService.test.ts
```diff
@@ -434,6 +434,41 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
...
+          },
+          body: {
+            sentenceId: 2600295124,
+          },
+        },
+        willRespondWith: {
+          status: 200,
+          body: {
+            id: 'd496e4a7-7cc1-44ea-ba67-c295084f1962',
+            sentenceId: 2600295124,
```
sldblog, line 457: Could we call this `relevantSentenceId` or `contextSentenceId`? To encode that we select the sentence to set the referral in context?

server/services/interventionsService.test.ts
```diff
@@ -434,6 +434,41 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
       expect(referral.furtherInformation).toBe('Some information about the service user')
     })
 
+    it('returns the updated referral when selecting the relevant sentence', async () => {
+      await provider.addInteraction({
+        state: 'There is an existing draft referral with ID of d496e4a7-7cc1-44ea-ba67-c295084f1962',
```
sldblog, line 439: Is this the same state as `There is an existing draft referral with ID of d496e4a7-7cc1-44ea-ba67-c295084f1962, and it has had a service category selected`?

Would it be fair to say this is "a newly started empty referral with ID of ..."?

server/services/interventionsService.test.ts
Gweaton: How about `selectedSentenceId`?

server/services/interventionsService.test.ts
Gweaton: Does the newly-started state matter here? I felt that the important thing was that we simply have an existing draft referral, no matter what stage of the refer journey we're in

server/services/interventionsService.test.ts
Gweaton: `relevantSentenceId` sounds good to me

lawrence-forooghian (approved):

mahalrmoj (approved):

PR #187, opened 2021-03-23
IC-1401 & IC-1402: Allow users to select relevant sentence on a referral

Gweaton:

Adds screen for selecting relevant sentence on a referral during the Refer journey.

This currently doesn't display the "order" specified in the designs, as we're not quite sure where this comes from, but it displays the active convictions for a CRN returned by the Community API.

If only one active conviction comes back from the Community API, we pre-select that field for the user.

### NOTE:

- The data in our local Community API environment is lacking for our test users, and Convictions are returned without `sentence`s, or `offence`s. In real life, we shouldn't be referring any service users who don't have a sentence or an offence, so I've chosen to raise an error if this is the case, so this won't work locally until we add some better test data to the local instance of the Community API.

To link a referral with a sentence in Delius.

## Screenshot

image

### DIFF
--- a/server/routes/referrals/relevantSentencePresenter.ts
+++ b/server/routes/referrals/relevantSentencePresenter.ts
@@ -1,0 +1,42 @@
+import { DeliusConviction } from '../../services/communityApiService'
+import { ServiceCategory } from '../../services/interventionsService'
+import PresenterUtils from '../../utils/presenterUtils'
+
+export default class RelevantSentencePresenter {
+  constructor(private readonly serviceCategory: ServiceCategory, private readonly convictions: DeliusConviction[]) {}
+
+  readonly title = `Select the relevant sentence for the ${this.serviceCategory.name.toLocaleLowerCase()} referral`
+
+  get relevantSentenceFields(): {
+    category: string
+    subcategory: string
+    endOfSentenceDate: string
+    value: number
+    checked: boolean
+  }[] {
+    return this.convictions.map(conviction => {
+      if (!conviction.offences) {
+        throw new Error(`No offences found for conviction id: ${conviction.convictionId}`)
+      }
+
+      if (!conviction.sentence) {
+        throw new Error(`No sentences found for conviction id: ${conviction.convictionId}`)
+      }
+
+      const mainOffence = conviction.offences.find(offence => offence.mainOffence)
+
+      if (!mainOffence) {
+        throw new Error('No main offence found')
+      }
+
+      return {
+        category: mainOffence.detail.mainCategoryDescription,
+        subcategory: mainOffence.detail.subCategoryDescription,
+        endOfSentenceDate:
+          PresenterUtils.govukFormattedDateFromStringOrNull(conviction.sentence.expectedSentenceEndDate) ?? 'Not found',
+        value: conviction.convictionId,
+        checked: false,
+      }
+    })
+  }
+}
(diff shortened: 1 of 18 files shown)


server/routes/referrals/relevantSentenceView.ts
```diff
@@ -0,0 +1,40 @@
...
+          isPageHeading: true,
+          classes: 'govuk-fieldset__legend--xl',
+        },
+      },
+      items: this.presenter.relevantSentenceFields.map(relevantSentence => {
+        return {
+          value: relevantSentence.value,
+          html: `${ViewUtils.escape(relevantSentence.category)}<br>Subcategory: ${ViewUtils.escape(
+            relevantSentence.subcategory
+          )}<br>End of sentence date: ${relevantSentence.endOfSentenceDate}`,
```
lawrence-forooghian, line 24: Missing an `escape` on the end of sentence date.

server/routes/referrals/relevantSentencePresenter.ts
```diff
@@ -1,12 +1,21 @@
...
-  constructor(private readonly serviceCategory: ServiceCategory, private readonly convictions: DeliusConviction[]) {}
+  constructor(
+    private readonly serviceCategory: ServiceCategory,
+    private readonly convictions: DeliusConviction[],
+    private readonly error: FormValidationError | null = null
+  ) {}
 
   readonly title = `Select the relevant sentence for the ${this.serviceCategory.name.toLocaleLowerCase()} referral`
 
+  readonly errorMessage = PresenterUtils.errorMessage(this.error, 'relevant-sentence-id')
```
lawrence-forooghian, line 17: Can we test this property please?

lawrence-forooghian (commented):
This looks great, just two minor comments.

And a couple of other bits:

## Preventing submission

I think we need to prevent submission of the referral form until the desired sentence has been selected (`ReferralFormPresenter#canSubmitReferral`).

## Automatic selection

That then leads me to another thought.

I'm a bit wary about the UX of the automatically selecting a radio button. It means that the user is unable to differentiate between the following two states of this page:

- sentence not yet selected, user needs to click "Save and continue" to save the sentence selection
- sentence already selected, no action needed from user on this screen

This means that it's hence impossible for a user to understand _why_ they're being prevented from submitting the referral form - they need to click the "Save and continue" button but there's nothing on the screen that indicates this.

I think it might be best to never automatically select, what do you think?

Edit: Also, to add a bit of an appeal to authority, the [Design System](https://design-system.service.gov.uk/components/radios/) says:

> Do not pre-select radio options as this makes it more likely that users will:
> - not realise they’ve missed a question
> - submit the wrong answer

Gweaton:
> This looks great, just two minor comments.
> 
> And a couple of other bits:
> ## Preventing submission
> 
> I think we need to prevent submission of the referral form until the desired sentence has been selected (`ReferralFormPresenter#canSubmitReferral`).

Well-remembered, have added this as a new separate commit.

> ## Automatic selection
> 
> That then leads me to another thought.
> 
> I'm a bit wary about the UX of the automatically selecting a radio button. It means that the user is unable to differentiate between the following two states of this page:
> 
>     * sentence not yet selected, user needs to click "Save and continue" to save the sentence selection
> 
>     * sentence already selected, no action needed from user on this screen
> 
> 
> This means that it's hence impossible for a user to understand _why_ they're being prevented from submitting the referral form - they need to click the "Save and continue" button but there's nothing on the screen that indicates this.
> 
> I think it might be best to never automatically select, what do you think?
> 
> Edit: Also, to add a bit of an appeal to authority, the [Design System](https://design-system.service.gov.uk/components/radios/) says:
> 
> > Do not pre-select radio options as this makes it more likely that users will:
> > 
> > * not realise they’ve missed a question
> > * submit the wrong answer

Also a good point - it did feel a bit weird, but then I also thought, is it weird to ask someone to select a single thing from a list of a single item. I think you're right here, and you have the design system backing you up too. I've fixed up my commits with that change.

lawrence-forooghian (approved):

PR #189, opened 2021-03-25
IC-1433: Add contract test for recording attendance

Gweaton:

Adds a contract test and Interventions Service client function to set a SU's attendance on an appointment.

Posting to `/action-plan/{action-plan-id}/appointment/{session-number}/record-attendance` with:

```
{ 
  attended: 'late',
  additionalAttendanceInformation: 'Alex missed the bus',
}
```

will return the whole updated `appointment` object.

To allow Service Providers to notify Probation Practitioners if a Service User fails to attend a session.

Next up, we'll build the screen to actually allow users to enter this feedback.

### DIFF
--- a/server/services/interventionsService.ts
+++ b/server/services/interventionsService.ts
@@ -157,11 +157,17 @@ export interface ActionPlanAppointment {
   sessionNumber: number
   appointmentTime: string | null
   durationInMinutes: number | null
+  attendance?: AppointmentAttendance
 }
 
 export interface ActionPlanAppointmentUpdate {
   appointmentTime: string
   durationInMinutes: number
+}
+
+interface AppointmentAttendance {
+  attended: 'yes' | 'no' | 'late'
+  additionalAttendanceInformation?: string
 }
 
 export default class InterventionsService {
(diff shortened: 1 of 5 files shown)


server/services/interventionsService.ts
```diff
@@ -157,13 +157,19 @@ export interface ActionPlanAppointment {
...
+  attendance?: AppointmentAttendance
 }
 
 export interface ActionPlanAppointmentUpdate {
   appointmentTime: string
   durationInMinutes: number
 }
 
+interface AppointmentAttendance {
+  attended: string
```
lawrence-forooghian, line 169: Could we restrict this to the allowed values (either with an `enum` or with `"value1" | "value2" | "value3"`)?

server/services/interventionsService.ts
```diff
@@ -447,4 +453,19 @@ export default class InterventionsService {
       data: { ...appointmentUpdate },
     })) as ActionPlanAppointment
   }
+
+  async recordAppointmentAttendance(
```
lawrence-forooghian, line 457: I would love it if we could add the interventions service mocks at the same time as we add new endpoints; it saves time and context switching when we subsequently need the mocks during feature development. Would you be able to do that please?

Gweaton: Have added the mocks (and also the ones we use in integration tests)

server/services/interventionsService.ts
Gweaton: Have opted for string union type, rather than `enum`, after a bit of googling. It doesn't seem to make a lot of difference either way, but string union means we can write less code as we don't have to import an enum throughout the codebase (among other things).

PR #190, opened 2021-03-28
IC-1438: Auto generate GOV.UK Frontend macro args TypeScript types

lawrence-forooghian:

Automatically generates the contents of the `server/util/govukFrontendTypes.ts` file using metadata contained in the GOV.UK Frontend package.

To save us from having to populate this file manually as we continue to use new components from the GOV.UK Frontend.


server/routes/serviceProviderReferrals/interventionProgressView.ts
```diff
@@ -34,7 +34,7 @@ export default class InterventionProgressView {
   }
 
   private actionPlanSummaryListArgs(tagMacro: (args: TagArgs) => string, csrfToken: string): SummaryListArgs {
-    const rows: SummaryListRow[] = [
+    const rows: SummaryListArgsRow[] = [
```
Gweaton, line 37: Was this supposed to be part of this commit? Is this a new type that was generated that we wrote wrong, originally? Would you mind just adding a quick line about this change to the commit message if so?

Gweaton (approved):
This is amazing work, well done! Just one thing re: commit message, but approving so as not to block you.

Would you mind adding something to the README to explain when this script should be run?

server/routes/serviceProviderReferrals/interventionProgressView.ts
lawrence-forooghian: Yeah, it's just that the name of the type that we were using before isn't consistent with the naming used in the generated types, so I had to change it to fix the compilation error. I'll mention in commit message.

PR #191, opened 2021-03-29
IC-1407 Refactor authentication code to improve usability of the application

tomsmyers:

    Refactor authentication to improve usability of the application:

     - take all the 'passport' and login/logout related configuration and routes and put it in one place
     - only call 'getUserDetails' when we retrieve the token instead of on every request.
     - refactor the auth middlewares so the each one is responsible for one thing. there are now three
       auth related middlewares. they are: loginRedirectMiddleware - this is responsible for checking if the
       user is logged in already, and if not redirecting them to the login page; tokenVerificationMiddleware -
       this is responsible for checking the validity of the token to ensure the logged in user always has a valid
       and non-expired token; authorisationMiddleware - this applies RBAC for the routes it is applied to, allowing
       us to configure specific permissions to be required for specific routes.
     - provide a page which can be shown when the user is successfully logged out, instead of just automatically
       redirecting back to a login screen.
     - store the token expiry and roles in res.locals. this change requires refacting every usage of 'res.locals.token'
       so i have pulled that change into a subsequent commit to make the review easier.
     - udpate the 'auth error' page to have more meaningful content
     -  rename UserService.getUser to 'getUserDetails' and update the display name so we don't have to use a random nunjucks filter


mockApis/auth.ts
```diff
@@ -214,11 +215,11 @@ export default class AuthServiceMocks {
   }
 
   stubServiceProviderToken = async (): Promise<unknown> => {
-    return this.stubToken('delius')
```
tomsmyers, line 217: you have no idea how long it took me to find this bug. 

i don't know how these tests were passing before - but i find it reassuring that this change set surfaced this.

mockApis/auth.ts
```diff
@@ -122,7 +123,7 @@ export default class AuthServiceMocks {
           'Content-Type': 'application/json;charset=UTF-8',
         },
         jsonBody: {
-          staffId: 231232,
+          userId: '9C2744E3-65CD-4B40-B036-95DBD6F9A871',
```
tomsmyers, line 126: this isn't actually used - but it's more realistic of the actual response

lawrence-forooghian (approved):

PR #192, opened 2021-03-29
IC-1291: Filter out convictions with no sentence

Gweaton:

Filters out convictions with no sentence.

Sentence doesn't appear to be a required field in nDelius, so it may not
be set. If this is the case, we don't want to throw an error in the user
journey as there's nothing the user can do to solve this - we simply
shouldn't display the conviction.

## TODO:

The community api service doesn't appear to be tested at all - it's a tricky one to test because we instantiate it with a Rest Client which we'll need to mock a lot of the responses for. I've added a Jira ticket to the tech debt epic (https://dsdmoj.atlassian.net/browse/IC-1471) to make sure we address this, but I don't want to dedicate much more time to this right now.


tomsmyers (approved):
could do with a test, but change is fine 👍

PR #199, opened 2021-03-30
Add time input component

lawrence-forooghian:

- Introduces a way of adding custom view components, similar to the components offered by the GOV.UK Frontend library
- Extends the existing script, to generate TypeScript types for the arguments of these components' Nunjucks macros
- Adds a time input component, which is a fork of the GOV.UK Frontend's date input component which provides a space for an AM / PM select.

To allow us to accept a time value from the user on the upcoming "edit action plan appointment" page.


Gweaton (approved):
Nice work, looks good.

PR #201, opened 2021-03-30
Add a class for representing a duration, and then the ability to present it as form input fields

lawrence-forooghian:

Adds a `Duration` class which represents a quantity of time. It provides convenience methods for getting stopwatch values.

It then adds a presenter utility method that will allow us to display a form input backed by a `Duration` object.

To allow us to build the "duration" input on the upcoming "edit action plan appointment details" form.

### DIFF
--- a/server/utils/duration.test.ts
+++ b/server/utils/duration.test.ts
@@ -1,0 +1,20 @@
+import Duration from './duration'
+
+describe(Duration, () => {
+  describe('.fromUnits', () => {
+    it('converts hours, minutes and seconds to seconds', () => {
+      expect(Duration.fromUnits(10, 5, 32).seconds).toEqual(36332)
+      expect(Duration.fromUnits(0, 500, 0).seconds).toEqual(30000)
+    })
+  })
+
+  describe('stopwatch values', () => {
+    it('returns the correct values', () => {
+      const duration = new Duration(10 * 3600 + 5 * 60 + 32.5)
+
+      expect(duration.stopwatchHours).toEqual(10)
+      expect(duration.stopwatchMinutes).toEqual(5)
+      expect(duration.stopwatchSeconds).toEqual(32.5)
+    })
+  })
+})
--- a/server/utils/presenterUtils.test.ts
+++ b/server/utils/presenterUtils.test.ts
@@ -1,6 +1,7 @@
 import PresenterUtils from './presenterUtils'
 import draftReferralFactory from '../../testutils/factories/draftReferral'
 import CalendarDay from './calendarDay'
+import Duration from './duration'
 
 describe(PresenterUtils, () => {
   describe('stringValue', () => {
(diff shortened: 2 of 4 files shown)


server/utils/presenterUtils.ts
```diff
@@ -26,6 +39,41 @@ export default class PresenterUtils {
...
+  durationValue(
+    modelValue: Duration | null,
+    userInputKey: string,
+    error: FormValidationError | null
+  ): DurationInputPresenter {
+    const [hoursKey, minutesKey] = ['hours', 'minutes'].map(suffix => `${userInputKey}-${suffix}`)
+
+    const errorMessage = PresenterUtils.errorMessage(error, hoursKey) ?? PresenterUtils.errorMessage(error, minutesKey)
+
+    let hoursValue = ''
```
Gweaton, line 51: Don't know if this matters or not, but I think we can just declare this rather than assigning it to an empty string as it's always assigned in the `else` below, i.e.`let hoursValue: string`

Gweaton (commented):
This looks good, just had a couple of thoughts.

server/utils/presenterUtils.test.ts
```diff
@@ -237,6 +238,82 @@ describe(PresenterUtils, () => {
...
+            const value = utils.durationValue(Duration.minutes(75), 'duration', null)
+
+            expect(value).toMatchObject({ hours: { value: '1' }, minutes: { value: '15' } })
+          })
+        })
+      })
+
+      describe('when there is user input data', () => {
+        it('returns the user input data', () => {
+          const utils = new PresenterUtils({ 'duration-hours': 'egg', 'duration-minutes': 7 })
```
Gweaton, line 267: When I first saw this, I was a little confused about why 'egg' was here, as I assumed it would be doing something with validation errors. Is it just to show that _any_ value can go here?

lawrence-forooghian: Yep, it's to show that whatever the user input data is, it'll get played back without any attempt to convert it into a number. Does that make sense?

server/utils/presenterUtils.ts
lawrence-forooghian: It's not assigned if the nested `if (modelValue !== null)` is false, though, so the compiler would complain.

server/utils/duration.test.ts
```diff
@@ -0,0 +1,19 @@
+import Duration from './duration'
+
+describe(Duration, () => {
+  describe('.minutes', () => {
+    it('converts minutes to seconds', () => {
+      expect(Duration.minutes(70).seconds).toEqual(70 * 60)
```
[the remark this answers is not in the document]
lawrence-forooghian: Good shout - I've replaced that `.minutes` method with a more general `.fromUnits` one, too.

server/utils/presenterUtils.ts
Gweaton: ah you're right

server/utils/presenterUtils.test.ts
Gweaton: 👍

Gweaton (approved):

PR #203, opened 2021-03-30
Add a class for representing a clock time, and then the ability to present it as form input fields

lawrence-forooghian:

Adds a `ClockTime` class which represents the values displayed on a clock, analogous to our existing `CalendarDay` class. It provides convenience methods for getting 12-hour values.

It then adds a presenter utility method that will allow us to display a form input backed by a `ClockTime` object.

To allow us to build the "time" input on the upcoming "edit action plan appointment details" form.

### DIFF
--- a/server/utils/clockTime.test.ts
+++ b/server/utils/clockTime.test.ts
@@ -1,0 +1,78 @@
+import ClockTime from './clockTime'
+
+describe(ClockTime, () => {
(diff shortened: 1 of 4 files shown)


Gweaton (approved):
This is great, I think I should also be able to use some of it for displaying 12hr format dates in my branch.

PR #204, opened 2021-03-30
Add methods for extracting date & time from formatted date time strings

Gweaton:

- Adds methods for extracting date & time from formatted date time strings
- Refactors existing `DateUtils` file to use ES6 syntax

In the Post Session Feedback journey, we display dates and times to users, but we currently only have a single dateTime field. Ideally this should be in 12hr format, but that will be addressed in https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=539&view=planning&selectedIssue=IC-1452&issueLimit=100


lawrence-forooghian (approved):

server/utils/dateUtils.ts
```diff
@@ -1,32 +1,50 @@
...
-  const day = `0${ukDate.day}`.slice(-2)
-  const hour = `0${ukDate.hours}`.slice(-2)
-  const minutes = `0${ukDate.minutes}`.slice(-2)
+    if (formattedDateTime === '') {
+      return formattedDateTime
+    }
 
-  return `${day} ${month} ${year}, ${hour}:${minutes}`
-}
+    return formattedDateTime.split(',')[0].trim()
```
lawrence-forooghian, line 38: Personally I'd probably have made these two separate methods that return the components, and then have a method which combines them, instead of returning the combined thing and then splitting it back out again, but I don't think it matters enough to change.

Gweaton: This is a very good point, I'm changing this as it'll also make it easier to update in future and change to 12hr format etc.

lawrence-forooghian (approved):

Floppy (commented):
Is there a driver towards implementing this stuff ourselves rather than using a library? `Moment` was the common one for a long time, but is now legacy. Helpfully though it provides a good [list of alternatives](https://momentjs.com/docs/#/-project-status/recommendations/) and [Luxon](https://moment.github.io/luxon/) looks at first pass like it will do what you've implemented here. Obviously this code is written, so there's no point changing it, but I'm interested in whether there's a general approach to reducing third party dependencies that means this we've implemented this ourselves?

lawrence-forooghian:
Hey @Floppy, sorry for taking such a long time to reply on this one. I think that for this one, which is just formatting, you're right, it would make sense to revisit this and either check whether there's a simpler way to do it with the built-in `Date` functionality, or using a library. If it's just a matter of slightly nicer ergonomics, then I'd probably favour sticking to the standard library. But if it's a big difference in effort, then an external library makes sense. We already have the [date-fns](https://date-fns.org/) library in the project and I imagine it does what we need here.

I've created a tech debt ticket IC-2076 for this.

PR #207, opened 2021-03-31
Add missing action plan appointment contract for /appointments/:sessionNumber

Gweaton:

Adds missing contract for `getActionPlanAppointment`.

It looks like we were missing a contract for fetching an individual
appointment. I'm using this endpoint to fetch appointments for the Post
Session Feedback page, so want to make sure it's covered.


lawrence-forooghian (approved):
Good catch 👍

PR #209, opened 2021-04-01
Add method to create ClockTime from 12-hour components

lawrence-forooghian:

Adds a method to create a `ClockTime` instance from 12-hour clock components.

To allow us to convert the user input on the upcoming "edit action plan appointment" page into a `ClockTime`.

### DIFF
--- a/server/utils/clockTime.ts
+++ b/server/utils/clockTime.ts
@@ -8,7 +8,7 @@ export default class ClockTime {
    */
   private constructor(readonly hour: number, readonly minute: number, readonly second: number) {}
 
-  static fromComponents(hour: number, minute: number, second: number): ClockTime | null {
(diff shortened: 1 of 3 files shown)


server/utils/clockTime.ts
```diff
@@ -10,6 +12,21 @@ export default class ClockTime {
     return ClockTime.isValid(hour, minute, second) ? new ClockTime(hour, minute, second) : null
   }
 
+  static fromTwelveHourComponents(
```
Gweaton, line 15: Now that we have this, is it worth renaming the other `fromComponents` method to specify what type of arguments it takes for clarity?

lawrence-forooghian: Sure, added a commit for it.

Gweaton (approved):
Thanks!

PR #211, opened 2021-04-01
Add method to get minutes from a Duration

lawrence-forooghian:

Adds a method to get the number of minutes that a `Duration` instance represents.

To allow us to convert the user input on the upcoming "edit action plan appointment details" into an number of minutes to send as the `durationInMinutes` value to the interventions service.

### DIFF
--- a/server/utils/duration.ts
+++ b/server/utils/duration.ts
@@ -5,6 +5,16 @@ export default class Duration {
     return new Duration(hours * 3600 + minutes * 60 + seconds)
   }
 
+  // If the duration represents a whole number of minutes, then this
+  // returns the number of minutes. Else, it returns null.
+  get minutes(): number | null {
+    if (this.seconds % 60 === 0) {
+      return this.seconds / 60
+    }
+
+    return null
+  }
+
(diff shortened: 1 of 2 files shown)


Gweaton (approved):

server/utils/duration.ts
```diff
@@ -5,6 +5,16 @@ export default class Duration {
     return new Duration(hours * 3600 + minutes * 60 + seconds)
   }
 
+  // If the duration represents a whole number of minutes, then this
+  // returns the number of minutes. Else, it returns null.
+  get minutes(): number | null {
+    if (this.seconds % 60 === 0) {
```
Floppy, line 11: Is this check needed? Do you explicitly only want to get a number of minutes when there are the right number of seconds? Or could you divide and `floor()` it? I'm worried about things like leap seconds, etc, here.

lawrence-forooghian: In practice, in our app, the `Duration` that this property is being called on will always be an integer number of minutes (since it's actually derived from a validated integer number of minutes entered by the user).

PR #214, opened 2021-04-01
Make appointment update params nullable

lawrence-forooghian:

Allows us to set an action plan appointment's duration and time to `null`.

To allow us to build the upcoming "edit action plan appointment" form, which describes the time and duration as optional.

### DIFF
--- a/server/services/interventionsService.ts
+++ b/server/services/interventionsService.ts
@@ -161,8 +161,8 @@ export interface ActionPlanAppointment {
 }
 
 export interface ActionPlanAppointmentUpdate {
-  appointmentTime: string
-  durationInMinutes: number
+  appointmentTime: string | null
+  durationInMinutes: number | null
 }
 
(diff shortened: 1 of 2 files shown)


server/services/interventionsService.ts
```diff
@@ -161,8 +161,8 @@ export interface ActionPlanAppointment {
 }
 
 export interface ActionPlanAppointmentUpdate {
-  appointmentTime: string
-  durationInMinutes: number
+  appointmentTime: string | null
```
Gweaton, line 164: Hmm, didn't realise this was optional. Will need to update my PR after this is merged.

Gweaton (approved):

server/services/interventionsService.ts
lawrence-forooghian: I think it's always been optional in the model? All I've changed here is the update DTO.

PR #218, opened 2021-04-06
IC-1298: Make intervention progress page components reusable across journeys

Gweaton:

Refactors SP intervention journey Service User banner, referral navigation templates and Referral Overview Page Presenter so we can re-use them in the PP journey.


tomsmyers (approved):
nice work - thanks!

PR #219, opened 2021-04-06
IC-1298: Add PP session progress page

Gweaton:

Adds PP session progress page with various appointment states.

This depends on a few of the refactoring changes in #218 that make various components in the SP journey re-usable.

To give the PP an overview of how the intervention is progressing.

## Screenshot

image

### DIFF
--- a/server/routes/probationPractitionerReferrals/interventionProgressPresenter.ts
+++ b/server/routes/probationPractitionerReferrals/interventionProgressPresenter.ts
@@ -26,7 +26,7 @@ export default class InterventionProgressPresenter {
   }
 
   readonly text = {
-    title: utils.convertToTitleCase(this.serviceCategory.name),
+    title: `${utils.convertToTitleCase(this.serviceCategory.name)} progress`,
   }
 
   readonly hasSessions = this.actionPlanAppointments.length !== 0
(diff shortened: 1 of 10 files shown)


server/routes/probationPractitionerReferrals/interventionProgressPresenter.ts
```diff
@@ -0,0 +1,66 @@
+import { ActionPlanAppointment, SentReferral, ServiceCategory } from '../../services/interventionsService'
```
tomsmyers, line 1: there's quite a bit of duplication in this module and the SP version - not a blocker, but would be nice to reuse logic like 'referralAssigned' 'hasSessions' 'tagArgs' etc

server/routes/probationPractitionerReferrals/interventionProgressView.ts
```diff
@@ -0,0 +1,41 @@
...
+          { text: tagMacro(row.tagArgs as TagArgs) },
+          { html: `${row.linkHtml}` },
+        ]
+      }),
+    }
+  }
+
+  private readonly backLinkArgs = {
+    text: 'Back',
+    href: '/probation-practitioner/dashboard',
```
tomsmyers, line 26: does this currently exist?

tomsmyers (approved):
looks good !

server/routes/probationPractitionerReferrals/interventionProgressView.ts
Gweaton: Yeah, but at the moment it's just the start of the "Make a referral" journey

server/routes/probationPractitionerReferrals/interventionProgressPresenter.ts
Gweaton: Have added IC-1507 to look into this, as there might be some re-working we can do but I don't want to spend too much time on it now.

PR #220, opened 2021-04-07
IC-1423: Add record behaviour function and contract test

Gweaton:

Adds Interventions Service function to record a Service User's behaviour after a session, as well as mocks and stubs for integration tests.

To allow the Service Provider to record the SU's behaviour and notify the PP if there is anything noteworthy.

### DIFF
--- a/server/services/interventionsService.test.ts
+++ b/server/services/interventionsService.test.ts
@@ -1901,9 +1901,11 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
             sessionNumber: 2,
             appointmentTime: '2021-05-13T13:30:00+01:00',
             durationInMinutes: 60,
-            attendance: {
-              attended: 'late',
-              additionalAttendanceInformation: 'Alex missed the bus',
+            sessionFeedback: {
+              attendance: {
+                attended: 'late',
+                additionalAttendanceInformation: 'Alex missed the bus',
+              },
             },
           }),
           headers: {
@@ -1921,8 +1923,8 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
           additionalAttendanceInformation: 'Alex missed the bus',
         }
       )
-      expect(appointment.attendance!.attended).toEqual('late')
-      expect(appointment.attendance!.additionalAttendanceInformation).toEqual('Alex missed the bus')
+      expect(appointment.sessionFeedback!.attendance!.attended).toEqual('late')
+      expect(appointment.sessionFeedback!.attendance!.additionalAttendanceInformation).toEqual('Alex missed the bus')
     })
   })
 })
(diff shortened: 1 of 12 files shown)


server/services/interventionsService.test.ts
```diff
@@ -1925,6 +1925,53 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
...
+        state:
+          'an action plan with ID 345059d4-1697-467b-8914-fedec9957279 exists and has 2 2-hour appointments already',
+        uponReceiving:
+          'a POST request to set the behaviour for session 2 on action plan with ID 345059d4-1697-467b-8914-fedec9957279',
+        withRequest: {
+          method: 'POST',
+          path: '/action-plan/345059d4-1697-467b-8914-fedec9957279/appointment/2/record-behaviour',
+          body: {
+            behaviourDescription: 'Alex was well behaved',
+            notifyProbationPractitioner: 'no',
```
tomsmyers, line 1941: can this be a json boolean?

server/services/interventionsService.test.ts
```diff
@@ -1925,6 +1925,53 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
...
+          },
+          headers: { Accept: 'application/json', Authorization: `Bearer ${token}` },
+        },
+        willRespondWith: {
+          status: 200,
+          body: Matchers.like({
+            sessionNumber: 2,
+            appointmentTime: '2021-05-13T13:30:00+01:00',
+            durationInMinutes: 60,
+            behaviour: {
```
tomsmyers, line 1951: can this information be included in the attendance DTO? also do we need the notify pp boolean in the response DTO?

```    "attendance": {
        "attended": "late",
        "additionalAttendanceInformation": "charlie missed the bus",
        "behaviour": "charlie bit me"
    }```

**Review by tomsmyers — COMMENTED** [review]
questions on req and res DTOs

File: server/services/interventionsService.test.ts (hunk 2 above)
Comment (tomsmyers) [reply server/services/interventionsService.test.ts:1951]: my thinking on this is that the behaviour is part and parcel of the session attendance, since it happened whilst the SU was attending the session. it seems strange to me to have it be it's own thing in the response.

File: server/services/interventionsService.test.ts (hunk 1 above)
Comment (Gweaton) [reply server/services/interventionsService.test.ts:1941]: Yep, sure

File: server/services/interventionsService.test.ts (hunk 2 above)
Comment (Gweaton) [reply server/services/interventionsService.test.ts:1951]: For me, I felt it would be more useful to have attendance as "did they attend?" and behaviour as its own thing to make it easier to update one separate entity per page.

Why would we not want the notify boolean in there? As it's only for sending a notification to the PP on the backend?

Comment (Gweaton) [reply server/services/interventionsService.test.ts:1951]: For reference, here's what we've decided the DTO will look like:

```
sessionFeedback: {
  attendance: {
    attended: "late",
    additionalAttendanceInformation: null,
  },
  behaviour: {
    behaviourDescription: "bad",
    notifyProbationPractitioner: false,
  }
}
```

tomsmyers (approved):

PR #223, opened 2021-04-08
Extract calendar day input

lawrence-forooghian:

Extracts the code for validating and parsing a calendar day form input, which is currently all contained within the completion deadline form. It then makes some improvemens to this validator, to return more detailed error messages.

To be able to reuse the calendar day input logic for the upcoming "edit action plan appointment" page.

### DIFF
--- a/server/utils/forms/inputs/calendarDayInput.test.ts
+++ b/server/utils/forms/inputs/calendarDayInput.test.ts
@@ -37,9 +37,9 @@ describe(CalendarDayInput, () => {
       expect(result.error).toEqual({
         errors: [
           {
-            errorSummaryLinkedField: 'deadline-day',
-            formFields: ['deadline-day', 'deadline-month', 'deadline-year'],
-            message: 'The deadline must be a real date',
+            errorSummaryLinkedField: 'deadline-year',
+            formFields: ['deadline-year'],
+            message: 'The deadline must include a year',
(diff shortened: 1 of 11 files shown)


server/utils/forms/inputs/calendarDayInput.test.ts
```diff
@@ -45,6 +45,26 @@ describe(CalendarDayInput, () => {
...
+      })
+
+      const result = await new CalendarDayInput(request, 'deadline', messages).validate()
+
+      expect(result.error).toEqual({
+        errors: [
+          {
+            errorSummaryLinkedField: 'deadline-month',
+            formFields: ['deadline-month'],
+            message: 'The deadline must include a month',
```
tomsmyers, line 62: if only month were filled in would the 'highest priority' be year?

tomsmyers (approved):
found this hard to review since it's lots of copy paste between files, but it all looks good and i trust it works. nice work.

server/utils/forms/inputs/calendarDayInput.test.ts
lawrence-forooghian: Yep, I think that's how it'd behave. I haven't explored it in too much detail in the tests because we know we need to revisit this properly and already have a Jira task for it (IC-706).

PR #225, opened 2021-04-08
Add duration input class

lawrence-forooghian:
**Dependency: This PR is dependent on #223 – please review that one first.**

Adds a class for validating and parsing a duration form input, continuing the pattern introduced in #223.

To allow us to accept duration input on the upcoming "edit action plan appointment" page.

### DIFF
--- a/server/utils/forms/inputs/durationInput.ts
+++ b/server/utils/forms/inputs/durationInput.ts
@@ -1,0 +1,100 @@
+import * as ExpressValidator from 'express-validator'
+import { Request } from 'express'
+import Duration from '../../duration'
+import { FormValidationError } from '../../formValidationError'
+import { FormValidationResult } from '../formValidationResult'
+import FormUtils from '../../formUtils'
+
+export interface DurationErrorMessages {
+  empty: string
+  invalidDuration: string
+}
+
+export default class DurationInput {
+  constructor(
+    private readonly request: Request,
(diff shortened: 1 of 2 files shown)


server/utils/forms/inputs/durationInput.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    return {
+      value: this.duration!,
+      error: null,
+    }
+  }
+
+  private error(result: ExpressValidator.Result): FormValidationError | null {
+    const error = FormUtils.validationErrorFromResult(result)
```
Gweaton, line 39: I'm finding the re-use of the word/variable `error` here a bit difficult to parse - I think it could boil down to us referring to `error.errors` in a few places - is there a more user-friendly variable name for what's returned from `validationErrorFormResult` that encapsulates the fact that it could contain multiple errors?

Gweaton (approved):
This looks like a lot of work, well done.

I did find it a bit hard to review the first commit of the PR, as there is a _lot_ going on there, but I'm hoping I haven't missed anything.

server/utils/forms/inputs/durationInput.ts
lawrence-forooghian: Very good point - I've renamed it the variable to `formValidationError` to match the type.

PR #224, opened 2021-04-08
Add 12 hour time input class

lawrence-forooghian:
**Dependency: This PR is dependent on #223 – please review that one first.**

## What does this pull request do?

Adds a class for validating and parsing a 12-hour time form input, continuing the pattern introduced in #223.

## What is the intent behind these changes?

To allow us to accept time input on the upcoming "edit action plan appointment" page.

PR #226, opened 2021-04-08
Add 12 hour datetime input class

lawrence-forooghian:

Combines the calendar day input of #223 and the 12-hour time input of #224 to create a class that takes date / time inputs, validates them, and returns a JavaScript `Date`.

In addition to validating the individual date / time, it also checks that the date and time combination is valid (daylight savings time edge case but best to be safe).

To allow us to create a timestamp to send for the `appointmentTime` value on the upcoming "edit action plan appointment" page.

### DIFF
--- a/server/utils/forms/inputs/twelveHourBritishDateTimeInput.ts
+++ b/server/utils/forms/inputs/twelveHourBritishDateTimeInput.ts
@@ -1,0 +1,71 @@
+import { Request } from 'express'
+import CalendarDay from '../../calendarDay'
+import ClockTime from '../../clockTime'
+import { FormValidationError } from '../../formValidationError'
+import { FormValidationResult } from '../formValidationResult'
+import CalendarDayInput, { CalendarDayErrorMessages } from './calendarDayInput'
+import TwelveHourClockTimeInput, { TwelveHourClockTimeErrorMessages } from './twelveHourClockTimeInput'
+
+export interface TwelveHourBritishDateTimeErrorMessages {
+  calendarDay: CalendarDayErrorMessages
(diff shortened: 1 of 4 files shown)


server/utils/forms/inputs/twelveHourBritishDateTimeInput.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    return null
+  }
+
+  private date(dayResult: FormValidationResult<CalendarDay>, timeResult: FormValidationResult<ClockTime>): Date | null {
+    if (dayResult.value === null || timeResult.value === null) {
+      return null
+    }
+
+    const date = dayResult.value.atTimeInBritain(timeResult.value)
```
Gweaton, line 68: I'm not sure we need to worry about assigning this to a variable if we're returning it on the next line.

server/utils/forms/inputs/twelveHourBritishDateTimeInput.ts
```diff
@@ -0,0 +1,71 @@
...
+            message: this.errorMessages.invalidTime,
+          },
+        ],
+      }
+    }
+
+    return null
+  }
+
+  private date(dayResult: FormValidationResult<CalendarDay>, timeResult: FormValidationResult<ClockTime>): Date | null {
```
Gweaton, line 63: I wonder if there's a clearer method name for this - it's taking the results of validation functions and returning null if they're not valid, right?

It's wordy, but would something like `validBritishTimeOrNull` convey that a bit more clearly?

Gweaton (approved):
A couple of small things

PR #228, opened 2021-04-08
IC-1286: Service provider caseworker can schedule an appointment

lawrence-forooghian:

Allows a service provider caseworker to schedule an action plan appointment.

Specifically, adds the ability to actually update the appointment to the "edit action plan appointment" page originally introduced in #194, which currently only allows the user to view details that have already been set.

To allow a service provider to record the details of an appointment.

## Screenshot

image


tomsmyers (approved):
🤙

PR #229, opened 2021-04-08
🐛 Set timezone to Europe/London (successfully, this time)

sldblog:

This PR fixes a 🐛 about setting the timezone to Europe/London in the Docker build.

As part of a set of conversations with @mahalrmoj we've discovered that the UI app is currently running on UTC:
```
$ k exec -ti hmpps-interventions-ui-5fc67cffcd-z4kgx --namespace=hmpps-interventions-dev -- date
Thu Apr  8 14:48:13 UTC 2021
``` 

### How?

- b6dc7cb adds a "build-time test" -- verifies that the timezone file in fact exists
- 774f922 then installs the package that provides that timezone file

Result:
```
$ docker build --tag=ui .
...<snip>...

$ docker run --rm -ti ui date
Thu Apr  8 15:47:44 BST 2021
```

To make sure the service is running in Europe/London timezone. It has no material impact on our APIs, but it's the least surprising default :) -- At least it doesn't _look_ like we're setting the timezone when we aren't.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,8 @@ FROM node:14-alpine3.13 as base
 LABEL maintainer="HMPPS Digital Studio <info@digital.justice.gov.uk>"
 
 ENV TZ=Europe/London
-RUN test -e "/usr/share/zoneinfo/$TZ" && \
+RUN apk add --no-cache tzdata && \
+      test -e "/usr/share/zoneinfo/$TZ" && \
       ln -snf "/usr/share/zoneinfo/$TZ" /etc/localtime && \
       echo "$TZ" > /etc/timezone
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,9 @@ FROM node:14-alpine3.13 as base
 LABEL maintainer="HMPPS Digital Studio <info@digital.justice.gov.uk>"
 
 ENV TZ=Europe/London
-RUN ln -snf "/usr/share/zoneinfo/$TZ" /etc/localtime && echo "$TZ" > /etc/timezone
+RUN test -e "/usr/share/zoneinfo/$TZ" && \
+      ln -snf "/usr/share/zoneinfo/$TZ" /etc/localtime && \
+      echo "$TZ" > /etc/timezone
 
 RUN addgroup --gid 2000 --system appgroup && \
     adduser --uid 2000 --system appuser && \


Dockerfile
```diff
@@ -7,7 +7,10 @@ FROM node:14-alpine3.13 as base
 LABEL maintainer="HMPPS Digital Studio <info@digital.justice.gov.uk>"
 
 ENV TZ=Europe/London
-RUN ln -snf "/usr/share/zoneinfo/$TZ" /etc/localtime && echo "$TZ" > /etc/timezone
+RUN apk add --no-cache tzdata && \
+      test -e "/usr/share/zoneinfo/$TZ" && \
```
sldblog, line 11: Instead of this line, we could also add a "docker build test" that would test the timezone by running the `date` command:
```
$ docker run --rm -ti <built_tag> date | egrep "(GMT|BST)"
Thu Apr  8 15:55:29 BST 2021
# exit status 0
```
vs
```
$ k exec -ti hmpps-interventions-ui-5fc67cffcd-z4kgx -- date | egrep "(GMT|BST)"
# exit status 1, thus failing the build
```
However, this isn't trivial to chain after the docker build, before the push at the moment

lawrence-forooghian:
What impact does this have on the application? Do you have an example of how the container's time zone is affecting its behaviour?

tomsmyers:
without properly understanding the context - it doesn't seem appropriate to rely on the TZ of the container for anything at all. we have libraries already in use in this repo that allow formatting in whatever TZ you like. e.g. 

```
    const ukDate = getZonedTime(date, findTimeZone('Europe/London'))

```

sldblog:
Apologies, I skimmed over a few important points

1. The most important bit is fixing the false promise of the `Dockerfile`: it sets `TZ` to London, so if it does that, that should work.
2. The second is that the service inherited this setting from the Kotlin template and it bothers my sense of consistency. It does have a suggestion open to switch to UTC by default. (https://github.com/ministryofjustice/hmpps-template-kotlin/issues/29)
3. Distant third (I'm not precious about this) is what we get back from the "now" instant. We're deployed in the London region, not having to work in other countries, so it can be the same time we see on the clock. This is really minor though.

tomsmyers (approved):
agree with your comments david. we shouldn't rely on this, but setting it correctly can't be a bad thing.

PR #231, opened 2021-04-09
IC-1425: Add page for recording an SU's behaviour during a session

Gweaton:

Adds page following attendance page where SP can record the SU's behaviour in a session.

This currently takes the user to the confirmation screen afterwards, but next up we'll be adding a "Check your answers" between these two pages.

If the SP marks the SU as "not attended", the user will go straight to the confirmation screen and skip this new behaviour page.

To allow the SP to record the SU's behaviour in the session.

## Screenshot (composed of 2 separate screenshots due to size of image in E2E tests, apologies for slight wonkiness)

image
image

### DIFF
--- a/server/routes/serviceProviderReferrals/postSessionBehaviourFeedbackPresenter.test.ts
+++ b/server/routes/serviceProviderReferrals/postSessionBehaviourFeedbackPresenter.test.ts
@@ -22,6 +22,68 @@ describe(PostSessionBehaviourFeedbackPresenter, () => {
         },
       })
     })
+
+    describe('when there are errors', () => {
+      it('populates the error messages for the fields with errors', () => {
+        const appointment = actionPlanAppointmentFactory.build()
+        const serviceUser = deliusServiceUserFactory.build()
+        const presenter = new PostSessionBehaviourFeedbackPresenter(appointment, serviceUser, {
+          errors: [
+            {
+              formFields: ['behaviour-description'],
+              errorSummaryLinkedField: 'behaviour-description',
+              message: 'behaviour msg',
+            },
+            {
+              formFields: ['notify-probation-practitioner'],
+              errorSummaryLinkedField: 'notify-probation-practitioner',
+              message: 'notify pp msg',
+            },
+          ],
+        })
+
+        expect(presenter.fields.behaviourDescription.errorMessage).toEqual('behaviour msg')
+        expect(presenter.fields.notifyProbationPractitioner.errorMessage).toEqual('notify pp msg')
+      })
+    })
+  })
+
+  describe('errorSummary', () => {
+    describe('when error is null', () => {
+      it('returns null', () => {
+        const appointment = actionPlanAppointmentFactory.build()
+        const serviceUser = deliusServiceUserFactory.build({ firstName: 'Alex' })
+        const presenter = new PostSessionBehaviourFeedbackPresenter(appointment, serviceUser)
+
+        expect(presenter.errorSummary).toBeNull()
+      })
+    })
+
+    describe('when error is not null', () => {
+      it('returns a summary of the errors sorted into the order their fields appear on the page', () => {
+        const appointment = actionPlanAppointmentFactory.build()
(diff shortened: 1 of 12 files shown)


server/routes/serviceProviderReferrals/postSessionBehaviourFeedbackPresenter.ts
```diff
@@ -0,0 +1,40 @@
...
+      hint: 'Select one option',
+    },
+  }
+
+  readonly serviceUserBannerPresenter = new ServiceUserBannerPresenter(this.serviceUser)
+
+  private readonly utils = new PresenterUtils(this.userInputData)
+
+  readonly fields = {
+    behaviourDescriptionValue: new PresenterUtils(this.userInputData).stringValue(
```
lawrence-forooghian, line 43: If you want to be consistent with the field below, I think this should be called `behaviourDescription`.

server/routes/serviceProviderReferrals/postSessionBehaviourFeedbackPresenter.test.ts
```diff
@@ -14,14 +14,84 @@ describe(PostSessionBehaviourFeedbackPresenter, () => {
         behaviourDescription: {
           question: `Describe Alex's behaviour in this session`,
           hint: 'For example, consider how well-engaged they were and what their body language was like.',
+          errorMessage: null,
```
lawrence-forooghian, line 17: We've normally been nesting these inside the `fields` object like this:

```
behaviourDescription: {
  value: 'foo',
  errorMessage: 'blah'
}
```

— could we keep this consistent?

server/routes/serviceProviderReferrals/serviceProviderReferralsController.test.ts
```diff
@@ -676,35 +676,70 @@ describe('GET /service-provider/action-plan/:actionPlanId/appointment/:sessionNu
...
+        .expect(
+          'Location',
+          `/service-provider/action-plan/${actionPlan.id}/appointment/${updatedAppointment.sessionNumber}/post-session-feedback/behaviour`
+        )
     })
+  })
 
-    const actionPlan = actionPlanFactory.build()
+  describe('when the Service Provider marks the Service User as not having attended the session', () => {
+    it('makes a request to the interventions service to record the Service User‘s attendance and redirects to the confirmation page', async () => {
```
lawrence-forooghian, line 713: Do you think it's worth adding an integration test for this journey, too? (Question, not necessarily a suggestion.)

lawrence-forooghian (approved):
Looks good to me, a few minor points. Nice and easy to review, thanks.

server/routes/serviceProviderReferrals/postSessionBehaviourFeedbackPresenter.test.ts
Gweaton: Just looking at `rarDaysPresenter`, `actionPlanNumberOfSessionsPresenter` and `riskInformationPresenter` they're nested in `text` - can you show me an example where they're nested under `fields`?

lawrence-forooghian: Yeah, it might have been different in the past. But recently, I added some new methods to `PresenterUtils`, which return a "field presenter" object which contains both the value and the error message. See them in use in `EditSessionPresenter`.

server/routes/serviceProviderReferrals/serviceProviderReferralsController.test.ts
Gweaton: Yeah I think it probably is - the more the merrier, if nothing else...

server/routes/serviceProviderReferrals/postSessionBehaviourFeedbackPresenter.test.ts
Gweaton: have updated this to how I _think_ you mean

PR #237, opened 2021-04-13
IC-1526 add service methods and contracts for submitting session feedback

tomsmyers:

adds contract for submitting session feedback

the contract is there, but the unit test part is commented out because it doesn't exist yet.

validate the API endpoint for submitting session feedback.

### DIFF
--- a/server/routes/probationPractitionerReferrals/interventionProgressPresenter.ts
+++ b/server/routes/probationPractitionerReferrals/interventionProgressPresenter.ts
@@ -43,7 +43,7 @@ export default class InterventionProgressPresenter {
         sessionNumber: appointment.sessionNumber,
         appointmentTime: DateUtils.formatDateTimeOrEmptyString(appointment.appointmentTime),
         tagArgs: this.tagArgs(appointment),
-        linkHtml: appointment.sessionFeedback?.attendance ? `<a class="govuk-link" href="#">View</a>` : '',
+        linkHtml: appointment.sessionFeedback?.submitted ? `<a class="govuk-link" href="#">View</a>` : '',
       }
     })
   }
--- a/integration_tests/integration/monitor.spec.js
+++ b/integration_tests/integration/monitor.spec.js
@@ -32,25 +32,15 @@ describe('Probation Practitioner monitor journey', () => {
       })
 
       const appointments = [
-        actionPlanAppointmentFactory.build({
+        actionPlanAppointmentFactory.attended('yes').build({
           sessionNumber: 1,
           appointmentTime: '2021-03-24T09:02:02Z',
           durationInMinutes: 75,
-          sessionFeedback: {
-            attendance: {
-              attended: 'yes',
-            },
-          },
         }),
-        actionPlanAppointmentFactory.build({
+        actionPlanAppointmentFactory.attended('no').build({
           sessionNumber: 2,
           appointmentTime: '2021-04-30T09:02:02Z',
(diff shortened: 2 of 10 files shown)


Gweaton (commented):
In previous PRs, we've been adding the data-fetching function (commented out) and the Mock functions for this (`mockApis/interventionsService.ts`) - would you have time to add them here too?

lawrence-forooghian:
> In previous PRs, we've been adding the data-fetching function (commented out) and the Mock functions for this (`mockApis/interventionsService.ts`) - would you have time to add them here too?

Can we add the related Cypress methods too?

tomsmyers:
@lawrence-forooghian @Gweaton sure i don't mind doing this, but in the future i think isolating just the interaction would be good practise, since it allows the reviewer to focus only on the contract. the stubs seems to me to be more involved with the writing of code in the service layer i.e. implementing the UI functionality that requires the interaction described here.

lawrence-forooghian:
> @lawrence-forooghian @Gweaton sure i don't mind doing this, but in the future i think isolating just the interaction would be good practise, since it allows the reviewer to focus only on the contract. the stubs seems to me to be more involved with the writing of code in the service layer i.e. implementing the UI functionality that requires the interaction described here.

That's true, but on the other hand it's easier to write these mocks when the API contract is fresh in your mind, especially if a different person is writing the contracts to actually building some feature around them.

server/routes/probationPractitionerReferrals/interventionProgressPresenter.ts
```diff
@@ -43,7 +43,7 @@ export default class InterventionProgressPresenter {
         sessionNumber: appointment.sessionNumber,
         appointmentTime: DateUtils.formatDateTimeOrEmptyString(appointment.appointmentTime),
         tagArgs: this.tagArgs(appointment),
-        linkHtml: appointment.sessionFeedback?.attendance ? `<a class="govuk-link" href="#">View</a>` : '',
+        linkHtml: appointment.sessionFeedback?.submitted ? `<a class="govuk-link" href="#">View</a>` : '',
```
tomsmyers, line 46: hooray for unit tests 🎉

Gweaton: This is a functional change - would it be too much of a pain to split it out into its own commit?

Gweaton (commented):
Thanks for taking care of this and adding the contract / implementation.

The "update" commit does quite a lot:
- adds a factory method for submitted
- extracts `Attendance` into an interface
- Changes the structure of `sessionFeedback`
- Changes the functionality of when the link should change on the PP side.

I'm ok with most of the above (although would ideally prefer it to be slightly broken up), but would you mind updating the commit message to explain the changes?

It'd be good to break that last thing (changing the display of the link based on "submitted", rather than "attendance") into its own commit, or leaving it out entirely (I was going to take care of it in my branch.

server/routes/probationPractitionerReferrals/interventionProgressPresenter.ts
tomsmyers: since `attendance` is no longer optional in the updated model, this change is nessecary here. we could alter this to look something like `sessionFeedback?.attendance.attended != null`, but then i don't think it's really correct.

happy to split the commit out, but fixing getting the tests green in both commits will be a pain. can i just expand on the commit message?

Gweaton (approved):

PR #239, opened 2021-04-15
IC-1424: Clarify entire Appointment (including null values) present in every response

Gweaton:

Adds whole response for Appointment for each contract test, including `null` values to make it clear what we actually receive from each response.

When developing the "check your answers page" and relying on these fields, I found it
difficult to determine whether the session feedback fields (attendance
and behaviour) would be present in the response from the API or not,
particularly before they'd been sent vs if they had been skipped as an
optional answer.

By ensuring we use the `actionPlanAppointmentFactory` in this step, we
know that the fields will be `null` but present on the response object
by default, as that's how they're configured in the factory.

I've also added the missing parts of the response to make sure they're
in every contract test to be totally clear what's coming back from every
request.

### DIFF
--- a/server/services/interventionsService.test.ts
+++ b/server/services/interventionsService.test.ts
@@ -1675,21 +1675,21 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
 
   describe('getActionPlanAppointments', () => {
     const actionPlanAppointments = [
-      {
+      actionPlanAppointmentFactory.build({
         sessionNumber: 1,
         appointmentTime: '2021-05-13T12:30:00Z',
         durationInMinutes: 120,
-      },
-      {
+      }),
+      actionPlanAppointmentFactory.build({
         sessionNumber: 2,
         appointmentTime: '2021-05-20T12:30:00Z',
         durationInMinutes: 120,
-      },
-      {
+      }),
+      actionPlanAppointmentFactory.build({
         sessionNumber: 3,
         appointmentTime: '2021-05-27T12:30:00Z',
         durationInMinutes: 120,
-      },
+      }),
     ]
 
     beforeEach(async () => {
@@ -1719,11 +1719,11 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
   })
 
   describe('getActionPlanAppointment', () => {
-    const actionPlanAppointment = {
+    const actionPlanAppointment = actionPlanAppointmentFactory.build({
       sessionNumber: 1,
       appointmentTime: '2021-05-13T12:30:00Z',
       durationInMinutes: 120,
-    }
+    })
 
     beforeEach(async () => {
       await provider.addInteraction({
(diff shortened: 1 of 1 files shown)


server/services/interventionsService.test.ts
```diff
@@ -1675,21 +1675,21 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
 
   describe('getActionPlanAppointments', () => {
     const actionPlanAppointments = [
-      {
+      actionPlanAppointmentFactory.build({
```
lawrence-forooghian, line 1678: I think we might need to be more explicit here, in a couple of places:

- The `state` of the interaction, which is currently `'an action plan with ID e5ed2f80-dfe2-4bf3-b5c4-d8d4486e963d exists and it has 3 scheduled appointments'`. This says nothing about whether the appointments have had their attendance / behaviour recorded yet, so we can't expect those fields to be `null` or `submitted` to be `false`. So if we do want to include all those fields in the expected response from the interaction, then we need to say something like `'and it has 3 scheduled appointments, none of which have had feedback submitted yet'`.
- The state in which we're asking the factory to build an appointment. I don't think we should make too many assumptions about the value created by `actionPlanAppointmentFactory.build()` — here we're assuming that it returns appointments that haven't had their feedback recorded yet, but I don't think we can make that assumption. Perhaps adding a `scheduled()` trait to the factory would make it clear that these appointments are scheduled but that feedback hasn't yet been given.

lawrence-forooghian (commented):
👍 I like the idea of using the factories where possible in these contract tests, since it makes it more likely that the data that we're using to test the rest of the application actually matches what's being returned by the API.

server/services/interventionsService.test.ts
lawrence-forooghian: Perhaps I'm overthinking this actually, do you have any thoughts?

Gweaton: I think what confused me (and led me to open this PR) was that despite the way we've designed our `Appointment` models with nullable values, I wasn't quite sure from reading the contracts what came back from the API (i.e. that there were always `null` values on everything on the nested `sessionFeedback` object.

What confused me was that the expected response didn't fail the tests when these fields weren't present on the response object we'd written - I think using the factories does make this clearer, as it's clear we're returning a whole object, rather than the handwritten JSON response (even though we hide all the fields in that `build` method). I think that as long as it's clear that we're returning our whole internal representation of an `Appointment` here, this is an improvement.

I do think there's something to be said for adding a state for the attendance feedback test to specify that it `doesn't have feedback submitted yet` (which I'll add in) but for the initial `getActionPlanAppointments` I don't think there's that much of a need to add that state, as we're not really asserting anything on that feedback yet (and it can be called when session feedback _is_ present in the response)

server/services/interventionsService.test.ts
```diff
@@ -1839,7 +1839,7 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
     it('returns an updated action plan appointment with the service user‘s attendance', async () => {
       await provider.addInteraction({
         state:
-          'an action plan with ID 345059d4-1697-467b-8914-fedec9957279 exists and has 2 2-hour appointments already',
+          'an action plan with ID 345059d4-1697-467b-8914-fedec9957279 exists and has an appointment for which no session feedback has been recorded',
```
Gweaton, line 1842: @tomsmyers @rgforsyth I think this change will break the contracts, would one of you mind updating it on the API side?

lawrence-forooghian (approved):

PR #240, opened 2021-04-15
IC-1424: Add check your answers page to post session feedback journey

Gweaton:

Adds check your answers page to post session feedback journey.

If the SP marks the SU as having not attended, they'll skip the behaviour question and go straight to the check your answers page (and won't see the behaviour questions on that page).

Clicking "confirm" on the check your answers page marks the referral as `submitted` on the backend.

To allow users to check their answers before submitting attendance/behaviour feedback.

### Service user did not attend

image

### Service user attended

image

### DIFF
--- a/server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
+++ b/server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
@@ -37,6 +37,8 @@ import PostSessionFeedbackConfirmationView from './postSessionFeedbackConfirmati
 import PostSessionBehaviourFeedbackPresenter from './postSessionBehaviourFeedbackPresenter'
 import PostSessionBehaviourFeedbackView from './postSessionBehaviourFeedbackView'
 import PostSessionBehaviourFeedbackForm from './postSessionBehaviourFeedbackForm'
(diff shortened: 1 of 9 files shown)


server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
```diff
@@ -544,6 +544,14 @@ export default class ServiceProviderReferralsController {
       Number(sessionNumber)
     )
 
+    if (req.method === 'POST') {
```
lawrence-forooghian, line 547: I think that we've been using this pattern of one method for the `GET` and the `POST` in the cases where both of them render the same view. Here it's a bit different, since the `POST` doesn't have any functionality in common with the `GET`. Perhaps it's best to put it in a separate method? Also, I wonder whether the correct URL for the action would be `/service-provider/action-plan/:actionPlanId/appointment/:sessionNumber/post-session-feedback/submit`?

lawrence-forooghian (approved):
Looks good, just the one comment.

server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
Gweaton: Yep, posting to `check-your-answers` felt very weird, this makes a lot more sense!

PR #241, opened 2021-04-15
IC-1529: Update intervention progress statuses after feedback

Gweaton:

Updates the PP and SP intervention progress pages with the right statuses once post session feedback has been submitted.

This currently doesn't display the red "awaiting feedback" status, or the green "today" ones, as I believe we were hesitant to start introducing statuses based on dates.

To give the SP/PP a quick overview of the stage of each session.

Screenshots:

image

image

### DIFF
--- a/server/routes/probationPractitionerReferrals/interventionProgressPresenter.ts
+++ b/server/routes/probationPractitionerReferrals/interventionProgressPresenter.ts
@@ -3,6 +3,8 @@ import utils from '../../utils/utils'
 import ReferralOverviewPagePresenter, { ReferralOverviewPageSection } from '../shared/referralOverviewPagePresenter'
 import { DeliusServiceUser } from '../../services/communityApiService'
 import DateUtils from '../../utils/dateUtils'
+import sessionStatus, { SessionStatus } from '../../utils/sessionStatus'
+import SessionStatusPresenter from '../shared/sessionStatusPresenter'
 
 export default class InterventionProgressPresenter {
   referralOverviewPagePresenter: ReferralOverviewPagePresenter
@@ -39,30 +41,46 @@ export default class InterventionProgressPresenter {
     }
 
     return this.actionPlanAppointments.map(appointment => {
+      const sessionTableParams = this.sessionTableParams(appointment)
+
       return {
         sessionNumber: appointment.sessionNumber,
         appointmentTime: DateUtils.formatDateTimeOrEmptyString(appointment.appointmentTime),
-        tagArgs: this.tagArgs(appointment),
-        linkHtml: appointment.sessionFeedback?.submitted ? `<a class="govuk-link" href="#">View</a>` : '',
+        tagArgs: { text: sessionTableParams.text, classes: sessionTableParams.tagClass },
+        linkHtml: sessionTableParams.linkHTML,
       }
     })
   }
 
-  private tagArgs(appointment: ActionPlanAppointment): Record<string, unknown> {
-    const sessionFeedbackAttendance = appointment.sessionFeedback?.attendance
+  private sessionTableParams(appointment: ActionPlanAppointment): { text: string; tagClass: string; linkHTML: string } {
+    const status = sessionStatus.forAppointment(appointment)
+    const presenter = new SessionStatusPresenter(status)
 
-    if (sessionFeedbackAttendance?.attended === 'no') {
-      return { text: 'FAILURE TO ATTEND', classes: 'govuk-tag--purple' }
+    switch (status) {
+      case SessionStatus.didNotAttend:
+        return {
+          text: presenter.text,
+          tagClass: presenter.tagClass,
+          linkHTML: `<a class="govuk-link" href="#">View feedback form</a>`,
+        }
+      case SessionStatus.completed:
+        return {
+          text: presenter.text,
+          tagClass: presenter.tagClass,
+          linkHTML: `<a class="govuk-link" href="#">View feedback form</a>`,
+        }
+      case SessionStatus.scheduled:
+        return {
+          text: presenter.text,
+          tagClass: presenter.tagClass,
+          linkHTML: '',
+        }
+      default:
+        return {
+          text: presenter.text,
+          tagClass: presenter.tagClass,
+          linkHTML: '',
+        }
     }
-
-    if (sessionFeedbackAttendance?.attended === 'yes' || sessionFeedbackAttendance?.attended === 'late') {
-      return { text: 'COMPLETED', classes: 'govuk-tag--green' }
-    }
-
-    if (appointment.appointmentTime) {
-      return { text: 'SCHEDULED', classes: 'govuk-tag--blue' }
-    }
-
-    return { text: 'NOT SCHEDULED', classes: 'govuk-tag--grey' }
   }
 }
--- a/server/routes/serviceProviderReferrals/interventionProgressPresenter.ts
+++ b/server/routes/serviceProviderReferrals/interventionProgressPresenter.ts
@@ -3,6 +3,8 @@ import utils from '../../utils/utils'
 import ReferralOverviewPagePresenter, { ReferralOverviewPageSection } from '../shared/referralOverviewPagePresenter'
 import { DeliusServiceUser } from '../../services/communityApiService'
 import DateUtils from '../../utils/dateUtils'
+import sessionStatus, { SessionStatus } from '../../utils/sessionStatus'
+import SessionStatusPresenter from '../shared/sessionStatusPresenter'
 
 export default class InterventionProgressPresenter {
   referralOverviewPagePresenter: ReferralOverviewPagePresenter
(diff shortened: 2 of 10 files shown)


server/routes/serviceProviderReferrals/interventionProgressPresenter.ts
```diff
@@ -53,18 +54,46 @@ export default class InterventionProgressPresenter {
...
-          ? `<a class="govuk-link" href="${editUrl}">Reschedule session</a><br><a class="govuk-link" href="/service-provider/action-plan/${this.actionPlan?.id}/appointment/${appointment.sessionNumber}/post-session-feedback/attendance">Give feedback</a>`
-          : `<a class="govuk-link" href="${editUrl}">Edit session details</a>`,
+        tagArgs: { text: sessionTableParams.text, classes: sessionTableParams.classes },
+        linkHtml: sessionTableParams.linkHTML,
       }
     })
   }
+
+  private sessionTableParams(appointment: ActionPlanAppointment): { text: string; classes: string; linkHTML: string } {
+    const sessionFeedbackAttendance = appointment.sessionFeedback.attendance
```
lawrence-forooghian, line 69: As you say in your next commit, there's some heavy overlap between this and the PP page, but they differ in the HTML used for the links.

I think that there are perhaps three separate concerns, which this method is mixing:

1. Given an appointment, what is its status? (all the `if` statements here)
2. Given an appointment status, what text and what colour tag should be displayed?
3. Given an appointment status, what link should be displayed?

The first two should be shared between the SP and PP pages, the third possibly not.

So how about this?

In `sessionStatus.ts`:

```typescript
enum SessionStatus {
  notScheduled,
  scheduled,
  completed,
  didNotAttend,
}

export default {
  forAppointment: (appointment: ActionPlanAppointment) => {
    // all the if statements to determine the status from the appointment and return a `SessionStatus` value
  }
}
```

Then something to encapsulate the shared presentation logic:

```typescript
class SessionStatusPresenter {
  constructor(private readonly status: SessionStatus) {}

  get text(): string {
    switch (this.status) {
      case SessionStatus.notAttended:
        return 'not attended'
      // etc
    }
  }

  get tagClass(): string {
    switch (this.status) {
      case SessionStatus.notAttended:
        return 'govuk-tag--grey'
      // etc
    }
  }

}
```

And then in the presenter you'd do something like

```typescript
private sessionTableParams(appointment: ActionPlanAppointment): { text: string; classes: string; linkHTML: string } {
  const status = sessionStatus.forAppointment(appointment)
  const presenter = new SessionStatusPresenter(status)

  switch (status) {
    case SessionStatus.notAttended:
      return { ...presenter, linkHTML: `<a class="govuk-link" href="#">View feedback form</a>` }
    // etc
  }
}
```

Gweaton: This was really helpful, and really satisfying - thanks for that. Have made these changes

server/routes/serviceProviderReferrals/interventionProgressPresenter.ts
```diff
@@ -53,18 +55,48 @@ export default class InterventionProgressPresenter {
...
+  private sessionTableParams(appointment: ActionPlanAppointment): { text: string; tagClass: string; linkHTML: string } {
+    const status = sessionStatus.forAppointment(appointment)
+    const presenter = new SessionStatusPresenter(status)
+
+    const editUrl = `/service-provider/action-plan/${this.actionPlan!.id}/sessions/${appointment.sessionNumber}/edit`
+
+    switch (status) {
+      case SessionStatus.didNotAttend:
+        return {
+          text: presenter.text,
```
lawrence-forooghian, line 78: Did the `{ ...presenter, linkHTML: 'blah blah' }` suggestion not work here? It would be more succinct.

server/utils/sessionStatusTags.ts
```diff
@@ -0,0 +1,18 @@
+export default {
+  notScheduled: {
+    text: 'NOT SCHEDULED',
```
lawrence-forooghian, line 3: I think we can get rid of this file now?

server/routes/serviceProviderReferrals/sessionStatusPresenter.test.ts
```diff
@@ -0,0 +1,36 @@
...
+        tagClass: 'govuk-tag--green',
+      },
+      {
+        status: SessionStatus.notScheduled,
+        text: 'not scheduled',
+        tagClass: 'govuk-tag--grey',
+      },
+    ]
+
+    expectedResults.forEach(({ status, text, tagClass }) => {
```
lawrence-forooghian, line 29: Jest has [built-in support](https://jestjs.io/docs/api#testeachtablename-fn-timeout) for this kind of table testing - I'm not exactly sure what the difference is with writing the loop yourself, but might be worth a look?

server/utils/sessionStatus.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export enum SessionStatus {
+  notScheduled,
+  scheduled,
+  completed,
+  didNotAttend,
+}
+
+export default {
+  forAppointment: (appointment: ActionPlanAppointment): SessionStatus => {
```
lawrence-forooghian, line 11: I think it would be good to test this - it's a self-contained bit of logic that will probably only get more complex as the service evolves. Perhaps worth a separate commit too.

server/routes/probationPractitionerReferrals/interventionProgressPresenter.ts
```diff
@@ -39,30 +41,46 @@ export default class InterventionProgressPresenter {
...
+    const presenter = new SessionStatusPresenter(status)
 
-    if (appointment.appointmentTime) {
-      return { text: 'SCHEDULED', classes: 'govuk-tag--blue' }
+    switch (status) {
+      case SessionStatus.didNotAttend:
+        return {
+          text: presenter.text,
+          tagClass: presenter.tagClass,
+          linkHTML: `<a class="govuk-link" href="#">View feedback form</a>`,
```
lawrence-forooghian, line 64: > This isn't very DRY (there's a lot we can re-use from the SP version of
this) but until we know what the links we're going to be using look
like, I'm keeping them separate.

Looks like this bit of the commit message is no longer applicable?

lawrence-forooghian (changes requested):

server/routes/serviceProviderReferrals/interventionProgressPresenter.ts
Gweaton: I get `Type '{ linkHTML: string; }' is missing the following properties from type '{ text: string; tagClass: string; linkHTML: string; }': text, tagClass` - it seems as though it's not picking up either of the fields.

Interestingly, it doesn't complain if I do:
```
const { text, tagClass } = new SessionStatusPresenter(status)
{
  ...{ text, tagClass },
  linkHTML: `<a class="govuk-link" href="#">View feedback form</a>`,
}
```

Do you think that's at all cleaner?


My theory _was_ that this was happening because `new SessionStatusPresenter(status)` doesn't return an object literal e.g. `{ text: "", tagClass: ""}`, it returns some other type of object, but that doesn't explain why the above is acceptable?

server/utils/sessionStatusTags.ts
Gweaton: Oops, sorry

server/routes/serviceProviderReferrals/sessionStatusPresenter.test.ts
Gweaton: I didn't know about this, it's pretty neat.

server/utils/sessionStatus.ts
Gweaton: Have added a test for this and split into its own commit.

server/routes/serviceProviderReferrals/interventionProgressPresenter.ts
lawrence-forooghian: It seems to be because the presenter is using getters to implement the `text` and `tagClass` properties, and it seems like the spread operator (`...`) doesn't gather properties that are defined by getters. Changing the presenter implementation to use normal properties (see example below) makes it compile, but I don't think it matters.

```typescript
export default class SessionStatusPresenter {
  constructor(private readonly status: SessionStatus) {}

  readonly text = (() => {
    switch (this.status) {
      case SessionStatus.didNotAttend:
        return 'did not attend'
      case SessionStatus.scheduled:
        return 'scheduled'
      case SessionStatus.completed:
        return 'completed'
      default:
        return 'not scheduled'
    }
  })()

  readonly tagClass = (() => {
    switch (this.status) {
      case SessionStatus.didNotAttend:
        return 'govuk-tag--purple'
      case SessionStatus.scheduled:
        return 'govuk-tag--blue'
      case SessionStatus.completed:
        return 'govuk-tag--green'
      default:
        return 'govuk-tag--grey'
    }
  })()
}
```

lawrence-forooghian (approved):

PR #242, opened 2021-04-16
Add Sentry exception monitoring

sldblog:

Largely following the steps at https://docs.sentry.io/platforms/node/guides/express/

Sentry's configuration:
- the client key (called "DSN"), coming from `sentry.ui_dsn` secret from Kubernetes
- the environment, so it's correctly tagging where it happened, configured via `ENV` variable

The hardcoded environment variables are now replaced with ranging through the variables defined in `values-{env}.yml` -- similar to what the service does

Enable exception monitoring on the UI side as well, following https://github.com/ministryofjustice/hmpps-interventions-service/pull/171

### DIFF
--- a/helm_deploy/hmpps-interventions-ui/templates/_envs.tpl
+++ b/helm_deploy/hmpps-interventions-ui/templates/_envs.tpl
@@ -55,26 +55,10 @@ env:
         name: sentry
         key: ui_dsn
 
-  - name: REDIS_TLS_ENABLED
-    value: {{ .Values.env.REDIS_TLS_ENABLED | quote }}
-
-  - name: HMPPS_AUTH_URL
-    value: {{ .Values.env.HMPPS_AUTH_URL | quote }}
-
-  - name: COMMUNITY_API_URL
-    value: {{ .Values.env.COMMUNITY_API_URL | quote }}
-
-  - name: OFFENDER_ASSESSMENTS_API_URL
-    value: {{ .Values.env.OFFENDER_ASSESSMENTS_API_URL | quote }}
-
-  - name: INTERVENTIONS_SERVICE_URL
-    value: {{ .Values.env.INTERVENTIONS_SERVICE_URL | quote }}
-
-  - name: TOKEN_VERIFICATION_API_URL
-    value: {{ .Values.env.TOKEN_VERIFICATION_API_URL | quote }}
-
-  - name: TOKEN_VERIFICATION_ENABLED
-    value: {{ .Values.env.TOKEN_VERIFICATION_ENABLED | quote }}
+  {{ range $key, $value := .Values.env }}
+  - name: {{ $key }}
+    value: {{ $value | quote }}
+  {{ end }}
 
   - name: NODE_ENV
     value: production
--- a/helm_deploy/hmpps-interventions-ui/templates/_envs.tpl
+++ b/helm_deploy/hmpps-interventions-ui/templates/_envs.tpl
@@ -49,6 +49,12 @@ env:
         name: elasticache-redis
         key: auth_token
 
+  - name: SENTRY_DSN
+    valueFrom:
+      secretKeyRef:
+        name: sentry
+        key: ui_dsn
+
   - name: REDIS_TLS_ENABLED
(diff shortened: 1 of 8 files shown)


server/app.ts
```diff
@@ -51,6 +52,12 @@ export default function createApp(
   // Server Configuration
   app.set('port', config.port)
 
+  // Reads config from SENTRY_DSN env variable, if exists
+  Sentry.init({ environment: config.sentry.environment })
```
sldblog, line 56: I tried to add this in `server/config.ts` as `dsn: get('SENTRY_DSN', 'dummy_dsn', requiredInProduction)` but ran into problems:

1. if the DSN is invalid, Sentry will actively refuse it, leading to boot errors
2. I cannot have `null` values with the `get` function and it did not feel like adding optionality into it is a good idea

So I'm going with the [documented effect](https://docs.sentry.io/platforms/node/guides/express/configuration/options/#dsn):
> The _DSN_ tells the SDK where to send the events. If this value is not provided, the SDK will try to read it from the `SENTRY_DSN` environment variable. If that variable also does not exist, the SDK will just not send any events.

helm_deploy/values-dev.yaml
```diff
@@ -15,6 +15,7 @@ ingress:
   path: /
 
 env:
+  ENV: dev
```
tomsmyers, line 18: we already have an env var that does this job - `NODE_ENV`. it's set in the dockerfile. can we reuse that for this purpose?

tomsmyers: unless this is specifically for sentry - in which case can we rename to `SENTRY_ENV` or similar?

tomsmyers (commented):
one question on env var usage

helm_deploy/values-dev.yaml
sldblog: Good question, I was thinking
- I named this `SENTRY_ENV` in the service code, but semantically, I'm describing what environment is the application is running in (~ `APP_ENV`)
- `NODE_ENV` would be similar to `RAILS_ENV`, `MIX_ENV` etc. typically used in other frameworks, so I'd be happy to use that -- is there a change in behaviour if it's not set to `production` everywhere?

sldblog: specifically, I'm concerned with the semantic difference

we have code like this, which seems to allude to "how" the app is running, locally or during testing
```
process.env.NODE_ENV === 'development' || process.env.NODE_ENV === 'test'
```

Whereas this new variable is more about what type of 'deployed' environment it is. It's not _specific_ to Sentry, but we do use it here first (I imagine AppInsights would be also interested?)

So I guess, which one would we prefer, `SENTRY_ENV` or `DEPLOYMENT_ENV`?

helm_deploy/hmpps-interventions-ui/templates/_envs.tpl
```diff
@@ -49,26 +49,16 @@ env:
         name: elasticache-redis
         key: auth_token
 
-  - name: REDIS_TLS_ENABLED
```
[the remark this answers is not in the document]
sldblog: done in 9a42a85aa91a0a0bf8bb1f9f9506ea8e7d86806a

helm_deploy/values-dev.yaml
sldblog: Does 62ad70dcd848256eca5d65e1088e96f7a99867e4 make sense?

server/app.ts
```diff
@@ -173,7 +180,11 @@ export default function createApp(
   )
 
   app.use((req, res, next) => next(createError(404, 'Not found')))
+
   app.use(authErrorHandler)
+
+  // The Sentry error handler must be before any other error middleware and after all controllers
+  app.use(Sentry.Handlers.errorHandler())
   app.use(errorHandler(config.production))
```
sldblog, line 188: Does this placement for `Sentry.Handlers.errorHandler` make sense?

tomsmyers (approved):
all looks good - thanks for updating.

sldblog:
🎩 Thank you!

PR #244, opened 2021-04-16
IC-1529: Move sesson feedback into `sessionDelivery` directory

Gweaton:

Moves session feedback into a `sessionDelivery/postSessionFeedback` directory.

We can re-use the check your answers page in the PP journey, which uses the presenters in this directory. I'd quite like us to re-structure the directories in this project based on section of the journey e.g. session delivery, intervention progress, rather than everything in one big SP/PP-focused directory, as there are a few places we can re-use components across the two sides.

## Next steps

This is a starting point, and if we think this is a good route to go down, we can continue restructuring this. If people think this is a good approach, I'd like to move the following into this directory:
- Creating an action plan
- Assigning referrals
- Adding/editing/viewing sessions

## Potential proposed structure, based on intervention progress screen

routes/interventionProgress/
  --/initialAssessment/
  --/actionPlan/
  --/sessionDelivery/
  --/endOfServiceReport/
  -- /serviceProvider/interventionProgress.njk
  -- /probationPractitioner/interventionProgress.njk


Gweaton:
I'm not 100% sure about this proposed restructure, it would be good to discuss as a team

lawrence-forooghian:
I took a look at the changes here, and I think that there's a sensible idea behind it, but I'm having trouble imagining how the sharing might look in practice. Perhaps with an example of something being shared between the PP and SP it would be clearer?

> We can re-use the check your answers page in the PP journey

How would this reuse work? Is there a "submit session feedback" journey as part of the PP service? (I can't remember something like that). Or do you mean that there's some general logic common to all "check your answers" pages that could be shared?

Gweaton:
> How would this reuse work? Is there a "submit session feedback" journey as part of the PP service? (I can't remember something like that). Or do you mean that there's some general logic common to all "check your answers" pages that could be shared?

@lawrence-forooghian At the moment, the check your answers page doesn't have any change links, and displays the answers selected by the PP, so essentially it looks like an "intervention details" style page (although there won't be a "confirm" button). I feel like the only thing that might change in the near future is adding in some change links, which we could possibly do via a conditional, based on an argument or sub-class style method of passing in the user type. I'm not 100% convinced by this approach, but let's chat about it at some point.

Gweaton:
Closing this, as it's not needed right now, and more discussion might be needed in future.

PR #245, opened 2021-04-16
IC-1557 Changes to sent referrals endpoints

tomsmyers:

uses the new and improved API endpoint for getting sent referrals. 

there are a couple of necessary changes here too, the first is updating the user objects to accurately reflect what we are storing in `res.locals.user` - it's kind of annoying we don't get proper type checking on this thing. the second is that we now must validate the SP user is associated with an organization on the UI side. this is a taste of things to come with regards to the bigger auth stories being planned out, but for now is quite simple. i've abstracted the logic to a helper function so it's clear how the organization is being parsed out. 

finally, the contracts are updated to validate the new API expectations.

ground work for PP dashboard

### DIFF
--- a/server/services/interventionsService.ts
+++ b/server/services/interventionsService.ts
@@ -319,11 +319,22 @@ export default class InterventionsService {
     })) as SentReferral
   }
 
-  async getSentReferrals(token: string): Promise<SentReferral[]> {
+  async getReferralsSentByProbationPractitioner(token: string, userId: string): Promise<SentReferral[]> {
     const restClient = this.createRestClient(token)
 
     return (await restClient.get({
       path: `/sent-referrals`,
+      query: { sentBy: userId },
+      headers: { Accept: 'application/json' },
+    })) as SentReferral[]
+  }
+
+  async getReferralsSentToServiceProvider(token: string, providerId: string): Promise<SentReferral[]> {
+    const restClient = this.createRestClient(token)
+
+    return (await restClient.get({
+      path: `/sent-referrals`,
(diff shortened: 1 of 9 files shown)


server/services/interventionsService.ts
```diff
@@ -319,11 +319,22 @@ export default class InterventionsService {
     })) as SentReferral
   }
 
-  async getSentReferrals(token: string): Promise<SentReferral[]> {
+  async getPPSentReferrals(token: string, userId: string): Promise<SentReferral[]> {
```
Gweaton, line 322: I'm normally against using abbreviations in method names/business logic, as I find it impacts readability or might not make sense to newcomers to the codebase, but I'm not sure in this case - I _think_ PP/SP will make sense to everyone and we reference it in a few places e.g. comments, but you never know.

I'd probably vote in favour of making this the full `getProbationPractitionerSentReferrals` (and the same for the SP one), but am not too fussy about this, so happy to let you decide what you think is best.

Gweaton (approved):

server/services/interventionsService.ts
tomsmyers: updated based on convo in dev chat

server/utils/authUtils.ts
```diff
@@ -0,0 +1,20 @@
+import { ServiceProviderOrg, UserDetails } from '../services/userService'
+
+function getSPUserOrganization(userDetails: UserDetails): ServiceProviderOrg {
```
Gweaton, line 3: Would you mind updating this one too now that we've updated the PP one?

PR #247, opened 2021-04-19
IC-I546 skeleton probation practitioner dashboard

tomsmyers:

adds a 'my cases' page for PP's which is the default landing page on login. it shows a list of all referrals that were sent by the logged in PP. 

changes the existing 'probation-practitioner/dashboard' page to '/probation-practitioner/find'

adds a simple (and poorly styled, see screenshot) top nav bar to both the aforementioned pages to allow switching between them. i'm not going to try to fix the way this looks in this PR. a ticket will address this later.

Screenshot 2021-04-21 at 08 19 36

i'm somewhat unhappy with the 'find start' page, both it's name and it's content. i think the draft referrals list should be on the main dashboard, and the 'find interventions' button should take you straight to the search results page. however that is outside the scope of this PR. 

there is also a new auth utils method here that stops SP users from viewing this page (albeit indirectly).

provide an entrypoint for PP's to view cases via the UI.


server/routes/probationPractitionerReferrals/myCasesPresenter.ts
```diff
@@ -0,0 +1,59 @@
...
+    { text: 'Service user', sort: 'ascending' },
+    { text: 'Service Category', sort: 'none' },
+    { text: 'Provider', sort: 'none' },
+    { text: 'Caseworker', sort: 'none' },
+    { text: 'Action', sort: 'none' },
+  ]
+
+  readonly tableRows: SortableTableRow[] = this.sentReferrals.map(referral => {
+    const { serviceCategoryId } = referral.referral
+    const serviceCategory = this.serviceCategories.find(aServiceCategory => aServiceCategory.id === serviceCategoryId)
```
tomsmyers, line 18: this isn't DRY, and i don't like that the caller has to be aware that it needs just the service categories that are present in the list of referrals. would like to find a better way to do this.

server/routes/probationPractitionerReferrals/myCasesPresenter.ts
```diff
@@ -0,0 +1,60 @@
...
+    if (serviceCategory === undefined) {
+      throw new Error(`Expected serviceCategories to contain service category with ID ${serviceCategoryId}`)
+    }
+
+    // i really want the actual names here, but that's an API call for each row - how can we improve this?
```
Gweaton, line 27: Should we be storing the full name on the referral instead / as well? Again, might be a bit of work

server/routes/probationPractitionerReferrals/myCasesPresenter.ts
Gweaton: hmm, I wonder if this calls for a refactor of the referral object to include the service category name as well as the ID now? Not ideal though...

Gweaton (commented):
Nice work - not _quite_ sure of the best approach. Haven't approved yet as we might want to discuss changing the API, unless you or anyone else can think of a simpler approach?

integration_tests/integration/probationPractitionerReferrals.spec.js
```diff
@@ -0,0 +1,73 @@
+import sentReferralFactory from '../../testutils/factories/sentReferral'
```
Gweaton, line 1: There's actually a `monitor.spec` that has PP tests in it, sorry (I thought it was a bit nicer to have smaller chunks based on part of user journey, rather than the whole user journey) - would you mind adding this test to that instead?

Gweaton: I'm wrong, this is part of the refer journey so it does make sense to have stuff in this file...

Gweaton (approved):

PR #248, opened 2021-04-19
Refactor the moj sortable table component into it's own view utils method

tomsmyers:

Refactor the moj sortable table component into it's own view utils method

In an upcoming commit this code gets reused exactly as is. Pulling it out now to make the next review smaller.

### DIFF
--- a/server/routes/serviceProviderReferrals/dashboardPresenter.test.ts
+++ b/server/routes/serviceProviderReferrals/dashboardPresenter.test.ts
@@ -3,21 +3,6 @@ import serviceCategoryFactory from '../../../testutils/factories/serviceCategory
 import sentReferralFactory from '../../../testutils/factories/sentReferral'
 
 describe(DashboardPresenter, () => {
-  describe('tableHeadings', () => {
-    it('returns the table’s headings', () => {
-      const presenter = new DashboardPresenter([], [])
-
-      expect(presenter.tableHeadings).toEqual([
-        'Date received',
-        'Referral',
-        'Service user',
-        'Intervention type',
-        'Caseworker',
-        'Action',
-      ])
-    })
-  })
-
   describe('tableRows', () => {
     it('returns the table’s rows', () => {
       const accommodationServiceCategory = serviceCategoryFactory.build({ name: 'accommodation' })
--- a/server/routes/serviceProviderReferrals/dashboardPresenter.ts
+++ b/server/routes/serviceProviderReferrals/dashboardPresenter.ts
@@ -2,41 +2,47 @@ import { SentReferral, ServiceCategory } from '../../services/interventionsServi
 import CalendarDay from '../../utils/calendarDay'
 import PresenterUtils from '../../utils/presenterUtils'
 import utils from '../../utils/utils'
+import { SortableTableHeaders, SortableTableRow } from '../../utils/viewUtils'
 
 export default class DashboardPresenter {
   constructor(private readonly referrals: SentReferral[], private readonly serviceCategories: ServiceCategory[]) {}
 
-  readonly tableHeadings = ['Date received', 'Referral', 'Service user', 'Intervention type', 'Caseworker', 'Action']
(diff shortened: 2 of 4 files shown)


server/routes/serviceProviderReferrals/dashboardPresenter.test.ts
```diff
@@ -3,21 +3,6 @@ import serviceCategoryFactory from '../../../testutils/factories/serviceCategory
 import sentReferralFactory from '../../../testutils/factories/sentReferral'
 
 describe(DashboardPresenter, () => {
-  describe('tableHeadings', () => {
```
tomsmyers, line 6: this test doesn't really appear to be testing anything so instead of copying the new code from the presenter in here i thought best to just remove

server/routes/serviceProviderReferrals/dashboardPresenter.ts
```diff
@@ -1,42 +1,47 @@
...
+  readonly tableHeadings: SortableTableHeaders = [
+    { text: 'Date received', sort: 'none' },
+    { text: 'Referral', sort: 'none' },
+    { text: 'Service user', sort: 'none' },
+    { text: 'Intervention type', sort: 'none' },
+    { text: 'Caseworker', sort: 'none' },
+    { text: 'Action', sort: 'none' },
+  ]
 
-  readonly tableRows: { text: string; sortValue: string | null; href: string | null }[][] = this.referrals.map(
```
tomsmyers, line 11: the shorter line length here means the `referral => {` is now on the same line and all the indentation below is affected. the code here isn't changed at all apart from the method signature.

server/routes/serviceProviderReferrals/dashboardPresenter.ts
```diff
@@ -1,42 +1,47 @@
...
-import PresenterUtils from '../../utils/presenterUtils'
+import PresenterUtils, { SortableTableHeaders, SortableTableRow } from '../../utils/presenterUtils'
 import utils from '../../utils/utils'
 
 export default class DashboardPresenter {
   constructor(private readonly referrals: SentReferral[], private readonly serviceCategories: ServiceCategory[]) {}
 
-  readonly tableHeadings = ['Date received', 'Referral', 'Service user', 'Intervention type', 'Caseworker', 'Action']
+  readonly tableHeadings: SortableTableHeaders = [
+    { text: 'Date received', sort: 'none' },
```
Gweaton, line 10: What does the new `sort` field do here, and why didn't we have one before?

server/utils/presenterUtils.ts
```diff
@@ -36,6 +36,9 @@ interface TwelveHourTimeInputPresenter {
   partOfDay: PartOfDayInputPresenter
 }
 
+export type SortableTableHeaders = { text: string; sort: 'ascending' | 'descending' | 'none' }[]
```
Gweaton, line 39: Genuine question (not a suggestion) - my understanding of types vs interfaces in Typescript is a little rusty - would making these interfaces have any impact on the code, or are they essentially interchangeable here?

Gweaton: and I wonder if this should go in the `viewUtils` alongside the `sortableTable` function?

Gweaton (commented):
Just a couple of questions, really

server/routes/serviceProviderReferrals/dashboardPresenter.ts
tomsmyers: previously all the headers had 'sort: none' hardcoded in the template. this allows them to be altered by the presenter. it sets the default sort for the table.

server/utils/presenterUtils.ts
tomsmyers: there's very minimal differences now in typescript. see https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces

regarding the layout, happy to move them 👍

Gweaton (approved):
Nice 🥇

PR #251, opened 2021-04-20
IC-1530: Make answers components reusable

Gweaton:

- Creates new session feedback partial that can be used on the "Check your Answers" page and the soon-to-be-added "View Feedback" page.
- Extracts answers that we want to render across both pages into a "FeedbackAnswersPresenter" that can be used on both screens.

To create re-usable components that we can use in the upcoming "View Feedback" page.

### DIFF
--- a/server/routes/serviceProviderReferrals/postSessionFeedbackCheckAnswersPresenter.ts
+++ b/server/routes/serviceProviderReferrals/postSessionFeedbackCheckAnswersPresenter.ts
@@ -1,22 +1,19 @@
 import { DeliusServiceUser } from '../../services/communityApiService'
 import { ActionPlanAppointment } from '../../services/interventionsService'
+import DateUtils from '../../utils/dateUtils'
 import { SummaryListItem } from '../../utils/summaryList'
+import FeedbackAnswersPresenter from '../shared/feedbackAnswersPresenter'
 import ServiceUserBannerPresenter from '../shared/serviceUserBannerPresenter'
-import PostSessionAttendanceFeedbackPresenter from './postSessionAttendanceFeedbackPresenter'
-import PostSessionBehaviourFeedbackPresenter from './postSessionBehaviourFeedbackPresenter'
 
 export default class PostSessionFeedbackCheckAnswersPresenter {
-  private readonly attendancePresenter: PostSessionAttendanceFeedbackPresenter
-
-  private readonly behaviourPresenter: PostSessionBehaviourFeedbackPresenter
+  readonly feedbackAnswersPresenter: FeedbackAnswersPresenter
 
   constructor(
     private readonly appointment: ActionPlanAppointment,
     private readonly serviceUser: DeliusServiceUser,
     private readonly actionPlanId: string
   ) {
-    this.attendancePresenter = new PostSessionAttendanceFeedbackPresenter(appointment, serviceUser)
-    this.behaviourPresenter = new PostSessionBehaviourFeedbackPresenter(appointment, serviceUser)
+    this.feedbackAnswersPresenter = new FeedbackAnswersPresenter(this.appointment, this.serviceUser)
   }
 
   readonly serviceUserBannerPresenter = new ServiceUserBannerPresenter(this.serviceUser)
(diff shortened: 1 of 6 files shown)


server/views/serviceProviderReferrals/postSessionFeedbackCheckAnswers.njk
```diff
@@ -1,34 +1,10 @@
 {% from "govuk/components/button/macro.njk" import govukButton %}
-{% from "govuk/components/radios/macro.njk" import govukRadios %}
 {% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
-{% from "govuk/components/textarea/macro.njk" import govukTextarea %}
 
 {% extends "./actionPlan/actionPlanFormTemplate.njk" %}
 
 {% block formSection %}
-  {{ govukSummaryList(summaryListArgs) }}
```
lawrence-forooghian, line 9: You seem to have lost the summary list here — it's not in the new partial.

server/routes/shared/feedbackAnswersPresenter.test.ts
```diff
@@ -0,0 +1,233 @@
+import actionPlanAppointmentFactory from '../../../testutils/factories/actionPlanAppointment'
```
lawrence-forooghian, line 1: I think that it would be good for this commit to make it clear that this is not new code, but rather is just being moved from `PostSessionFeedbackCheckAnswersPresenter`. Imagine a person in the future who doesn't understand a particular part of this file and so runs a `git blame` to find out more about the commit that introduced it. To them, it'll look like this code is all new, hiding all of the changes that were made to this code in its old location.

I'd say that it'd be best for this commit to, instead of introducing a new class that's a copy and paste of an existing one, also start _using_ the new class and remove the old one - i.e. incorporate the final commit `Use feedbackAnswersPresenter in checkAnswersPresenter` into this one.

Although small, focused commits are a good thing, I think that if we have to prioritise between the commits containing a self-contained change and the commits being very small, then the former probably wins. We review the commits once (where small commits are useful) but we'll probably revisit them without any context multiple times in the future (where self-contained commits are useful).

server/routes/serviceProviderReferrals/postSessionFeedbackCheckAnswersPresenter.ts
```diff
@@ -77,7 +78,16 @@ export default class PostSessionFeedbackCheckAnswersPresenter {
     }
   }
 
-  get sessionDetailsSummary(): SummaryListItem[] {
-    return this.attendancePresenter.sessionDetailsSummary
-  }
+  readonly sessionDetailsSummary: SummaryListItem[] = [
+    {
```
lawrence-forooghian, line 28: We could still have an `attendancePresenter` property that we instantiate inside this class — to be used just for deriving the value of `sessionDetailsSummary` —  _in addition to_ `feedbackAnswersPresenter`. Then we could keep it DRY (if that's the goal).

lawrence-forooghian (commented):
Looks like a good pattern, thanks! As noted in a comment, I think that this PR is really a single refactor and might actually be clearer in just 2 commits (one to extract the partial and another to extract the presenter).

server/routes/shared/feedbackAnswersPresenter.test.ts
Gweaton: That's good feedback, I hadn't really thought about the future value of these commits. I felt it made reviewing easier as it split it into smaller chunks, but you're right - it wouldn't be all that helpful to a future `git blame`r. Have squashed the last 3 commits into one and updated the commit message.

server/routes/serviceProviderReferrals/postSessionFeedbackCheckAnswersPresenter.ts
Gweaton: I've decided to keep it separate and less DRY for now, but explained why in the commit message. Once we know a bit more about the fields we're displaying across the SP and PP journeys, I'm wondering if we want to turn it into a presenter of its own.

server/views/serviceProviderReferrals/postSessionFeedbackCheckAnswers.njk
Gweaton: This was a real pain due to the way this branch was committed and this kept disappearing and reappearing at various points in history during my many rebases. I think I've managed to solve it now though!

lawrence-forooghian (approved):

PR #260, opened 2021-04-23
IC-1321: Allow cancellation comments to be null

Gweaton:

Allows cancellation comments to be `null` when cancelling a referral.

Cancellation comments is an optional field in the UI journey.

### DIFF
--- a/server/services/interventionsService.ts
+++ b/server/services/interventionsService.ts
@@ -624,7 +624,7 @@ export default class InterventionsService {
     token: string,
     referralId: string,
     reasonCode: string,
-    cancellationComments: string
+    cancellationComments: string | null
   ): Promise<EndedReferral> {
     const restClient = this.createRestClient(token)
     return (await restClient.post({


server/services/interventionsService.ts
```diff
@@ -624,7 +624,7 @@ export default class InterventionsService {
     token: string,
     referralId: string,
     reasonCode: string,
-    cancellationComments: string
+    cancellationComments: string | null
```
Gweaton, line 627: Are we happy sending `null` values, or do we want it to be `undefined`?

tomsmyers: sending null values is kind of an antipattern in our API since they are ignored, but it's fine. 

the reason is that you may reasoably expect a null value to update the value to null. i.e. in a PATCH endpoint where something is already set, you might expect sending null to 'unset' it. but we don't support that. and generally discourage sended the value at all.

tomsmyers (approved):

PR #262, opened 2021-04-26
IC-1349 use new 'getReferral' method for populating referral in SP progress page.

tomsmyers:

- the first commit is large, but a really simple refactor of 'ServiceCategory' to 'ServiceCategoryFull'. This is the DTO with all the outcomes and complexity levels in. The reason for this is that the new endpoint returns a 'ServiceCategory' which contains just the name and id.
- the next commit changes a simple shared presenter class to take a referral id instead of a 'SentReferral'. This allows us to use this class with any type of referral object, including the new 'generic' one.
- finally the new service layer method is implemented. There are some simple test changes required, and a new factory class. But most of the changes are removing the need to pass 'service category' to the presenter, and removing the need to create service categories and mock them in the tests. It's neater that way IMO. 

two-fold, firstly general improvements in the API layer, but primarily this will unblock work on determining if a referral has beebn cancelled in the SP progress page.

### DIFF
--- a/package.json
+++ b/package.json
@@ -147,6 +147,7 @@
     "@types/redis": "^2.8.28",
     "@types/superagent": "^4.1.10",
     "@types/supertest": "^2.0.10",
+    "@types/uuid": "^8.3.0",
     "@typescript-eslint/eslint-plugin": "^4.4.1",
     "@typescript-eslint/parser": "^4.4.1",
     "concurrently": "^5.3.0",
@@ -178,6 +179,7 @@
     "supertest": "^4.0.2",
     "ts-jest": "^26.4.1",
     "ts-node": "^9.1.1",
-    "typescript": "^4.0.3"
+    "typescript": "^4.0.3",
+    "uuid": "^3.4.0"
   }
 }
--- a/server/services/interventionsService.ts
+++ b/server/services/interventionsService.ts
@@ -14,7 +14,7 @@ export interface InterventionsServiceValidationError {
 
 type WithNullableValues<T> = { [K in keyof T]: T[K] | null }
 
-export interface DraftReferralFields {
+export interface FormFields {
   completionDeadline: string | null
   complexityLevelId: string | null
   furtherInformation: string | null
@@ -31,7 +31,7 @@ export interface DraftReferralFields {
   maximumRarDays: number | null
 }
 
-export interface SentReferralFields {
+export interface SentFields {
   sentAt: string
   sentBy: AuthUser
   referenceNumber: string
@@ -40,7 +40,7 @@ export interface SentReferralFields {
   endOfServiceReport: EndOfServiceReport | null
 }
 
-export interface EndedReferralFields {
+export interface EndedFields {
   endedAt: string
   endedBy: AuthUser
   cancellationReason: string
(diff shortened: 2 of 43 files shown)


server/services/interventionsService.ts
```diff
@@ -14,6 +14,51 @@ export interface InterventionsServiceValidationError {
...
+export interface Referral {
+  id: string
+  createdAt: string
+  serviceUser: ServiceUser
+  serviceCategory: ServiceCategory
+  serviceProvider: ServiceProvider
+
+  draftReferralFields: DraftReferralFields
+  sentReferralFields: SentReferralFields | null
+  endedReferralField: EndedReferralFields | null
```
Gweaton, line 59: `endedReferralFields`

server/services/interventionsService.ts
```diff
@@ -14,6 +14,51 @@ export interface InterventionsServiceValidationError {
...
+}
+
+export interface Referral {
+  id: string
+  createdAt: string
+  serviceUser: ServiceUser
+  serviceCategory: ServiceCategory
+  serviceProvider: ServiceProvider
+
+  draftReferralFields: DraftReferralFields
```
Gweaton, line 57: `formFields`

server/services/interventionsService.ts
```diff
@@ -14,6 +14,51 @@ export interface InterventionsServiceValidationError {
...
+  sentReferralFields: SentReferralFields | null
```
Gweaton, line 58: sentFields

server/services/interventionsService.ts
tomsmyers: done

server/services/interventionsService.ts
tomsmyers: done

server/services/interventionsService.ts
tomsmyers: done

testutils/factories/referral.ts
```diff
@@ -26,7 +48,7 @@ class ReferralFactory extends Factory<Referral> {
   }
 
   assigned() {
-    return this.params({
+    return this.draft().params({
```
Gweaton, line 51: Do we want this to be `this.sent()` rather than `draft()`? It doesn't make all that much difference in terms of data structure, but you can only assign a referral that has been "sent" so it might be clearer for people looking at the codebase.

package.json
```diff
@@ -178,6 +179,7 @@
     "supertest": "^4.0.2",
     "ts-jest": "^26.4.1",
     "ts-node": "^9.1.1",
-    "typescript": "^4.0.3"
+    "typescript": "^4.0.3",
+    "uuid": "^3.4.0"
```
Gweaton, line 183: Would you mind adding what this is used for to the commit message (and that we've added a package)?

Gweaton (approved):
I think this makes sense - we do need this generic endpoint, and it'd be good to start using it as much as possible. As this is quite a big change, I wonder if it's worth running past @lawrence-forooghian as well, just so we're all on the same page about the new structure before we start designing more features around the old API?

testutils/factories/referral.ts
tomsmyers: the issues with using the `sent()` constructor here is that you have to overwrite _all_ the sent fields because they are nested anyway.

package.json
tomsmyers: sure thing. 

just for reference here too. this was added to generate random UUIDs to make the factory data more realistic looking instead of just using the Fishery sequence for UUIDs.

lawrence-forooghian (approved):

PR #263, opened 2021-04-26
IC-1305: Probation practitioner user can view a submitted end of service report

lawrence-forooghian:

Adds a page for the probation practitioner to view a submitted end of service report.

To allow the probation practitioner to find out how a finished intervention went.

## Screenshot

Screenshot 2021-04-26 at 09 38 27

### DIFF
--- a/server/routes/serviceProviderReferrals/endOfServiceReportCheckAnswersPresenter.ts
+++ b/server/routes/serviceProviderReferrals/endOfServiceReportCheckAnswersPresenter.ts
@@ -1,11 +1,5 @@
-import {
-  EndOfServiceReport,
-  EndOfServiceReportOutcome,
-  SentReferral,
-  ServiceCategory,
-} from '../../services/interventionsService'
-import PresenterUtils from '../../utils/presenterUtils'
-import { SummaryListItem } from '../../utils/summaryList'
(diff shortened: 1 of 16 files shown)


server/routes/serviceProviderReferrals/endOfServiceReportCheckAnswersPresenter.ts
```diff
@@ -32,6 +20,7 @@ export default class EndOfServiceReportCheckAnswersPresenter {
   readonly answersPresenter = new EndOfServiceReportAnswersPresenter(
     this.referral,
     this.endOfServiceReport,
-    this.serviceCategory
+    this.serviceCategory,
+    true
```
Gweaton, line 24: Does this need to be here if we've made `true` the default argument in the presenter? Or do you think it's better for readability here?

Gweaton (approved):
Just one question, looks good otherwise

server/routes/serviceProviderReferrals/endOfServiceReportCheckAnswersPresenter.ts
lawrence-forooghian: Yeah, I added the default argument only to be able to skip it in the tests where it made no difference; I think it's good to be explicit about the value.

PR #264, opened 2021-04-26
IC-1321 & IC-1330: Check answers and submit referral cancellation journey 

Gweaton:

Adds form methods, check your answers and rough confirmation screen for cancelling a referral.

I'm opening this as a draft, as I'm a bit unsure of the best way for doing this, as this is the first example of us not updating a draft object on the backend (as we do with `DraftReferrals`), and trying to do everything in-session. Something doesn't feel quite right, particularly the repeated redirects, and doing a lot of functionality in the Controller, rather than in view/presenter methods, but I'd like to get a second pair of eyes on this.

To allow PPs to cancel a referral for reasons specified by the Interventions Service.


Gweaton:
Closing this in favour of #265, #268 and soon-to-be-opened final PR for confirmation screen.

PR #269, opened 2021-04-27
Probation practitioner can navigate to end of service report

lawrence-forooghian:

This adds a link to the probation practitioner intervention progress
page, so that they can see a submitted end of service report. Until now,
they’ve only been able to view the end of service report by following a
link in a notification email.

The implementation is copied from that of the session progress table in
the same file. I think there are some improvements we could make here
(presenters shouldn’t be emitting HTML) but I want to get this done in
time for show and tell, and I’ll stick with consistency and what works.

To allow a probation practitioner to view an end of service report at any time.

### No end of service report submitted

image

### End of service report submitted

image


tomsmyers (approved):
nice

PR #271, opened 2021-04-28
IC-1321: Add SU banner to referral cancellation reasons

Gweaton:

Adds missing SU banner to referral cancellation page. We'll probably want to do this in a more consistent way to stop us having to do it everywhere, but this can come as a later refactor.

## Screenshot

image

### DIFF
--- a/server/views/probationPractitionerReferrals/referralCancellationReason.njk
+++ b/server/views/probationPractitionerReferrals/referralCancellationReason.njk
@@ -2,6 +2,7 @@
 {% from "govuk/components/radios/macro.njk" import govukRadios %}
 {% from "govuk/components/button/macro.njk" import govukButton %}
 {% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
+{% from "govuk/components/notification-banner/macro.njk" import govukNotificationBanner %}
 
 {% extends "../partials/layout.njk" %}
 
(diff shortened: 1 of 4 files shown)


server/views/probationPractitionerReferrals/referralCancellationReason.njk
```diff
@@ -10,6 +11,10 @@
   - GOV.UK{% endblock %}
 
 {% block content %}
+   {% if serviceUserNotificationBannerArgs !== undefined and serviceUserNotificationBannerArgs !== null %}
```
tomsmyers, line 14: do we have this check elsewhere?

tomsmyers (approved):
nice - agree on the idea of having this show up for more pages by default. 

perhaps a middleware we can easily apply to a view, and then a section in the global template that displays it if it's present.

server/views/probationPractitionerReferrals/referralCancellationReason.njk
Gweaton: I took it from `actionPlanFormTemplate.njk` where the check is probably more useful as it's a partial that we extend in a few places so it might be undefined/null in those places, whereas here it'll be defined. I'm tempted to leave it as-is though as I've put it in all parts of this journey, until we extract it out into a more re-usable component, then we can replace it everywhere.

lawrence-forooghian:
I'm currently adding the banner to a bunch of end of service report pages - I might see what we can do that's reusable.

PR #274, opened 2021-04-28
IC 1345: Add screen where the SP can see that the intervention has been ended

andrewburrowsmoj:

Adds an intervention notification banner onto the service user progress page when the referral has been ended by the PP user.
Attached is the screenshot for how it looks: image

The text was derived from the comment left on ticket https://dsdmoj.atlassian.net/browse/IC-1543

Enabled visibility of intervention status to the SP user.

### DIFF
--- a/server/routes/serviceProviderReferrals/interventionProgressPresenter.test.ts
+++ b/server/routes/serviceProviderReferrals/interventionProgressPresenter.test.ts
@@ -7,6 +7,47 @@ import actionPlanAppointmentFactory from '../../../testutils/factories/actionPla
 import endOfServiceReportFactory from '../../../testutils/factories/endOfServiceReport'
 
 describe(InterventionProgressPresenter, () => {
+  describe('referralEnded', () => {
+    it('returns true when the referral has ended', () => {
+      const referral = sentReferralFactory.endRequested().build()
+      const serviceCategory = serviceCategoryFactory.build()
+      const serviceUser = serviceUserFactory.build()
+      const presenter = new InterventionProgressPresenter(referral, serviceCategory, null, serviceUser, [])
+
+      expect(presenter.referralEnded).toEqual(true)
+    })
+    it('returns false when the referral has not ended', () => {
+      const referral = sentReferralFactory.build()
+      const serviceCategory = serviceCategoryFactory.build()
+      const serviceUser = serviceUserFactory.build()
+      const presenter = new InterventionProgressPresenter(referral, serviceCategory, null, serviceUser, [])
+
+      expect(presenter.referralEnded).toEqual(false)
+    })
+  })
+  describe('referralEndedFields', () => {
+    it('returns ended fields when the referral has ended', () => {
+      const referral = sentReferralFactory.endRequested().build()
+      const serviceCategory = serviceCategoryFactory.build()
+      const serviceUser = serviceUserFactory.build()
+      const presenter = new InterventionProgressPresenter(referral, serviceCategory, null, serviceUser, [])
+
+      expect(presenter.referralEndedFields.endRequestedComments).toEqual(referral.endRequestedComments)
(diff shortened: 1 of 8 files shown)


server/routes/serviceProviderReferrals/interventionProgressPresenter.test.ts
```diff
@@ -7,6 +7,47 @@ import actionPlanAppointmentFactory from '../../../testutils/factories/actionPla
...
 
 describe(InterventionProgressPresenter, () => {
+  describe('referralEnded', () => {
+    it('returns true when the referral has ended', () => {
+      const referral = sentReferralFactory.endRequested().build()
+      const serviceCategory = serviceCategoryFactory.build()
+      const serviceUser = serviceUserFactory.build()
+      const presenter = new InterventionProgressPresenter(referral, serviceCategory, null, serviceUser, [])
+
+      expect(presenter.referralAssigned).toEqual(true)
```
tomsmyers, line 17: should this be checking referralAssigned or referralEnded?

server/routes/serviceProviderReferrals/interventionProgressView.ts
```diff
@@ -1,11 +1,46 @@
...
 
 import ViewUtils from '../../utils/viewUtils'
 import InterventionProgressPresenter from './interventionProgressPresenter'
+import DateUtils from '../../utils/dateUtils'
 
 export default class InterventionProgressView {
   constructor(private readonly presenter: InterventionProgressPresenter) {}
 
+  private cancelledReferralNotificationBannerArgs(): NotificationBannerArgs {
+    let html = ''
```
tomsmyers, line 17: const vars are preferred everywhere if possible.

server/routes/serviceProviderReferrals/interventionProgressView.ts
```diff
@@ -1,11 +1,46 @@
...
+} from '../../utils/govukFrontendTypes'
 
 import ViewUtils from '../../utils/viewUtils'
 import InterventionProgressPresenter from './interventionProgressPresenter'
+import DateUtils from '../../utils/dateUtils'
 
 export default class InterventionProgressView {
   constructor(private readonly presenter: InterventionProgressPresenter) {}
 
+  private cancelledReferralNotificationBannerArgs(): NotificationBannerArgs {
```
tomsmyers, line 16: nitpick: this could be a getter or readonly field instead of a method

tomsmyers (approved):
great work! just one possible typo in a unit test top fix.

testutils/factories/sentReferral.ts
```diff
@@ -67,7 +67,7 @@ class SentReferralFactory extends Factory<SentReferral> {
 
   endRequested() {
     return this.assigned().params({
-      endRequestedAt: '2021-4-28T20:45:21.986389Z',
+      endRequestedAt: '2021-04-28T20:45:21.986389Z',
```
Gweaton, line 70: I've fixed this in its own commit on my branch to explain why the change was made - are you ok for me to merge that one first (or feel free to cherry-pick it on to this branch if you'd like to get it in first)

server/routes/serviceProviderReferrals/interventionProgressView.ts
andrewburrowsmoj: In this case the fields values are nullable, so therefore I need to default the cancellationReasonHTML and cancellationCommentsHTML to an empty value. This is why I've used `let` here instead of `const`. 

I suppose that if this method is ever called then the endRequestedAt and endRequestedReason should never be null and I could thrown an exception instead.

andrewburrowsmoj: Oh, unless you're just referring to the `html` var, in which case I'll change that to `const`.

tomsmyers (approved):

PR #278, opened 2021-04-29
IC-1639: Display message to PP on intervention progress page when they have requested to end a referral

Gweaton:

Displays a message to PP on intervention progress page when they have requested to end a referral to show requested end date, rather than a link to cancel the referral again.

To prevent them from being able to cancel a referral, and know that they have requested an end to the service.

## Screenshot

image

### DIFF
--- a/integration_tests/integration/monitor.spec.js
+++ b/integration_tests/integration/monitor.spec.js
@@ -257,6 +257,19 @@ describe('Probation Practitioner monitor journey', () => {
 
       cy.contains('Cancel referral').click()
       cy.contains('This referral has been cancelled')
+
+      const referralWithEndRequested = sentReferralFactory.endRequested().build({
+        ...referralParams,
+        assignedTo: { username: probationPractitioner.username },
+        actionPlanId: actionPlan.id,
+        endRequestedAt: '2021-04-04T20:45:21.986389Z',
+      })
+
+      cy.stubGetSentReferral(referralWithEndRequested.id, referralWithEndRequested)
+
+      cy.visit(`/probation-practitioner/referrals/${referralWithEndRequested.id}/progress`)
+
+      cy.contains('You requested to end this service on 04 Apr 2021')
     })
   })
 })
--- a/server/routes/probationPractitionerReferrals/interventionProgressPresenter.ts
+++ b/server/routes/probationPractitionerReferrals/interventionProgressPresenter.ts
@@ -38,6 +38,20 @@ export default class InterventionProgressPresenter {
     return this.referral.assignedTo !== null
   }
 
+  get referralEndRequested(): boolean {
+    return this.referral.endRequestedAt !== null
+  }
+
+  get referralEndRequestedText(): string {
+    const shortenedDateString = DateUtils.getDateStringFromDateTimeString(this.referral.endRequestedAt)
+
+    if (!shortenedDateString) {
+      return ''
+    }
+
+    return `You requested to end this service on ${shortenedDateString}.`
+  }
+
   readonly text = {
     title: `${utils.convertToTitleCase(this.serviceCategory.name)} progress`,
   }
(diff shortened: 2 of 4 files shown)


tomsmyers:
🚨 new terminology alert !

i think we should quickly clarify 'terminated' with the wider team before this gets merged

integration_tests/integration/monitor.spec.js
```diff
@@ -257,6 +257,19 @@ describe('Probation Practitioner monitor journey', () => {
 
       cy.contains('Cancel referral').click()
       cy.contains('This referral has been cancelled')
+
+      const referralWithEndRequested = sentReferralFactory.endRequested().build({
+        ...referralParams,
+        assignedTo: { username: probationPractitioner.username },
+        actionPlanId: actionPlan.id,
+        endRequestedAt: '2021-04-04T20:45:21.986389Z',
```
andrewburrowsmoj, line 265: The endRequestedAt var is already defined in the factory. (I'm guessing it's to try and fix a broken intervention test)

server/routes/probationPractitionerReferrals/interventionProgressPresenter.ts
```diff
@@ -39,6 +39,20 @@ export default class InterventionProgressPresenter {
     return this.referral.assignedTo !== null
   }
 
+  get referralEndRequested(): boolean {
+    return this.referral.endRequestedAt !== null
+  }
+
+  get referralEndRequestedText(): string {
```
andrewburrowsmoj, line 46: So I wasn't sure if the logic for constructing the string should be in the Presenter or the View class. 

My understanding of the roles are:
- Presenter: derive fields necessary for constructing a page
- View: format fields into the form that the page expects that fields to be

So in my PR I just created a new method which passes the endedFields and the View was responsible for formatting that into a form that is required for the cancellation banner.

I suppose the benefit of having it in the Presenter class is that it is testable since there are no unit tests for the View classes.

integration_tests/integration/monitor.spec.js
Gweaton: Yeah - I put this in (even though it's essentially a duplicate value to the factory) as I feel it's a bit clearer to have the test setup in the same place as the expectation for the date being displayed below, so you don't have to click into another file (the factory one) to see what date value we're testing here - does that feel like duplication / a bit over the top to you?

andrewburrowsmoj: Sorry, I must sound like I'm nitpicking; I assumed it was something you missed as I did something similar to work out why it wasn't working due to the date typo.

I slightly prefer having it within the factory class because there is a single point of reference to reduce work on any change to model structure; but it makes sense this way as well.

server/routes/probationPractitionerReferrals/interventionProgressPresenter.ts
Gweaton: We've been a bit inconsistent in the codebase. The way I see it is - presenter does most of the heavy lifting/logic for wrangling data fields into text if we need to interpolate them somewhere, and also contains some text to be passed to the view.

I guess we have a similar understanding, but possibly different interpretations of formatting fields - I guess I'd see formatting fields with HTML as a View concern, but for shorter strings without any formatting I'd opt for the presenter.

I'm not 100% on this, it'd be good to figure out a consistent way forward!

andrewburrowsmoj: I see. I think so long as we're consistent is all that matters. 
Given I'm completely new I'll make sure to follow your conventions - I'll make the changes to my last PR to ensure that the formatting is done in the presenter class.

andrewburrowsmoj (approved):

tomsmyers (changes requested):
based on conversation with the designers we should avoid the word 'terminated'. let's just use 'ended' or 'requested to end' for now.

tomsmyers (approved):

PR #279, opened 2021-04-29
Add service user banner to end of service report pages

lawrence-forooghian:

1. Refactors the service user banner presenter into a separate view and presenter, so that the presenter sticks to its responsibilities and doesn't generate HTML.
2. Adds a mechanism for passing a view and presenter to each page’s layout ("layout" in a Nunjucks sense, i.e. the base `layout.njk` template that provides the structure for each of our pages).
3. Uses this mechanism to replace the existing ad-hoc code for putting a service user banner on a page, which is currently handed on a page-by-page basis.
4. Adds service user banners to the end of service report pages.

## What _doesn't_ this pull request do?

It doesn't try to do anything about how the data for the service user banner is _fetched_. Each controller remains responsible for fetching the correct `DeliusServiceUser` instance from Community API. I think I've heard people mentioning possibly creating some sort of middleware that's responsible for this fetching, but:

- I don't think that the repeated fetching is much of a problem, compared to that of repeatedly setting up views, templates and presenters to format and display the banner
- A middleware would need quite a lot of context about the controller action it was being called for, in order to be able to find the correct referral ID and hence the service user's CRN. Bear in mind that some of our pages don't use a referral ID in their URL; they use the ID of some other resource which is linked to a referral, requiring a fetch of that resource first.
- If the middleware were meant to fetch data that needed to then be passed to a controller action, how would we do this in a way that's type-safe? Middlewares usually operate by sticking data into the request or response, which are both very untyped objects.

To make it easy to add the service banner to future pages, and to make it easy to add other elements which are common to all pages, but which have some kind of presentation logic, in the future.

## What should we do next?

1.    Go through all controller actions in the app where we're passing a null service user – meaning that it's not displaying a service user banner — and make it display the banner if appropriate.
 1.   Move other presentation logic that's currently being handled in the layout.njk template or a service class into LayoutPresenter (e.g. the header that shows the name and organisation of the logged-in user).


server/routes/findInterventions/findInterventionsController.ts
```diff
@@ -20,7 +21,7 @@ export default class FindInterventionsController {
     const presenter = new SearchResultsPresenter(interventions, filter, pccRegions)
     const view = new SearchResultsView(presenter)
 
-    res.render(...view.renderArgs)
+    ControllerUtils.renderWithLayout(res, view)
```
Gweaton: So just to clarify, we'll be using this everywhere we render a view (rather than `res.render` as that's wrapped in this function?)

server/routes/shared/layoutPresenter.ts
```diff
@@ -1 +1,8 @@
-export default class LayoutPresenter {}
+import { DeliusServiceUser } from '../../services/communityApiService'
+import ServiceUserBannerPresenter from './serviceUserBannerPresenter'
+
+export default class LayoutPresenter {
+  constructor(private readonly serviceUser: DeliusServiceUser | null) {}
```
Gweaton, line 5: Have you considered making this a default argument, so we don't have to add in all the `null`s everywhere?

If so, this could probably come as a refactor when we start adding the service user banner in more places, as it'll be easier to replace the `null`s.

server/routes/serviceProviderReferrals/serviceProviderReferralsController.test.ts
```diff
@@ -89,7 +89,7 @@ describe('GET /service-provider/referrals/:id/details', () => {
       surname: 'Beaks',
       email: 'bernard.beaks@justice.gov.uk',
     })
-    const serviceUser = deliusServiceUserFactory.build({
+    const deliusServiceUser = deliusServiceUserFactory.build({
```
Gweaton, line 92: Good idea, this has confused me quite a bit recently.

Gweaton (approved):
This is a mammoth task, well done. I had hoped could use some kind of middleware for this rather than having to update this everywhere, so thanks for explaining why that wasn't possible/the best way of doing things. Just had a couple of questions, really.

server/routes/findInterventions/findInterventionsController.ts
lawrence-forooghian: Yep, that’s right - at least, if you want to render a view that uses our `layout.njk` layout.

server/routes/shared/layoutPresenter.ts
lawrence-forooghian: Yeah, I did consider it and then didn’t, for some reason. It’d probably be a good idea, but as you say I think it can wait.

PR #280, opened 2021-04-29
Refactor: Break Session Table HTML down into structured data

Gweaton:

Refactors SP intervention progress presenter to no longer pass HTML to the View layer, instead constructing the HTML inside the view layer from structured data passed by the presenter.

Passing structured data around the application rather than long HTML strings feels safer and less brittle to test, as we're not testing whole strings and asserting that the HTML attributes are in a specific order. It should be the View's responsibility to add HTML to the data passed to it.

### DIFF
--- a/server/routes/serviceProviderReferrals/interventionProgressView.ts
+++ b/server/routes/serviceProviderReferrals/interventionProgressView.ts
@@ -110,10 +110,16 @@ export default class InterventionProgressView {
           { text: `Session ${row.sessionNumber}` },
           { text: `${row.appointmentTime}` },
           { text: tagMacro(row.tagArgs as TagArgs) },
-          { html: `${row.linkHtml}` },
+          { html: this.linkHtml(row.links) },
         ]
       }),
     }
+  }
+
+  private linkHtml(links: { text: string; href: string }[]): string {
+    return links
+      .map(link => `<a class="govuk-link" href="${ViewUtils.escape(link.href)}">${ViewUtils.escape(link.text)}</a>`)
+      .join('<br>')
   }
 
   private readonly backLinkArgs = {
--- a/server/routes/probationPractitionerReferrals/interventionProgressPresenter.test.ts
+++ b/server/routes/probationPractitionerReferrals/interventionProgressPresenter.test.ts
@@ -34,8 +34,8 @@ describe(InterventionProgressPresenter, () => {
               classes: 'govuk-tag--grey',
             },
             link: {
-              text: '',
-              href: '',
+              text: null,
+              href: null,
             },
           },
         ])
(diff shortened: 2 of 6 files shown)


server/routes/serviceProviderReferrals/interventionProgressView.ts
```diff
@@ -110,12 +110,24 @@ export default class InterventionProgressView {
...
+          { html: this.linkHtml(row.links) },
         ]
       }),
     }
   }
 
+  private linkHtml(links: { text: string; href: string }[]): string {
+    const linkArray = links.map(link => {
+      if (link.text && link.href) {
+        return `<a class="govuk-link" href="${link.href}">${link.text}</a>`
```
lawrence-forooghian, line 122: These both need a `ViewUtils.escape`.

lawrence-forooghian (changes requested):
Love it, a good refactor.

server/routes/serviceProviderReferrals/interventionProgressView.ts
```diff
@@ -110,12 +110,24 @@ export default class InterventionProgressView {
...
-          { html: `${row.linkHtml}` },
+          { html: this.linkHtml(row.links) },
         ]
       }),
     }
   }
 
+  private linkHtml(links: { text: string; href: string }[]): string {
+    const linkArray = links.map(link => {
+      if (link.text && link.href) {
```
lawrence-forooghian, line 121: When would this condition be false?

Gweaton: When link/href is an empty string (which we set it to if we don't want to display a link)

server/routes/serviceProviderReferrals/interventionProgressView.ts
Gweaton: Even if there's no user input? These are coming from the urls / text fields we've hardcoded in the presenter.

server/routes/serviceProviderReferrals/interventionProgressView.ts
lawrence-forooghian: Hmm, could we use `null` to indicate this instead? It's clearer, I think.

lawrence-forooghian: (and it's what we do elsewhere)

server/routes/serviceProviderReferrals/interventionProgressView.ts
lawrence-forooghian: Yeah, what if there are non HTML-safe characters in the text, for example? They still need to be escaped.

Gweaton: Ok, have updated this

server/routes/serviceProviderReferrals/interventionProgressView.ts
Gweaton: Sorry, I got this confused with the PP code path (where we do set them to empty strings), they're always set here. I've removed the check here and added a separate commit for making the PP ones nullable.

lawrence-forooghian (approved):

PR #282, opened 2021-04-30
Refactor logged-in user header to use presenter

lawrence-forooghian:

Removes presentation logic from the view and service classes, and moves it to a presenter.

Architectural consistency.


Gweaton (approved):
Nice refactor

PR #285, opened 2021-05-04
IC-1601: Make Region Find filters scrollable

Gweaton:

Makes Region filters scrollable in Find journey.

To stop users having to scroll to bottom of page to select a region.

## Screenshot

image

### DIFF
--- a/assets/sass/application.sass
+++ b/assets/sass/application.sass
@@ -6,5 +6,6 @@ $path: "/assets/images/"
 
 @import './components/header-bar'
 @import './components/notification-banner'
+@import './components/find-filters'
 
 @import './local'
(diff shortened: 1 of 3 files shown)


andrewburrowsmoj:
Could you pass the css class into the checkbox `classes` nunjuck macro rather than setting it on the parent?

This would prevent the need to change the css if the order changes.

Gweaton:
> Could you pass the css class into the checkbox `classes` nunjuck macro rather than setting it on the parent?
> (as per docs: https://design-system.service.gov.uk/components/checkboxes/ )
> 
> This would prevent the need to change the css if the order changes.
> 
> Or alternatively, if that's not an option, wrap each checkbox element and set the css class on there.

I did give that a go, but really struggled to get the lists with 1 or 2 elements in them to not scroll. It looks like it was caused by the `input` element being too tall - I've now set that to be `100%` which seems to have fixed the issue.

andrewburrowsmoj (approved):
Nice solution.

PR #287, opened 2021-05-05
IC-1535: Fix links on 'Make a referral' screen

andrewburrowsmoj:

Only makes links active on the "Make a referral" form page based on current progress of form completion.
Links are disabled until the user has completed all the details from the previous page.
Optional string values are considered incomplete when they are null, if the user does not enter a value then they are passed an empty string. There are no other optional types.

image

To force users to complete the form in sequence

### DIFF
--- a/server/routes/referrals/referralFormPresenter.ts
+++ b/server/routes/referrals/referralFormPresenter.ts
@@ -1,159 +1,192 @@
+/* eslint max-classes-per-file: 0 */
 import DraftReferral from '../../models/draftReferral'
 
 export default class ReferralFormPresenter {
-  constructor(private readonly referral: DraftReferral, private readonly serviceCategoryName: string) {}
+  private readonly taskValues: typeof ReferralFormPresenter.TaskValues.prototype
 
-  get sections(): ReferralFormSectionPresenter[] {
-    const serviceUserSectionFormStatus = this.referralFormStatus(this.requiredServiceUserFormValues)
-    const interventionDetailsSectionFormStatus = this.referralFormStatus(
-      this.requiredInterventionFormValues,
-      serviceUserSectionFormStatus
-    )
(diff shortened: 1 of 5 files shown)


server/routes/referrals/referralFormPresenter.ts
```diff
@@ -4,120 +4,125 @@ export default class ReferralFormPresenter {
...
+      this.rarDaysTaskValues &&
+      this.furtherInformationTaskValues
+    )
+  }
+
+  private get checkYourAnswersSectionValues(): DraftReferralValues {
+    return this.checkAnswersTaskValues
+  }
+
+  private get serviceUserDetailsTaskValues(): DraftReferralValues {
```
Gweaton, line 125: I was testing this out locally with CRN `X320741` (one that has a sentence available to select from the local community API) - I couldn't figure out why the second link wouldn't become visible, and it's because the test data doesn't have a value for all of these fields - see screenshot:

image

I don't think this is a very realistic example of data we'll get from the production APIs, but it suggests that those values _could_ be `null`, in which case the SU details tasks will never be completed. I wonder if it's worth just putting a check on CRN, FirstName, date of birth? It's hard to know which fields will always be present in the response though...

I suppose if we were to pick up this ticket: https://dsdmoj.atlassian.net/browse/IC-1245, we could set the values to `None` or `N/A` or something like that, but I'm not sure the priority of that ticket. What do you think?

Gweaton: Maybe, like you demoed this morning, we could just always make that link available as a first step?

Gweaton:
I've just tried filling in the form, and it looks like the `Check your answers` section is marked as `completed` before I click into it - is that a change introduced in this PR (looking at the `calculateStatus` section) - can you reproduce that on your side?

Gweaton (approved):
Thanks for the changes!

PR #288, opened 2021-05-05
IC-1603: Add the contracts needed to build the "select service category" page for cohort interventions

lawrence-forooghian:

Adds the API contracts needed to build the "select service category" page for cohort interventions. Specifically, it:

- allows an intervention to present a list of service categories, which the probation practitioner will pick from as part of the referral to a cohort intervention
- provides the information needed to be able to fetch an intervention given a referral
- allows a referral to have a list of service categories

To allow us to start building the referral journey for cohort interventions.

### DIFF
--- a/server/models/intervention.ts
+++ b/server/models/intervention.ts
@@ -17,6 +17,7 @@ export default interface Intervention {
   npsRegion: NPSRegion | null
   pccRegions: PCCRegion[]
   serviceCategory: ServiceCategory
+  serviceCategories: ServiceCategory[]
   serviceProvider: ServiceProvider
   eligibility: Eligibility
 }
--- a/server/services/interventionsService.test.ts
+++ b/server/services/interventionsService.test.ts
@@ -995,6 +995,7 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
         name: 'Harmony Living',
       },
       serviceCategoryId: '428ee70f-3001-4399-95a6-ad25eaaede16',
+      serviceCategoryIds: ['428ee70f-3001-4399-95a6-ad25eaaede16'],
       complexityLevelId: 'd0db50b0-4a50-4fc7-a006-9c97530e38b2',
       furtherInformation: 'Some information about the service user',
       relevantSentenceId: 2600295124,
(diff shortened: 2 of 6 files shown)


Gweaton (approved):

server/services/interventionsService.test.ts
```diff
@@ -799,6 +801,44 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
...
+          },
+          body: {
+            serviceCategoryIds: ['c81b6da3-77ef-4e9d-b8c7-c703b2cc7e8f', 'eaed6f70-f8cb-40dd-a0ca-8d91c5906d12'],
+          },
+        },
+        willRespondWith: {
+          status: 200,
+          body: {
+            id: 'd496e4a7-7cc1-44ea-ba67-c295084f1962',
+            serviceCategoryIds: ['c81b6da3-77ef-4e9d-b8c7-c703b2cc7e8f', 'eaed6f70-f8cb-40dd-a0ca-8d91c5906d12'],
```
sldblog, line 825: Should we scope these as "selected" or "chosen" categories? Or is that implied?

server/models/intervention.ts
```diff
@@ -17,6 +17,7 @@ export default interface Intervention {
   npsRegion: NPSRegion | null
   pccRegions: PCCRegion[]
   serviceCategory: ServiceCategory
```
sldblog, line 19: Can we also remove the singular category, please? It is superseded with the list

sldblog: To expand on this, we can't set the singular category as it was inferred from the contract.

Now we need a subset of categories allowed by the contract...

Gweaton: I believe Lawrence kept the singular category to make this change backwards compatible, so we don't have to update it all in one go, as we'll be writing code to switching over to the new behaviour and access multiple service categories in future stories, during which we can phase out the superseded behaviour.

server/services/interventionsService.test.ts
Gweaton: We've sort of implied this in the past, e.g. we have `desiredOutcomesIds` & `complexityLevelId` (which are those selected by the user) - I guess once it's no longer a "draft" referral, the "choice" element feels slightly less important, what do you think?

server/models/intervention.ts
lawrence-forooghian: I'm not working today, but saw this and thought I should respond to avoid confusion – yes, what George said. I think that if you take a look at my commit messages, I explained this, but sorry if it's still not clear!

PR #289, opened 2021-05-10
IC-1603 Add page for selecting service categories for a Cohort Referral

Gweaton:

Adds screen for selecting service types on a Cohort referral. This curently isn't linked to by the referral form (that's coming in IC-1686) and isn't linked up to any other parts of the journey. I've added mocks to get this to work locally, as the backend hasn't been built for this feature.

Cohort referrals encompass multiple service categories, so the PP needs to select the relevant ones for the service user, after which they'll select desired outcomes and complexity level for these.

## Screenshots

image

### DIFF
--- a/server/utils/errorMessages.ts
+++ b/server/utils/errorMessages.ts
@@ -15,6 +15,9 @@ export default {
     invalidDate: 'The date by which the service needs to be completed must be a real date',
     mustBeInFuture: 'The date by which the service needs to be completed must be in the future',
   },
+  serviceCategories: {
+    empty: 'At least one service type must be selected',
+  },
   complexityLevel: {
     empty: 'Select a complexity level',
   },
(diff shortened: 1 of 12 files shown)


server/utils/errorMessages.ts
```diff
@@ -15,6 +15,9 @@ export default {
     invalidDate: 'The date by which the service needs to be completed must be a real date',
     mustBeInFuture: 'The date by which the service needs to be completed must be in the future',
   },
+  serviceCategories: {
+    empty: 'Select service types',
```
tomsmyers, line 19: i think this error could be more descriptive. e.g. "at least one service category must be selected"

tomsmyers (approved):
just the one change requested on the the error text, but otherwise nice work!

tomsmyers:
also a question on the header text "service types", is this the agreed upon terminology? why is it different to "service category"?

Gweaton:
> also a question on the header text "service types", is this the agreed upon terminology? why is it different to "service category"?

Good question, I have seen both terms used recently, although we only currently refer to `category` in the codebase. It might be a new development but will ask in the demo time tomorrow.

PR #290, opened 2021-05-10
IC-1604: Add contract tests for selecting desired outcomes on cohort referrals

Gweaton:

Adds `cohortDesiredOutcomes` field to draft referrals.

This is draft, for now, as I'd like to get feedback on the naming and structure.

This will store the user-selected desired outcomes on a cohort referral
against the relevant service category.

I'm not 100% sure about the naming of this, as it may be the case that
we want this to be a total replacement for how we're currently storing
`desiredOutcomesIds` for single-service referrals, in which case, we can
probably drop the `cohort` but I thought I'd leave it on there for now
to make it backwards compatible, but we may want to just call it
something like `desiredOutcomes` in future.


server/models/cohortDesiredOutcomes.ts
```diff
@@ -0,0 +1,4 @@
+export default interface CohortDesiredOutcomes {
+  serviceCategoryId: string
+  desiredOutcomesIds: string[]
```
Gweaton, line 3: Again, are Ids enough here, or do we want the descriptions?

server/models/draftReferral.ts
```diff
@@ -14,6 +15,7 @@ export interface ReferralFields {
   furtherInformation: string
   relevantSentenceId: number
   desiredOutcomesIds: string[]
+  cohortDesiredOutcomes: CohortDesiredOutcomes[]
```
Gweaton, line 18: If this will eventually supersede `desiredOutcomesIds`, should we drop the `cohort` from this?

Gweaton:
Closing this, as we don't believe we'll actually need to store desired outcomes against a service category id as they are all uniquely identifiable and belong to a service category, so storing them as an array (as we currently do) should be fine.

PR #292, opened 2021-05-11
Fix: Unfocus referrals controller test

Gweaton:

Removes the `f` from a focused test to ensure all tests are running.

It looks like this was added in error in ba42422 and although since
then I've seen 'x skipped tests' in the output, I seem to not have put 2
and 2 together.


tomsmyers (approved):
i didn't actually know about this. seems like a slightly weird feature!

PR #293, opened 2021-05-11
IC-1604: Add contracts for new desired outcomes structure to support cohort referrals

Gweaton:

Adds support for selecting multiple desired outcomes on cohort referrals.

This adds a contract that assuming we have the following endpoint on the backend that accepts a PATCH request: 
`/draft-referral/${referralId}/desired-outcomes`.

I've opted for a PATCH, rather than a PUT here, as PATCH should be used for a partial update to a resource, while PUT should be used when replacing that resource. Although we are effectively replacing the `desiredOutcomes` resource, that resource is actually on the `Referral`, which is the entity we're updating.

This PR doesn't use the new method yet.

To allow PPs to select desired outcomes for a specific Service Category.

We initially thought we would be able to avoid adding a new endpoint and data structure, as we could identify which service category a desired outcome belongs to by fetching the Service Category and looking through its Desired Outcomes Ids. This has become a lot more complicated when trying to set the desired outcomes across multiple pages, as we don't want to overwrite desired outcomes that have been set in a previous page/step. If a user comes back to the previous page and unselects a desired outcome, we need to make sure we update that object on the backend as well.

### DIFF
--- a/server/services/interventionsService.ts
+++ b/server/services/interventionsService.ts
@@ -14,6 +14,7 @@ import ServiceCategory from '../models/serviceCategory'
 import ActionPlan, { ActionPlanAppointment, AppointmentAttendance, AppointmentBehaviour } from '../models/actionPlan'
 import DraftReferral from '../models/draftReferral'
 import SentReferral from '../models/sentReferral'
+import ReferralDesiredOutcomes from '../models/referralDesiredOutcomes'
 
 export type InterventionsServiceError = SanitisedError & { validationErrors?: InterventionsServiceValidationError[] }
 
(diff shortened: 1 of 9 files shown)


Gweaton:
Converted to draft after discussion - will do some renaming and we'll use a new endpoint

server/services/interventionsService.ts
```diff
@@ -153,6 +154,24 @@ export default class InterventionsService {
...
+  async setDesiredOutcomesForServiceCategory(
+    token: string,
+    referralId: string,
+    desiredOutcomes: Partial<ReferralDesiredOutcomes>
+  ): Promise<DraftReferral> {
+    const restClient = this.createRestClient(token)
+
+    try {
+      return (await restClient.patch({
+        path: `/draft-referral/${referralId}/desired-outcomes`,
```
Gweaton, line 166: @tomsmyers as discussed, have updated this

tomsmyers (approved):
nice, thanks!

PR #297, opened 2021-05-12
IC-1604: Add page for selecting desired outcomes on cohort referrals

Gweaton:

Adds a page for selecting desired outcomes on cohort referrals.

This isn't currently linked to in the referral form, as there's going to be a bit of logic that will be covered in IC-1598 9 (once the complexity level page has also been added).

To allow PPs to select desired outcomes for specific service categories on cohort referrals.

## Screenshot (mocked data - hence the wrong desired outcome text)

image

### DIFF
--- a/server/routes/referrals/desiredOutcomesForm.test.ts
+++ b/server/routes/referrals/desiredOutcomesForm.test.ts
@@ -1,89 +1,36 @@
-import { Request } from 'express'
+import TestUtils from '../../../testutils/testUtils'
 import DesiredOutcomesForm from './desiredOutcomesForm'
 
 describe(DesiredOutcomesForm, () => {
-  describe('isValid', () => {
-    it('returns true when the desired-outcomes-ids property is present and not empty in the body', async () => {
-      const form = await DesiredOutcomesForm.createForm({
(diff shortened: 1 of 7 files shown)


server/routes/referrals/desiredOutcomesForm.test.ts
```diff
@@ -1,89 +1,36 @@
...
 import DesiredOutcomesForm from './desiredOutcomesForm'
 
 describe(DesiredOutcomesForm, () => {
-  describe('isValid', () => {
-    it('returns true when the desired-outcomes-ids property is present and not empty in the body', async () => {
-      const form = await DesiredOutcomesForm.createForm({
-        body: {
+  describe('data', () => {
+    describe('when desired outcomes ids are passed', () => {
+      it('returns params for update when the service-category-ids property is present and not empty in the body', async () => {
```
lawrence-forooghian, line 7: I think that we can lose the first part of these `it` statements, which have been moved into the `describe`. So e.g. this one would just be `it('returns params for update')` etc.

lawrence-forooghian (approved):

PR #300, opened 2021-05-13
IC-1605: Add contracts and method for setting complexity level per service category on cohort referrals

Gweaton:

Adds contracts and method for setting complexity level per service category on cohort referrals.

This change allows us to associate a complexity level with a given service category, like we do with Desired Outcomes, as now that cohort referrals have been brought into the mix, there's the possibility of a referral having multiple service categories selected on it.

### DIFF
--- a/server/services/interventionsService.test.ts
+++ b/server/services/interventionsService.test.ts
@@ -1144,10 +1144,6 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
           serviceCategoryId: '428ee70f-3001-4399-95a6-ad25eaaede16',
           desiredOutcomesIds: ['301ead30-30a4-4c7c-8296-2768abfb59b5', '65924ac6-9724-455b-ad30-906936291421'],
         },
-        {
-          serviceCategoryId: '3c7d6bc9-540a-4aef-a3fe-dfdff7a3c124',
-          desiredOutcomesIds: ['263821e1-3ad1-44ee-8ec5-c1a925f7a828', '0694fcc9-833f-4756-8d93-28199c0ec58a'],
-        },
       ],
       additionalNeedsInformation: 'Alex is currently sleeping on her aunt’s sofa',
       accessibilityNeeds: 'She uses a wheelchair',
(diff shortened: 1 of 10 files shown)


andrewburrowsmoj (approved):
Looks good - assuming that the TODO for patching ComplexityLevel will eventually refer to ComplexityLevel object rather than DraftReferral.

server/services/interventionsService.test.ts
```diff
@@ -939,6 +1009,56 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
...
   })
 
+  describe('setComplexityLevelForServiceCategory', () => {
+    it('returns the updated referral when selecting a complexity level on a cohort referral', async () => {
+      await provider.addInteraction({
+        state: `There is an existing draft cohort referral with ID of 06716f8e-f507-42d4-bdcc-44c90e18dbd7, and it has had multiple service categories selected`,
+        uponReceiving: 'a PATCH request to set the complexity level for a service category on a referral',
+        withRequest: {
+          method: 'PATCH',
+          path: `/draft-referral/06716f8e-f507-42d4-bdcc-44c90e18dbd7/complexity-levels`,
```
tomsmyers, line 1019: should the trailing `s` be here on this endpoint?

tomsmyers: you can only set one at a time right?

Gweaton: I went back and forth on this, mostly in terms of RESTful naming I thought plurals was the way to go, I know we had a bunch of conversations about this back in the day! - I don't think we've been all that consistent, although we do mostly refer to a single `draft-referral` here, so I'm happy to switch to singular here if you think it's more logical?

tomsmyers: i don't really mind. the only reason i commented is because for the outcomes one you are passing 1 or more outcome per service cat. for this one you are just passing one everytime.

Gweaton: You're right - I'll switch it, no biggie!

PR #301, opened 2021-05-14
IC-1605: Add page for selecting complexity levels on Cohort referrals

Gweaton:

Adds a page for selecting complexity levels on cohort referrals.

This isn't currently linked to in the referral form, but that will come next.

We'll also use this to replace the existing page for selecting a complexity level on single-service referrals.

To allow PPs to select complexity levels on Cohort Referrals

## Screenshot

image

### DIFF
--- a/server/routes/index.ts
+++ b/server/routes/index.ts
@@ -211,6 +211,9 @@ export default function routes(router: Router, services: Services): Router {
   post('/referrals/:id/service-categories', (req, res) => referralsController.updateServiceCategories(req, res))
   get('/referrals/:id/complexity-level', (req, res) => referralsController.viewComplexityLevel(req, res))
   post('/referrals/:id/complexity-level', (req, res) => referralsController.updateComplexityLevel(req, res))
+  get('/referrals/:referralId/service-category/:serviceCategoryId/complexity-level', (req, res) =>
+    referralsController.viewCohortComplexityLevel(req, res)
+  )
   get('/referrals/:id/completion-deadline', (req, res) => referralsController.viewCompletionDeadline(req, res))
   post('/referrals/:id/completion-deadline', (req, res) => referralsController.updateCompletionDeadline(req, res))
   get('/referrals/:id/further-information', (req, res) => referralsController.viewFurtherInformation(req, res))
--- a/server/routes/index.ts
+++ b/server/routes/index.ts
@@ -214,6 +214,9 @@ export default function routes(router: Router, services: Services): Router {
   get('/referrals/:referralId/service-category/:serviceCategoryId/complexity-level', (req, res) =>
     referralsController.viewCohortComplexityLevel(req, res)
   )
+  post('/referrals/:referralId/service-category/:serviceCategoryId/complexity-level', (req, res) =>
+    referralsController.updateCohortComplexityLevel(req, res)
+  )
   get('/referrals/:id/completion-deadline', (req, res) => referralsController.viewCompletionDeadline(req, res))
   post('/referrals/:id/completion-deadline', (req, res) => referralsController.updateCompletionDeadline(req, res))
   get('/referrals/:id/further-information', (req, res) => referralsController.viewFurtherInformation(req, res))
--- a/server/routes/referrals/referralsController.ts
+++ b/server/routes/referrals/referralsController.ts
@@ -243,6 +243,26 @@ export default class ReferralsController {
     ControllerUtils.renderWithLayout(res, view, serviceUser)
   }
 
+  async viewCohortComplexityLevel(req: Request, res: Response): Promise<void> {
+    const { accessToken } = res.locals.user.token
+    const { serviceCategoryId } = req.params
+    const referral = await this.interventionsService.getDraftReferral(accessToken, req.params.referralId)
+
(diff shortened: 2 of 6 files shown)


server/routes/index.ts
```diff
@@ -211,6 +211,12 @@ export default function routes(router: Router, services: Services): Router {
   post('/referrals/:id/service-categories', (req, res) => referralsController.updateServiceCategories(req, res))
   get('/referrals/:id/complexity-level', (req, res) => referralsController.viewComplexityLevel(req, res))
   post('/referrals/:id/complexity-level', (req, res) => referralsController.updateComplexityLevel(req, res))
+  get('/referrals/:referralId/service-category/:serviceCategoryId/complexity-level', (req, res) =>
+    referralsController.selectCohortComplexityLevel(req, res)
+  )
+  post('/referrals/:referralId/service-category/:serviceCategoryId/complexity-level', (req, res) =>
```
andrewburrowsmoj, line 217: To me it's a bit confusing to combine the GET and the POST into a single method on the referralsController.
I don't think that select is the correct wording for updating/creating a linking between complexity and service category

Gweaton: The combining of GET and POST into a single method is something we've recently switched to doing on the project, so I've been following that pattern here - if it's a bit confusing, maybe it's something we should think about refactoring. I'd quite like to do that in a separate PR if possible though, so this one doesn't drag on.

Re: the term "select" - is there an alternative term you think might be more suitable? I chose "select" driven by the user interaction - the page being one for the to PP selecting the complexity level/desired outcomes on a referral (rather than a service category - this is just a way of grouping the desired outcomes/complexity level, rather than setting it on a service category).

I suppose we have `viewX` and `updateX` on other routes but I can't think of a great term that encompasses both, so I'd named the function after the user actions, like we have with `addNumberOfSessionsToActionPlan` in the `ServiceProviderReferralsController`.

Gweaton: Actually, I don't mind splitting this into 2 functions now to save us from having to deal with more tech debt in future. I can refactor the desired outcomes stuff in its own PR

andrewburrowsmoj: That would be great. It's just something that stood out to me as a bit confusing when trying to understand it being new to the code.

server/routes/referrals/referralsController.ts
```diff
@@ -243,6 +243,31 @@ export default class ReferralsController {
...
+    const { accessToken } = res.locals.user.token
+    const referral = await this.interventionsService.getDraftReferral(accessToken, req.params.id)
+
+    if (!referral.serviceCategoryIds) {
+      throw new Error('Attempting to view complexity level without service categories selected')
+    }
+
+    const serviceCategoryId = referral.serviceCategoryIds.find(id => id === req.params.serviceCategoryId)
+
+    if (!serviceCategoryId) {
```
tomsmyers, line 256: nit: seems a roundabout way to do array.includes?

server/routes/index.ts
tomsmyers: personally i prefer it combined into a single method in terms of ability to handle error conditions and readability of those code paths. but i'm not evangelical about it.

server/routes/referrals/referralsController.ts
Gweaton: It really does - didn't realise I'd missed that, thanks

server/routes/referrals/referralsController.ts
```diff
@@ -468,12 +514,16 @@ export default class ReferralsController {
...
-        return res.redirect(`/referrals/${referralId}/form`)
+        try {
+          await this.interventionsService.setDesiredOutcomesForServiceCategory(accessToken, referralId, {
+            serviceCategoryId,
+            ...data.paramsForUpdate,
+          })
+
+          return res.redirect(`/referrals/${referralId}/form`)
+        } catch (e) {
+          formError = createFormValidationErrorOrRethrow(e)
```
tomsmyers, line 525: what is the impact of this? were we not handling formerrors for updating desired outcomes before?

tomsmyers (approved):
couple of small comment. nice work though!

server/routes/index.ts
tomsmyers: in terms of naming - why not `getOrUpdateX`

andrewburrowsmoj: I think getOrUpdate would be better - even though it's not as elegant as a single verb it's more clear about the behaviour

server/routes/referrals/referralsController.ts
Gweaton: I'm really confused about what this method is doing. Andrew quite rightly noted that we weren't using it everywhere, so I thought I'd add it in. I thought it was about surfacing Interventions Service-level errors to the user in a more helpful way, but there seems to be no difference with or without it when testing locally (albeit with stubbed responses, I might not be stubbing them correctly)

server/routes/index.ts
Gweaton: Ok in the interests of getting this over the line so I can start replacing existing behaviour, I'm going to leave this as I've just changed it (to two separate functions). As we'll be removing the older functions for setting desired outcomes and service categories, I might do some refactoring to try and tidy up as part of that work. Does that seem sensible?

PR #303, opened 2021-05-17
IC-1715: Set ServiceCategoryIds for Single Referral Journey when refe…

andrewburrowsmoj:
…rring a new user

Sets the selected ServiceCategoryId automatically for when the referral is a single referral. 

Back end changes are still to be done - assuming this is IC-1682. This ticket can not be started until IC-1670 is finished.

Ensure that the front-end is compatible with back end model changes for cohort referrals.

### DIFF
--- a/integration_tests/integration/referral.spec.js
+++ b/integration_tests/integration/referral.spec.js
@@ -3,6 +3,7 @@ import sentReferralFactory from '../../testutils/factories/sentReferral'
 import serviceCategoryFactory from '../../testutils/factories/serviceCategory'
 import deliusServiceUserFactory from '../../testutils/factories/deliusServiceUser'
 import deliusConvictionFactory from '../../testutils/factories/deliusConviction'
+import interventionFactory from '../../testutils/factories/intervention'
 
 describe('Referral form', () => {
   beforeEach(() => {
@@ -14,7 +15,13 @@ describe('Referral form', () => {
 
   it('User starts a referral, fills in the form, and submits it', () => {
     const deliusServiceUser = deliusServiceUserFactory.build({ firstName: 'Alex' })
-
+    const randomInterventionId = '99ee16d3-130a-4d8f-97c5-f1a42119a382'
+    const intervention = interventionFactory.build({
+      title: 'Better solutions (anger management)',
+      serviceCategories: ['accommodation'].map(name => serviceCategoryFactory.build({ name })),
+      id: randomInterventionId,
+      serviceProvider: { name: 'Harmony Living' },
+    })
     const serviceCategory = serviceCategoryFactory.build({
       name: 'accommodation',
       desiredOutcomes: [
@@ -108,6 +115,7 @@ describe('Referral form', () => {
     const sentReferral = sentReferralFactory.fromFields(completedDraftReferral).build()
 
     cy.stubGetServiceUserByCRN('X123456', deliusServiceUser)
+    cy.stubGetIntervention(intervention.id, intervention)
     cy.stubCreateDraftReferral(draftReferral)
     cy.stubGetServiceCategory(serviceCategory.id, serviceCategory)
     cy.stubGetSentReferrals([])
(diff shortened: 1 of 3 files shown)


integration_tests/integration/referral.spec.js
```diff
@@ -14,7 +15,13 @@ describe('Referral form', () => {
 
   it('User starts a referral, fills in the form, and submits it', () => {
     const deliusServiceUser = deliusServiceUserFactory.build({ firstName: 'Alex' })
-
+    const randomInterventionId = '99ee16d3-130a-4d8f-97c5-f1a42119a382'
+    const intervention = interventionFactory.build({
+      title: 'Better solutions (anger management)',
+      serviceCategories: ['accommodation'].map(name => serviceCategoryFactory.build({ name })),
```
Gweaton, line 21: Did you mean to put more in here, or is there a reason you've mapped over the single element array, rather than `serviceCategoryFactory.build({name: accommodation})`?

server/routes/referrals/referralsController.test.ts
```diff
@@ -69,31 +69,63 @@ describe('GET /intervention/:id/refer', () => {
...
+        id: 'b33c19d1-7414-4014-b543-e543e59c5b39',
+        name: 'social inclusion',
+      })
+      const intervention = interventionFactory.build({
+        serviceCategories: [serviceCategory1, serviceCategory2],
+      })
+      beforeEach(() => {
+        interventionsService.getIntervention.mockResolvedValue(intervention)
+      })
+      it('should update referral with service user and redirect to ...', async () => {
```
Gweaton, line 120: Sorry if I didn't explain this quite right earlier, this should redirect to the referral form like we do above - the form will be a bit different to the one we use for a single-service referral - the page where we select the relevant service categories comes as step 2 in the task list, if I remember correctly, but we can add that in later.

server/routes/referrals/referralsController.ts
```diff
@@ -101,21 +101,26 @@ export default class ReferralsController {
...
+        this.interventionsService.getIntervention(res.locals.user.token.accessToken, interventionId),
+      ])
+      if (intervention.serviceCategories.length === 1) {
+        await this.interventionsService.patchDraftReferral(res.locals.user.token.accessToken, referral.id, {
+          serviceUser: this.interventionsService.serializeDeliusServiceUser(serviceUser),
+          serviceCategoryIds: [intervention.serviceCategories[0].id],
+        })
+        res.redirect(303, `/referrals/${referral.id}/form`)
+      } else {
+        await this.interventionsService.patchDraftReferral(res.locals.user.token.accessToken, referral.id, {
```
Gweaton, line 115: Just thinking out loud - if we're redirecting back to the form (see other comment), do you think it could be cleaner to do something like:
```
const serviceCategoryIds = intervention.ServiceCategories.length === 1 ? [intervention.serviceCategories[0].id] : null

await this.interventionsService.patchDraftReferral(res.locals.user.token.accessToken, referral.id, {
  serviceUser: this.interventionsService.serializeDeliusServiceUser(serviceUser),
  serviceCategoryIds: serviceCategoryIds,
})

res.redirect(303, `/referrals/${referral.id}/form`)
```

I guess we'd always be patching `serviceCategoryIds` here, do you think that'd have any performance issues (or other non-performance issues)?

Gweaton (commented):
Just a couple of small things

server/routes/referrals/referralsController.test.ts
andrewburrowsmoj: Ok. I misunderstood it as redirecting to a new page.
I've changed it as you suggested

server/routes/referrals/referralsController.ts
andrewburrowsmoj: Yes, I think it was also because I was expecting the rediction to be different as well. I've used your suggestion.

integration_tests/integration/referral.spec.js
andrewburrowsmoj: I did it as an array to make it easier to test it with a cohort service behaviour by just plugging in a second string into the array

Gweaton (approved):
Nice, thanks for making those changes.

PR #307, opened 2021-05-19
IC-1541: Display user-facing error when trying to set a conflicting time & date for an appointment

Gweaton:

Displays an error to the user when editing an appointment and trying to set the time & date to that of another pre-scheduled appointment. In this case, we'll get a 409 status code from the Community API.

I'm not 100% convinced by my approach here - I initially tried using our existing `formError` pattern to pass in this error, but as we're not targeting a single `input` field here, like we do elsewhere (we're targeting a collection of `input` fields, so we can't use the existing macros - see screenshot), I've had to handle this slightly differently, hence the addition of the `serverError` argument to the `EditSessionPresenter`.

Without this change, the UI takes the user back to the intervention progress page, as if it's been a successful edit, even though the Community API has rejected this update.

## Screenshot

image

### DIFF
--- a/server/routes/serviceProviderReferrals/editSessionPresenter.ts
+++ b/server/routes/serviceProviderReferrals/editSessionPresenter.ts
@@ -8,8 +8,9 @@ import { FormValidationError } from '../../utils/formValidationError'
 export default class EditSessionPresenter {
   constructor(
     private readonly appointment: ActionPlanAppointment,
-    private readonly error: FormValidationError | null = null,
-    private readonly userInputData: Record<string, unknown> | null = null
+    private readonly validationError: FormValidationError | null = null,
+    private readonly userInputData: Record<string, unknown> | null = null,
+    private readonly serverError: FormValidationError | null = null
   ) {}
 
   readonly text = {
@@ -18,7 +19,11 @@ export default class EditSessionPresenter {
 
   private readonly utils = new PresenterUtils(this.userInputData)
 
-  readonly errorSummary = PresenterUtils.errorSummary(this.error)
+  readonly errorSummary = this.serverError
+    ? PresenterUtils.errorSummary(this.serverError)
+    : PresenterUtils.errorSummary(this.validationError)
+
+  readonly serverErrorMessage = PresenterUtils.errorMessage(this.serverError, 'session-input')
 
   readonly fields = {
     date: this.utils.dateValue(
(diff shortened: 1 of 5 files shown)


server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
```diff
@@ -435,6 +435,7 @@ export default class ServiceProviderReferralsController {
 
     let userInputData: Record<string, unknown> | null = null
     let formError: FormValidationError | null = null
+    let serverError: FormValidationError | null = null
```
Gweaton, line 438: should this be a `ServerError` type? I left it as a `FormValidationError` as it's got the same interface and because it is being displayed at the top of the form on submit, but I could duplicate that interface.

andrewburrowsmoj: I think even a serverError can be considered a FormValidationError if it's 4xx (generally)...

I would differentiate the errors based on if it's the user's fault or out of the user's control in which case we should tell the user to contract support (or retry).

Gweaton: Yes, it'll go to a page similar to the one you showed this morning (ugly generic error page)

server/views/serviceProviderReferrals/editSession.njk
```diff
@@ -12,21 +12,31 @@
...
-        {% endif %}
+        <input type="hidden" name="_csrf" value="{{csrfToken}}">
 
         <h1 class="govuk-heading-xl">{{ presenter.text.title }}</h1>
 
-        {{ govukDateInput(dateInputArgs) }}
-        {{ appTimeInput(timeInputArgs) }}
-        {{ govukDateInput(durationDateInputArgs) }}
+        <div id='session-input' class="{{ serverError.classes }}">
+          {% if serverError.text !== null %}
```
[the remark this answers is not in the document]
andrewburrowsmoj: Perhaps could encapsulate it into a single error object?

{ message: string; error: {field: string; message: string}[], classes: string }

and have serverError() return this or `null` rather than checking `text`?

server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
andrewburrowsmoj: ah ok, so in this case, in my opinion it still makes sense for it to be a FromValidationError.

server/views/serviceProviderReferrals/editSession.njk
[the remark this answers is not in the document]
Gweaton: Good plan, I've done this, although without the `error` field inside.

server/routes/serviceProviderReferrals/editSessionPresenter.ts
```diff
@@ -9,7 +9,8 @@ export default class EditSessionPresenter {
   constructor(
     private readonly appointment: ActionPlanAppointment,
     private readonly error: FormValidationError | null = null,
```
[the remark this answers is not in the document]
Gweaton: Yep, have renamed this

andrewburrowsmoj (approved):

PR #308, opened 2021-05-19
IC-1711/IC-1712: Added desired outcomes and complexity levels on 'Mak…

andrewburrowsmoj:
…e a referral' form for cohort referrals

Adjusts the "Make a referral" page to be able to process cohort referrals to show desired outcomes and complexity level links.

It also adjusts the behaviour of "save and continue" for complexity level page if the referral is a cohort referral.

Before the user has selected service categories the nested desired outcomes and complexity will not be visible. However for this PR it is assumed that the service category ids will have been selected already. Single referrals will still look like the existing screen does.

It was not possible to separate this into two separate commits for desired outcome and complexity level which is why they are combined together.

**note**: I've confirmed with @rwk-moj if it's acceptable to not completely adhere to the line separators as depicted in the examples he provided. He says it's okay but he will need to approve it first. The reason is because a pure css solution isn't possible given that sections aren't separated but instead each row is a new \<li\> within the same \<ol\>. In order to replicate the image I'd either have to change the HTML elements or write some messy nun-jucks code 

Provide ability for the user to view the status of cohort referrals

image

### DIFF
--- a/integration_tests/integration/make_a_referral/referralSectionVerifier.js
+++ b/integration_tests/integration/make_a_referral/referralSectionVerifier.js
@@ -1,0 +1,75 @@
+// eslint-disable-next-line max-classes-per-file
+export default class ReferralSectionVerifier {
+  static get verifySection() {
+    return new _ReferralSectionChecker()
+  }
+}
+function hrefAttrChainer(isActive) {
+  return isActive ? 'have.attr' : 'not.have.attr'
+}
+class _ReferralSectionChecker {
+  reviewServiceUserInformation(activeLinks) {
+    cy.get('[data-cy=url]')
+      .contains('Confirm service user’s personal details')
+      .should(hrefAttrChainer(activeLinks.confirmServiceUserDetails), 'href')
+    cy.get('[data-cy=url]')
+      .contains('Service user’s risk information')
+      .should(hrefAttrChainer(activeLinks.riskInformation), 'href')
+    cy.get('[data-cy=url]')
+      .contains('Service user’s needs and requirements')
+      .should(hrefAttrChainer(activeLinks.needsAndRequirements), 'href')
+    return this
+  }
+
+  interventionReferralDetails(activeLinks) {
+    cy.get('[data-cy=url]')
+      .contains('Confirm the relevant sentence for the accommodation referral')
+      .should(hrefAttrChainer(activeLinks.relevantSentence), 'href')
+    cy.get('[data-cy=url]')
+      .contains('Select required complexity level')
+      .should(hrefAttrChainer(activeLinks.requiredComplexityLevel), 'href')
+    cy.get('[data-cy=url]')
+      .contains('Select desired outcomes')
+      .should(hrefAttrChainer(activeLinks.desiredOutcomes), 'href')
+    cy.get('[data-cy=url]')
+      .contains('Enter when the accommodation service need to be completed')
+      .should(hrefAttrChainer(activeLinks.completedDate), 'href')
+    cy.get('[data-cy=url]').contains('Enter enforceable days used').should(hrefAttrChainer(activeLinks.rarDays), 'href')
+    cy.get('[data-cy=url]')
+      .contains('Further information for service provider')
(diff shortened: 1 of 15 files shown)


integration_tests/integration/make_a_referral/referralSectionVerifier.js
```diff
@@ -0,0 +1,65 @@
...
+export default class ReferralSectionVerifier {
+  static get verifySection() {
+    return new _ReferralSectionChecker()
+  }
+}
+function hrefAttrChainer(isActive) {
+  return isActive ? 'have.attr' : 'not.have.attr'
+}
+class _ReferralSectionChecker {
+  // eslint-disable-next-line class-methods-use-this
```
Gweaton, line 11: I wonder if we should turn this rule off, I don't think it's that useful a rule - what do you think?

Gweaton (approved):
This looks like a lot of work, well done 🏅 

As you said, this is a large PR - I've taken a look as best I can and it looks like it makes sense.

I find it a bit tricky to review large PRs like this in close detail (particularly when it's all a single commit) - in future, breaking things down into smaller commits or even smaller PRs would make things a bit easier, although I appreciate it's pretty hard to do so with the referral form and the number of changes you've had to make!

For future reference, things that could have gone in their own commit:
- Updating the README
- Adding the ReferralSectionVerifier (possibly for existing behaviour, or before it's used) - a detailed commit message might make this easier to review so it's not coupled with all the changes.
- Renaming methods in the `factories/draftReferral.ts`


Despite all that though, it does look like a lot of these changes did need to happen at once, so don't worry about it too much!

Have just left a couple of small comments.

integration_tests/integration/make_a_referral/referralSectionVerifier.js
```diff
@@ -0,0 +1,65 @@
...
+  reviewServiceUserInformation(activeLinks) {
```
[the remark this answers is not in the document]
andrewburrowsmoj: yes that makes sense, I think the array solution was to reduce the amount of code repetition but then that sacrifices readability. I'll do as you suggest

integration_tests/integration/make_a_referral/referralSectionVerifier.js
andrewburrowsmoj: Sorry, I'm not sure what you mean?

This rule is required to ensure that the linter passes. It complains because "This" isn't used within the class method

Gweaton: I meant, should we turn the rule off in eslint config? I don't think it's a very useful linting rule that we should enforce.

andrewburrowsmoj:
I don't know how to respond to your main comment. You make good points - Tom keeps mentioning the same thing to me. It's a habit of squashing everything for a single PR.
I made most of these changes before he mentioned that last time. Otherwise I would likely have split things out a bit more. Do you still want me to split things out ?

Gweaton:
> I don't know how to respond directly to your main comment so I will have to make a new comment here.... You make good points - Tom keeps mentioning the same thing to me. It's a habit of squashing everything for a single PR.
> I made most of these changes before he mentioned that last time. Otherwise I would likely have split things out a bit more. Do you still want me to split things out ?

No, it's fine for this one, just something that would be helpful in future!

integration_tests/integration/make_a_referral/referralSectionVerifier.js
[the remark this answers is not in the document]
andrewburrowsmoj: I've made the fix that you've asked for.

integration_tests/integration/make_a_referral/referralSectionVerifier.js
andrewburrowsmoj: In this case the rule is necessary to avoid linting failures.

It fails because `this` is not used within the method; it's probably suggesting that the method should be static.

Instead what I've done now is increased the chaining so that these methods return `this` so no need to add these off-rules anymore.

PR #313, opened 2021-05-24
IC-1771 Display each selected service category for the referral in the SP details page.

tomsmyers:

This change adds a field for "Service type" and a new section for each selected
service category. The selected complexity level is shown for each service category.
The desired outcomes are currently not shown. This will come in a later commit.

Screenshot 2021-05-24 at 17 32 44

Accurate referral details page for cohort referrals.

### DIFF
--- a/integration_tests/integration/serviceProviderReferrals.spec.js
+++ b/integration_tests/integration/serviceProviderReferrals.spec.js
@@ -35,8 +35,6 @@ describe('Service provider referrals dashboard', () => {
           serviceUser: { firstName: 'George', lastName: 'Michael' },
           serviceCategoryId: accommodationServiceCategory.id,
           serviceCategoryIds: [accommodationServiceCategory.id, socialInclusionServiceCategory.id],
-          // desiredOutcomesIds: ['65924ac6-9724-455b-ad30-906936291421', 'e7f199de-eee1-4f57-a8c9-69281ea6cd4d'],
-          // complexityLevelId: '110f2405-d944-4c15-836c-0c6684e2aa78',
         },
       }),
       sentReferralFactory.build({
@@ -47,8 +45,16 @@ describe('Service provider referrals dashboard', () => {
           serviceUser: { firstName: 'Jenny', lastName: 'Jones', crn: 'X123456' },
           serviceCategoryId: socialInclusionServiceCategory.id,
           serviceCategoryIds: [accommodationServiceCategory.id, socialInclusionServiceCategory.id],
-          // desiredOutcomesIds: ['65924ac6-9724-455b-ad30-906936291421', 'e7f199de-eee1-4f57-a8c9-69281ea6cd4d'],
-          // complexityLevelId: '110f2405-d944-4c15-836c-0c6684e2aa78',
+          complexityLevels: [
+            {
+              serviceCategoryId: accommodationServiceCategory.id,
+              complexityLevelId: 'd0db50b0-4a50-4fc7-a006-9c97530e38b2',
+            },
+            {
+              serviceCategoryId: socialInclusionServiceCategory.id,
+              complexityLevelId: '110f2405-d944-4c15-836c-0c6684e2aa78',
+            },
+          ],
         },
       }),
     ]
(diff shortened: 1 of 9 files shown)


server/routes/serviceProviderReferrals/showReferralPresenter.ts
```diff
@@ -39,24 +41,41 @@ export default class ShowReferralPresenter {
...
+
+    return items
+  }
+
+  private getReferralComplexityLevelForServiceCategory(serviceCategory: ServiceCategory): ComplexityLevel | null {
+    const complexityLevelId = this.sentReferral.referral.complexityLevels.find(
+      it => it.serviceCategoryId === serviceCategory.id
+    )?.complexityLevelId
+
+    return serviceCategory.complexityLevels.find(it => it.id === complexityLevelId) || null
```
Gweaton, line 72: How likely is it that we'll return `null` here? I've been thinking about how nullable our codebase is recently, and wondering if throwing errors would make it easier to work with? Maybe not, as it makes the page unusable vs rendering an empty field, although would an empty field be useful from a user perspective?

server/routes/serviceProviderReferrals/showReferralPresenter.ts
```diff
@@ -39,24 +41,41 @@ export default class ShowReferralPresenter {
...
+  serviceCategorySection(serviceCategory: ServiceCategory, tagMacro: (args: TagArgs) => string): SummaryListItem[] {
+    const items: SummaryListItem[] = []
+
+    const complexityLevel = this.getReferralComplexityLevelForServiceCategory(serviceCategory)
+    if (complexityLevel === null) {
+      items.push({ key: 'Complexity level', lines: ['Level not found'] })
+    } else {
+      items.push({
+        key: 'Complexity level',
+        lines: [tagMacro(PresenterUtils.complexityLevelTagArgs(complexityLevel)), complexityLevel.description],
```
Gweaton, line 58: Does`lines: [complexityLevel === null ? 'Level not found' : tagMacro(PresenterUtils.complexityLevelTagArgs(complexityLevel)), complexityLevel.description]` work here? Would mean we can remove the conditionals and the pushing into an array here, right?

server/utils/presenterUtils.ts
```diff
@@ -284,4 +286,34 @@ export default class PresenterUtils {
 
     return notFoundMessage
   }
+
+  static complexityLevelTagArgs(complexityLevel: ComplexityLevel): TagArgs {
+    const title = complexityLevel?.title.toUpperCase()
```
Gweaton, line 291: I don't think you need the `?` here?

integration_tests/integration/serviceProviderReferrals.spec.js
```diff
@@ -19,25 +20,35 @@ describe('Service provider referrals dashboard', () => {
...
         referenceNumber: 'ABCABCA2',
         referral: {
-          serviceCategoryId: socialInclusionServiceCategory.id,
+          interventionId: intervention.id,
           serviceUser: { firstName: 'Jenny', lastName: 'Jones', crn: 'X123456' },
-          desiredOutcomesIds: ['65924ac6-9724-455b-ad30-906936291421', 'e7f199de-eee1-4f57-a8c9-69281ea6cd4d'],
-          complexityLevelId: '110f2405-d944-4c15-836c-0c6684e2aa78',
+          serviceCategoryId: socialInclusionServiceCategory.id,
+          serviceCategoryIds: [accommodationServiceCategory.id, socialInclusionServiceCategory.id],
+          // desiredOutcomesIds: ['65924ac6-9724-455b-ad30-906936291421', 'e7f199de-eee1-4f57-a8c9-69281ea6cd4d'],
```
Gweaton, line 50: Shall we get rid of these comments now?

Gweaton (approved):
Nice work - a couple of changes, nothing major, but have approved so you can merge when addressed (if you think necessary)

server/routes/serviceProviderReferrals/showReferralPresenter.ts
tomsmyers: yeh i know what you mean. 

like it's totally an error condition if the selected complexity level doesn't exist for the selected service category. i'd rather not kill the page, but we could log an error?

server/utils/presenterUtils.ts
tomsmyers: ah good spot - this is probably left over from me moving the code from elsewhere

server/routes/serviceProviderReferrals/showReferralPresenter.ts
tomsmyers: i'm gonna leave it for now and when i do the desiredoutcomes i'll look at refactoring

integration_tests/integration/serviceProviderReferrals.spec.js
tomsmyers: done

PR #316, opened 2021-05-26
IC-1726: Display contract type in referral cancellation journey

Gweaton:

Replaces service category with contract type name in PP referral cancellation pages.

To make the page work with cohort services and not try to display multiple service categories.


tomsmyers (approved):
nice

PR #319, opened 2021-05-26
add feedback tag to the page layout

tomsmyers:

add feedback tag to the page layout

ensure any initial PP users know the service is under active development.


server/views/partials/layout.njk
```diff
@@ -33,6 +34,15 @@
 
 {% block header %}
   {% include "./header.njk" %}
+
+  {{ govukPhaseBanner({
+    tag: {
+      text: "alpha"
```
Gweaton, line 40: As discussed on slack, this should be beta

Gweaton (approved):
Might be worth updating the commit message, but only if you have the energy!

PR #321, opened 2021-05-26
IC-1727: Use contract type name on SP dashboard and referral progress page

Gweaton:

Uses contract type name on SP dashboard & referral progress page rather than service category.

To support cohort referrals.

### DIFF
--- a/integration_tests/integration/serviceProviderReferrals.spec.js
+++ b/integration_tests/integration/serviceProviderReferrals.spec.js
@@ -20,10 +20,14 @@ describe('Service provider referrals dashboard', () => {
     const accommodationServiceCategory = serviceCategoryFactory.build({ name: 'accommodation' })
     const socialInclusionServiceCategory = serviceCategoryFactory.build({ name: 'social inclusion' })
 
-    const intervention = interventionFactory.build({
+    const personalWellbeingIntervention = interventionFactory.build({
       contractType: { code: 'PWB', name: 'Personal wellbeing' },
-      serviceCategory: accommodationServiceCategory, // fixme: remove
       serviceCategories: [accommodationServiceCategory, socialInclusionServiceCategory],
+    })
+
+    const socialInclusionIntervention = interventionFactory.build({
+      contractType: { code: 'SOC', name: 'Social inclusion' },
+      serviceCategories: [socialInclusionServiceCategory],
     })
 
     const sentReferrals = [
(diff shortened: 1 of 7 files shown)


integration_tests/integration/serviceProviderReferrals.spec.js
```diff
@@ -241,10 +242,20 @@ describe('Service provider referrals dashboard', () => {
...
+      contractType: { code: 'SOC', name: 'Social inclusion' },
+      serviceCategories: [serviceCategory],
+    })
+
     const selectedDesiredOutcomesIds = [desiredOutcomes[0].id, desiredOutcomes[1].id]
     const referralParams = {
-      referral: { serviceCategoryId: serviceCategory.id, desiredOutcomesIds: selectedDesiredOutcomesIds },
+      referral: {
+        interventionId: accommodationIntervention.id,
+        serviceCategoryId: serviceCategory.id,
```
tomsmyers, line 256: should we replace this with (or just add) 'serviceCategoryIds' here now too?

server/routes/serviceProviderReferrals/dashboardPresenter.ts
```diff
@@ -18,12 +18,12 @@ export default class DashboardPresenter {
...
   readonly tableRows: SortableTableRow[] = this.referrals.map(referral => {
-    const { serviceCategoryId } = referral.referral
-    const serviceCategory = this.serviceCategories.find(aServiceCategory => aServiceCategory.id === serviceCategoryId)
-    if (serviceCategory === undefined) {
-      throw new Error(`Expected serviceCategories to contain service category with ID ${serviceCategoryId}`)
+    const interventionForReferral = this.interventions.find(
+      intervention => intervention.id === referral.referral.interventionId
+    )
+    if (interventionForReferral === undefined) {
+      throw new Error(`Expected referral to be linked to an intervention with ID ${referral.referral.interventionId}`)
```
tomsmyers, line 25: just an observation: it's a shame we can't add structured arguments to errors like we can with log messages

tomsmyers: also i think this error message could be improved. with no context i don't think it means a lot. how about something more descriptive like "couldn't populate dashboard row because the referral's intervention is missing from the list passed to the presenter" or something

tomsmyers (approved):
couple small comments, nothing major.

integration_tests/integration/serviceProviderReferrals.spec.js
Gweaton: It'll break if we remove this, as it's still used on the action plan page, but I've added `serviceCategoryIds` and added a TODO for removing the old field when that task is picked up.

server/routes/serviceProviderReferrals/dashboardPresenter.ts
Gweaton: Have updated this error message, you're right - it wasn't that helpful. It's still impossible for it to occur, as we fetch those interventions from the referral's `interventionId`, but 🤷

PR #324, opened 2021-05-27
IC-1727: Use contract type in SP referral assignment pages

Gweaton:

Use contract type in SP referral assignment pages, rather than service category, now that we have cohort services that may
have multiple.

This commit also updates tests to ensure we're presenting the service type in sentence case, rather than lower case, as advised by the content team.

To update the journey to support cohort referrals.

### DIFF
--- a/integration_tests/integration/serviceProviderReferrals.spec.js
+++ b/integration_tests/integration/serviceProviderReferrals.spec.js
@@ -170,7 +170,6 @@ describe('Service provider referrals dashboard', () => {
     const referralParams = {
       referral: {
         interventionId: intervention.id,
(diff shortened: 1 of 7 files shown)


integration_tests/integration/serviceProviderReferrals.spec.js
```diff
@@ -170,7 +170,6 @@ describe('Service provider referrals dashboard', () => {
     const referralParams = {
       referral: {
         interventionId: intervention.id,
-        serviceCategoryId: intervention.serviceCategories[0].id,
```
tomsmyers, line 173: great to see these dissapearing

PR #325, opened 2021-05-27
IC-1808: 🤖 Get PRs for out of date and insecure dependencies

sldblog:

Turns on dependabot on the project

To get library upgrades in the pipeline quickly and painlessly (given we trust our tests)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+- package-ecosystem: npm
+  directory: "/"
+  schedule:
+    interval: daily
+    time: "03:00"
+    timezone: Europe/London
+  open-pull-requests-limit: 10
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
+updates:
+- package-ecosystem: npm
+  directory: "/"
+  schedule:
+    interval: daily
+    time: "03:00"
+    timezone: Europe/London
+  open-pull-requests-limit: 10
```
Gweaton, line 9: Do you think we could turn this down to 5, or is that too few? I personally find it a bit overwhelming when there are loads of PRs open, despite the importance of keeping dependencies up to date, what do you think?

Gweaton (approved):

PR #326, opened 2021-05-27
Ic 1683: Adds "Select Service Categories" for cohort referrals on "Make a referral" screen

andrewburrowsmoj:

Adds "Select Service Categories" for cohort referrals on "Make a referral" screen

Provide ability for SP user to select the service categories required for draft referral.

image

### DIFF
--- a/integration_tests/integration/make_a_referral/referralSectionVerifier.js
+++ b/integration_tests/integration/make_a_referral/referralSectionVerifier.js
@@ -41,8 +41,16 @@ class _ReferralSectionChecker {
     return this
   }
 
+  selectServiceCagories(activeLinks) {
+    cy.get("[data-cy=url]:contains('Select service categories for the Womans Service referral')").should(
+      hrefAttrChainer(activeLinks.selectServiceCategories),
+      'href'
+    )
+    return this
+  }
+
   cohortInterventionReferralDetails(activeLinks) {
-    cy.get("[data-cy=url]:contains('Confirm the relevant sentence for the intervention referral')").should(
+    cy.get("[data-cy=url]:contains('Confirm the relevant sentence for the Womans Service referral')").should(
       hrefAttrChainer(activeLinks.relevantSentence),
       'href'
     )
@@ -59,7 +67,7 @@ class _ReferralSectionChecker {
       .eq(1)
       .should(hrefAttrChainer(activeLinks.desiredOutcomes1), 'href')
     cy.get('[data-cy=url]')
-      .contains('Enter when the intervention service need to be completed')
+      .contains('Enter when the Womans Service referral need to be completed')
       .should(hrefAttrChainer(activeLinks.completedDate), 'href')
     cy.get('[data-cy=url]').contains('Enter enforceable days used').should(hrefAttrChainer(activeLinks.rarDays), 'href')
     cy.get('[data-cy=url]')
--- a/integration_tests/integration/referral.spec.js
+++ b/integration_tests/integration/referral.spec.js
@@ -342,10 +342,15 @@ describe('Referral form', () => {
         ],
       })
       const intervention = interventionFactory.build({
+        title: 'Womans Service',
         serviceCategories: [accommodationServiceCategory, socialInclusionServiceCategory],
+        contractType: {
+          code: 'WMS',
+          name: 'Womans Service',
+        },
       })
       const draftReferral = draftReferralFactory.serviceUserSelected().build({
-        serviceCategoryIds: [accommodationServiceCategory.id, socialInclusionServiceCategory.id],
+        serviceCategoryIds: null,
         interventionId: intervention.id,
         serviceProvider: {
           name: 'Harmony Living',
@@ -355,7 +360,17 @@ describe('Referral form', () => {
       const completedServiceUserDetailsDraftReferral = draftReferralFactory
         .filledFormUpToNeedsAndRequirements([accommodationServiceCategory, socialInclusionServiceCategory])
         .build({
-          serviceCategoryIds: [accommodationServiceCategory.id, socialInclusionServiceCategory.id],
+          serviceCategoryIds: null,
+          interventionId: intervention.id,
+          serviceProvider: {
+            name: 'Harmony Living',
+          },
+        })
+
+      const completedSelectingServiceCategories = draftReferralFactory
+        .filledFormUpToNeedsAndRequirements([accommodationServiceCategory, socialInclusionServiceCategory])
+        .selectedServiceCategories([accommodationServiceCategory, socialInclusionServiceCategory])
+        .build({
           interventionId: intervention.id,
           serviceProvider: {
             name: 'Harmony Living',
@@ -365,7 +380,6 @@ describe('Referral form', () => {
       const completedDraftReferral = draftReferralFactory
         .filledFormUpToFurtherInformation([accommodationServiceCategory, socialInclusionServiceCategory])
         .build({
-          serviceCategoryIds: [accommodationServiceCategory.id, socialInclusionServiceCategory.id],
           interventionId: intervention.id,
           serviceProvider: {
             name: 'Harmony Living',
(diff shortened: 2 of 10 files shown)


integration_tests/integration/make_a_referral/referralSectionVerifier.js
```diff
@@ -41,6 +41,14 @@ class _ReferralSectionChecker {
     return this
   }
 
+  selectServiceCagories(activeLinks) {
+    cy.get("[data-cy=url]:contains('Select service types')").should(
```
tomsmyers, line 45: i think 'types' isn't the right name here. i think 'service type' is what we are using to refer to the top level contract type, and these are 'service categories'.

integration_tests/integration/referral.spec.js
```diff
@@ -463,12 +466,33 @@ describe('Referral form', () => {
...
+      ReferralSectionVerifier.verifySection
+        .reviewServiceUserInformation({
+          confirmServiceUserDetails: true,
+          riskInformation: true,
+          needsAndRequirements: true,
+        })
+        .selectServiceCagories({ selectServiceCategories: true })
+        .checkYourAnswers({ checkAnswers: false })
+      cy.contains('Select service types').click()
+      cy.get('h1').contains('What service types are you referring Alex to?')
```
tomsmyers, line 479: same comment here about 'type' vs 'category'

server/routes/referrals/referralFormPresenter.ts
```diff
@@ -1,118 +1,123 @@
...
-import ServiceCategory from '../../models/serviceCategory'
+import Intervention from '../../models/intervention'
 
 export default class ReferralFormPresenter {
   private readonly taskValues: typeof ReferralFormPresenter.TaskValues.prototype
 
   private readonly sectionValues: typeof ReferralFormPresenter.SectionValues.prototype
 
-  constructor(private readonly referral: DraftReferral, private readonly serviceCategories: ServiceCategory[]) {
+  private readonly formSectionBuilder: typeof ReferralFormPresenter.FormSectionBuilder.prototype
```
tomsmyers, line 10: what are these `typeof ...prototype` types about?

server/routes/referrals/referralFormPresenter.ts
```diff
@@ -1,118 +1,123 @@
...
     if (this.isCohortIntervention) {
-      referralDetails = {
-        type: 'multi',
-        title: 'Add intervention referral details',
+      return this.formSectionBuilder.buildCohortReferralSections()
+    }
+    return this.formSectionBuilder.buildSingleReferralSections()
+  }
+
+  private static FormSectionBuilder = class {
```
tomsmyers, line 34: ah now i see what those type definitions were above. what's the reason for having these anonymous class member vars here, vs just having proper classes so we can use the types without having to use `typeof`?

server/routes/referrals/referralFormPresenter.ts
```diff
@@ -1,118 +1,123 @@
...
-        taskListSections: [
+        tasks: [
           {
-            tasks: [
-              {
-                title: `Confirm the relevant sentence for the intervention referral`,
-                url: this.calculateTaskUrl('relevant-sentence', this.taskValues.needsAndRequirements),
-              },
-            ],
+            title: `Select service types for the ${this.intervention.title} referral`,
```
tomsmyers, line 100: i think this should be the contract type not the intervention title, to be consistent with other pages. 

```Select service types for the ${this.intervention.contractType.name} referral```

Comment (tomsmyers) [reply server/routes/referrals/referralFormPresenter.ts:100]: and as mentioned above, i think it should be 'service categories' not types

File: server/routes/referrals/referralFormPresenter.ts [hunk 6]
```diff
@@ -148,48 +153,127 @@ export default class ReferralFormPresenter {
...
+        number: '3',
+        status: this.calculateStatus(
+          this.sectionValues.serviceCategoryReferralDetails,
+          selectServiceCategoriesSection.status!
+        ),
+        taskListSections: [
+          {
+            tasks: [
+              {
+                title: `Confirm the relevant sentence for the intervention referral`,
```
tomsmyers, line 172: i know this is just moved code, but i wondered whether 'intervention referral' here should be `${this.intervention.contractType.name}` referral, to be consistent with this pattern elsewhere.

server/routes/referrals/referralFormPresenter.ts
```diff
@@ -148,48 +153,127 @@ export default class ReferralFormPresenter {
...
+                url: this.calculateTaskUrl('relevant-sentence', this.taskValues.cohortServiceCategories),
+              },
+            ],
+          },
+        ]
+          .concat(
+            this.intervention.serviceCategories
+              .filter(serviceCat => {
+                if (this.referral.serviceCategoryIds != null) {
```
tomsmyers, line 181: nit: this should be `!==` to avoid unexpected null coalescing.

tomsmyers: unless that's what we want here.

server/routes/referrals/referralFormPresenter.ts
```diff
@@ -1,118 +1,123 @@
...
+        referralDetailsSection,
+        checkYourAnswersSection,
+      ]
+    }
+
+    private buildReviewServiceUserInformationSection(): ReferralFormSingleListSectionPresenter {
+      return {
+        type: 'single',
+        title: 'Review service user’s information',
+        number: '1',
```
tomsmyers, line 68: what is 'number' used for ?

also a small code smell: a field called 'number' that's a string?

server/routes/referrals/referralFormPresenter.ts
andrewburrowsmoj: Sure I will change it also in the section title as well ->
Add ${this.intervention.contractType.name} referral details

server/routes/referrals/referralFormPresenter.ts
andrewburrowsmoj: thanks for the spot

server/routes/referrals/referralFormPresenter.ts
andrewburrowsmoj: This number is what is displayed for the section header, i.e.:
```
1. Review service user's information
```

I agree that it would be better if it was done with css. This was the implementation before I touched on changing it for cohort referrals; would you like me to make this change as well?

server/routes/referrals/referralFormPresenter.ts
andrewburrowsmoj: I was trying to get around the eslint rule for `max-classes-per-file`, but I've disabled the rule now for this file, so I'll change them to classes to make it a bit more readable.

server/routes/referrals/referralFormPresenter.ts
tomsmyers: ah i see - no it's ok. the main reason i noticed this was the conditional 3?4 for cohort referrals. seems a little flakey. let's leave for now and change to auto numbering the sections later.

server/routes/referrals/referralFormPresenter.ts
tomsmyers: they could always go in their own files too.

integration_tests/integration/referral.spec.js
andrewburrowsmoj: changed

server/routes/referrals/referralFormPresenter.ts
andrewburrowsmoj: done

server/routes/referrals/referralFormPresenter.ts
andrewburrowsmoj: changed

Gweaton (approved):

andrewburrowsmoj:
Hi @tomsmyers. 
Were you going to put in some of your own changes in this branch as well?

PR #327, opened 2021-05-27
change the two list referrals API endpoints to remove the no longer used query params

tomsmyers:

change 'sent referrals' API endpoints to remove the no longer used query params.

this change was made on the API during the access work, but never updated on the UI. luckily the endpoints continued to function, the query params were just ignored. however there are failing contracts associated with the old style endpoints. 

ensure we are using the latest API specs, and get the contract tests green again.


Gweaton (approved):
In `integration_tests/plugins/index.js` we have a `stubGetSentReferrals` function which I don't think relates to a a real function but mocks a response from the `/sent-referrals` endpoint. We call it in most of the tests which rely on the old `getReferralsSentBy/ToX` function, but I wonder if it'd make it clearer to rename it to this new function in those tests too?

tomsmyers:
> In `integration_tests/plugins/index.js` we have a `stubGetSentReferrals` function which I don't think relates to a a real function but mocks a response from the `/sent-referrals` endpoint. We call it in most of the tests which rely on the old `getReferralsSentBy/ToX` function, but I wonder if it'd make it clearer to rename it to this new function in those tests too?

@Gweaton done. i've also added a commit on top of this to deal with the draft referral endpoints too.

PR #329, opened 2021-05-28
open feedback survey link in new tab

tomsmyers:

open feedback survey link in new tab

stop users losing their work if they click the feedback link


sldblog (approved):
LGTM
Probably for @rwk-moj to review as well

PR #330, opened 2021-05-28
IC-1814: PP EOSR uses serviceCategoryIds instead of serviceCategory

andrewburrowsmoj:

Change the PP EOSR journey to use serviceCategoryIds instead of serviceCategoryId and desiredOutcomes rather than desiredOutcomeIds

Make the UI compatible with cohort referrals. Allows the PP user to see EOSR for cohort referrals.

### DIFF
--- a/integration_tests/integration/probationPractitionerReferrals.spec.js
+++ b/integration_tests/integration/probationPractitionerReferrals.spec.js
@@ -79,10 +79,12 @@ describe('Probation practitioner referrals dashboard', () => {
     cy.login()
 
     const serviceCategory = serviceCategoryFactory.build()
+    const intervention = interventionFactory.build({ serviceCategories: [serviceCategory] })
     const deliusServiceUser = deliusServiceUserFactory.build()
     const referral = sentReferralFactory.build({
       referral: {
-        serviceCategoryId: serviceCategory.id,
+        interventionId: intervention.id,
+        serviceCategoryIds: [serviceCategory.id],
         desiredOutcomesIds: [serviceCategory.desiredOutcomes[0].id],
         serviceUser: { crn: deliusServiceUser.otherIds.crn },
       },
@@ -103,6 +105,7 @@ describe('Probation practitioner referrals dashboard', () => {
     cy.stubGetEndOfServiceReport(endOfServiceReport.id, endOfServiceReport)
     cy.stubGetSentReferral(referral.id, referral)
     cy.stubGetServiceCategory(serviceCategory.id, serviceCategory)
+    cy.stubGetIntervention(intervention.id, intervention)
     cy.stubGetServiceUserByCRN(deliusServiceUser.otherIds.crn, deliusServiceUser)
 
     cy.visit(`/probation-practitioner/end-of-service-report/${endOfServiceReport.id}`)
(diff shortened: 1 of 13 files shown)


integration_tests/integration/serviceProviderReferrals.spec.js
```diff
@@ -842,7 +842,9 @@ describe('Service provider referrals dashboard', () => {
       referral: {
         interventionId: accommodationIntervention.id,
         serviceCategoryId: serviceCategory.id,
```
Gweaton, line 844: Can we remove `serviceCategoryId` now? Or is there still another section of the journey that relies on it?

integration_tests/integration/probationPractitionerReferrals.spec.js
```diff
@@ -77,45 +77,66 @@ describe('Probation practitioner referrals dashboard', () => {
...
+    })
+    const intervention = interventionFactory.build({ serviceCategories: [serviceCategory1, serviceCategory2] })
     const deliusServiceUser = deliusServiceUserFactory.build()
     const referral = sentReferralFactory.build({
       referral: {
         interventionId: intervention.id,
-        serviceCategoryIds: [serviceCategory.id],
-        desiredOutcomesIds: [serviceCategory.desiredOutcomes[0].id],
+        serviceCategoryIds: [serviceCategory1.id, serviceCategory2.id],
+        desiredOutcomesIds: [serviceCategory1.desiredOutcomes[0].id],
```
Gweaton, line 95: Can we remove `desiredOutcomesIds` here now?

integration_tests/integration/serviceProviderReferrals.spec.js
```diff
@@ -842,7 +842,9 @@ describe('Service provider referrals dashboard', () => {
...
+        serviceCategoryIds: [serviceCategory.id],
         desiredOutcomesIds: selectedDesiredOutcomesIds,
```
Gweaton, line 846: again, can we remove this? Or do the tests fail?

Gweaton (approved):
Looks good, just a couple of things we might be able to remove?

tomsmyers:
hey @andrewburrowsmoj @Gweaton are these questions still outstanding? would be great to remove those fields to get the contract tests passing again.

integration_tests/integration/serviceProviderReferrals.spec.js
andrewburrowsmoj: I'm modifying the PP EOSR in this commit, however because I had to touch on some shared libraries means I have to do a few small changes on the SP side - there is another ticket that is for SP EOSR changes.

Gweaton: ah sorry, that does make sense. All good then, we can remove these when you've done the SP side.

Gweaton:
> hey @andrewburrowsmoj @Gweaton are these questions still outstanding? would be great to remove those fields to get the contract tests passing again.

This is good to merge - we can't quite remove these fields yet, as Andrew said, that'll come next when we update the SP journey.

integration_tests/integration/probationPractitionerReferrals.spec.js
andrewburrowsmoj: thanks I've removed it here - as it's connected to PP

PR #338, opened 2021-05-28
Switch to local network access to the service

sldblog:

Switch to local network access to the service to avoid our requests being considered as malicious attempts

The Web Application Firewall is blocking legitimate UI-to-service requests

- We don't want to turn the WAF off
- We don't know how to configure the WAF or debug why it's blocking
- We do have the apps running in the same namespace

Same namespace means they are on the same isolated network, meaning the ui can directly call service with its service name as host; this eliminates the need to go out and back in via the ingress

However, it loses TLS, but that is most likely fine due to it being on an isolated network anyway


sldblog:
👍 from cybersecurity

> All the waf will do for you really is determine that a UI to API call is not structured correctly, and that check should happen before the UI really.  As long as the API is configured to only receive calls from that namespace then I can’t see a major risk here
>
> If you’re using JWTs then the WAF is probably doing less than that…

tomsmyers (approved):
the urls sure do look weird in the config - but i understand the logic here so all good for me

PR #346, opened 2021-06-01
Fix desired outcomes and complexity level selected values for cohort referrals

lawrence-forooghian:

Updates the desired outcomes and complexity level pages on the probation practitioner "make a referral" journey, to correctly display any already-selected value. This was previously only working for single-service referrals.

To bring the cohort referral journey in line with the single-service journey.


tomsmyers (approved):
nice 👍

PR #353, opened 2021-06-01
IC-1815: Remove references for old single referral fields on SP EOSR journey

andrewburrowsmoj:
## What does this pull request do?

Replaces use of serviceCategoryId with serviceCategoryIds, desiredOutcomeIds with desiredOutcomes and complexityLevelId with complexityLevels for SP EOSR journey.

## What is the intent behind these changes?

Provide base to use new cohort structure for SP EOSR journey

PR #354, opened 2021-06-01
Extract business logic to a class for reuse

lawrence-forooghian:

The piece of knowledge that a cohort intervention is one that has
multiple service categories should only exist in one place in the app.
It’s currently living in a presenter, but I need to use it in another
presenter.

If we had a "model" class called Intervention then we’d stick this as a
method in there, but we don’t - we just have a DTO interface that
contains the data returned by the interventions service API. I had
thought that perhaps we could create a model class that decorates the
DTO, allowing us to add further methods and properties. But there
doesn’t seem to be a simple way to do that in TypeScript.

So for now, I’ll just create a "decorator" class, which is really just a
utility class that takes a DTO and defines some properties derived from
it. "Decorator" is a bit of a misuse of the term; but the alternative
was another "Utils" type thing e.g. "InterventionUtils", which seemed a
bit vague. At least the naming I’ve gone for communicates what I was
_trying_ to do.

To make a stab at creating a pattern that stops us from scattering this kind of thing in places it shouldn't be.

### DIFF
--- a/server/decorators/interventionDecorator.test.ts
+++ b/server/decorators/interventionDecorator.test.ts
@@ -1,0 +1,23 @@
+import InterventionDecorator from './interventionDecorator'
+import interventionFactory from '../../testutils/factories/intervention'
+import serviceCategoryFactory from '../../testutils/factories/serviceCategory'
+
+describe(InterventionDecorator, () => {
+  describe('isCohortIntervention', () => {
+    it('returns false when the intervention has a single service category', () => {
+      const intervention = interventionFactory.build({ serviceCategories: [serviceCategoryFactory.build()] })
+      const decorator = new InterventionDecorator(intervention)
+
+      expect(decorator.isCohortIntervention).toEqual(false)
+    })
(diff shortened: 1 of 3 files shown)


andrewburrowsmoj (approved):
Nice. I wrote a function in my latest PR that extracts the service categories in an intervention which also match the selected service categories.

I wonder if that could somehow be put into this class - but it would definitely not be a decorator after that.

(https://github.com/ministryofjustice/hmpps-interventions-ui/pull/353/files)

Gweaton (approved):
Nice refactor, looks good

tomsmyers (approved):
nice - i like the idea. doesn't this kinda just amount to the same as writing a utils type package though?

lawrence-forooghian:
> nice - i like the idea. doesn't this kinda just amount to the same as writing a utils type package though?

Yeah, it really does. If the naming turns out to be unhelpful we can just stick them in utils.

PR #355, opened 2021-06-01
Ic 1731: Remove old single referral fields

andrewburrowsmoj:

Removes the `serviceCategoryId`, `desiredOutcomeIds` and `complexityLevelId` fields from contract tests and draft / sent referral objects

Ensure that we're not relying on the old referral fields.

Note: This is branched off IC-1815 which is still waiting for review

### DIFF
--- a/server/models/draftReferral.ts
+++ b/server/models/draftReferral.ts
@@ -10,13 +10,10 @@ export interface ReferralFields {
   completionDeadline: string
   serviceProvider: ServiceProvider
   interventionId: string
-  serviceCategoryId: string // deprecated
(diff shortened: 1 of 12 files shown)


andrewburrowsmoj:
Note to reviewer:
I've changed occurrences of
```
 if (referral.serviceCategoryId === null) {    
```
to
```
if (referral.serviceCategoryIds === null || referral.serviceCategoryIds.length &lt; 1) {
```

However this will all be unnecessary with my latest ticket (IC-1825) which will remove these checks entirely because it's getting the intervention.

Gweaton (approved):
Nice

PR #359, opened 2021-06-02
IC-1826: Ensure we're using Sentence case for Contract type names and service categories in Refer journey

Gweaton:

Ensures we're consistently using Sentence case for Contract type names and service categories (after discussion with content team) in Refer journey by using `utils.convertToProperCase`. e.g. Social inclusion, Women's services, Accommodation.

Apologies for the large commit - it was difficult to do this in smaller chunks do the integration tests, but it's essentially just wrapping names in that function everywhere and updating tests to expect this.

To ensure we're consistent with our casing throughout the Refer journey.

### DIFF
--- a/integration_tests/integration/make_a_referral/referralSectionVerifier.js
+++ b/integration_tests/integration/make_a_referral/referralSectionVerifier.js
@@ -23,7 +23,7 @@ class _ReferralSectionChecker {
 
   interventionReferralDetails(activeLinks) {
     cy.get('[data-cy=url]')
-      .contains('Confirm the relevant sentence for the accommodation referral')
+      .contains('Confirm the relevant sentence for the Accommodation referral')
       .should(hrefAttrChainer(activeLinks.relevantSentence), 'href')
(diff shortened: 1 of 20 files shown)


lawrence-forooghian (commented):
Unless we're treating "accommodation" etc as proper nouns, which I think wouldn't make much sense, then the change here is putting everything into title case, not sentence case. So I think that everything here should actually be lower-case, shouldn't it?

Also, I think that our logic for title-caseing things probably won't work for service types that contain minor words (like "and"), which shouldn't be capitalised. e.g. I think it would turn "education, training and employment" into "Education, Training And Employment".

lawrence-forooghian:
Oh, and I meant to say at the beginning: definitely a good idea to make things consistent, just not sure if what we've settled on is the right one.

lawrence-forooghian:
(Unless you've already checked this with content folks unless ignore me.)

lawrence-forooghian (approved):

PR #362, opened 2021-06-02
IC-1821: Update action plan to latest designs

Gweaton:

Updates Action Plan "Add activity" and "Review" pages to match the latest designs. Activities are no longer linked to a specific desired outcome, so we'll want to update the API to reflect these changes (to come in a future PR).

The journey should also work with cohort referrals, but we should test this on dev with realistic data.

This also removes page numbers from the action plan journey, as they're not in the designs.

As mentioned above, this doesn't touch the contract tests, but we'll want to update the `updateDraftActionPlan` function to not link an Activity to desired outcomes once this is merged.

To make the action plan journey work for both cohort and non-cohort referrals.

### Add activity page

image

### Review action plan page

image

### DIFF
--- a/assets/sass/components/_inset-text.scss
+++ b/assets/sass/components/_inset-text.scss
@@ -1,0 +1,4 @@
+.app-inset-text--grey {
+  background: govuk-colour('light-grey');
+  border-left: 10px solid govuk-colour('blue');
+}
--- a/server/routes/serviceProviderReferrals/addActionPlanActivitiesForm.test.ts
+++ b/server/routes/serviceProviderReferrals/addActionPlanActivitiesForm.test.ts
@@ -48,7 +48,7 @@ describe(AddActionPlanActivitiesForm, () => {
     })
   })
 
-  describe('errors', () => {
+  describe('error', () => {
     describe('when there is a non-empty text string for "description"', () => {
       it('returns an empty array', async () => {
         const form = await AddActionPlanActivitiesForm.createForm({
@@ -58,12 +58,12 @@ describe(AddActionPlanActivitiesForm, () => {
           },
         } as Request)
 
-        expect(form.errors).toEqual([])
+        expect(form.error).toEqual(null)
       })
     })
 
     describe('when there is an empty string for "description"', () => {
-      it('returns an array with an error for the relevant desired outcome ID', async () => {
+      it('returns an error', async () => {
         const form = await AddActionPlanActivitiesForm.createForm({
           body: {
             description: '',
@@ -71,20 +71,15 @@ describe(AddActionPlanActivitiesForm, () => {
           },
         } as Request)
 
-        expect(form.errors).toEqual([
-          {
-            desiredOutcomeId: '29843fdf-8b88-4b08-a0f9-dfbd3208fd2e',
-            error: {
-              errors: [
-                {
-                  errorSummaryLinkedField: 'description',
-                  formFields: ['description'],
-                  message: 'Enter an activity',
-                },
-              ],
+        expect(form.error).toEqual({
+          errors: [
+            {
+              errorSummaryLinkedField: 'description',
+              formFields: ['description'],
+              message: 'Enter an activity',
             },
-          },
-        ])
+          ],
+        })
       })
     })
   })
--- a/server/routes/serviceProviderReferrals/addActionPlanActivitiesPresenter.test.ts
+++ b/server/routes/serviceProviderReferrals/addActionPlanActivitiesPresenter.test.ts
@@ -46,7 +46,34 @@ describe(AddActionPlanActivitiesPresenter, () => {
 
   describe('text', () => {
     describe('title', () => {
-      it('includes the name of the service category', () => {
+      it('specifies the number of the activity to be added', () => {
+        const socialInclusionServiceCategory = serviceCategoryFactory.build({ name: 'social inclusion' })
+        const actionPlan = actionPlanFactory.oneActivityAdded().build()
+
+        const presenter = new AddActionPlanActivitiesPresenter(
+          sentReferral,
+          [socialInclusionServiceCategory],
+          actionPlan
+        )
+
+        expect(presenter.text.title).toEqual('Add activity 2 to action plan')
+      })
+    })
+
+    describe('referredOutcomesHeader', () => {
+      it('includes the name of the service user', () => {
+        const actionPlan = actionPlanFactory.justCreated(sentReferral.id).build()
+
+        const presenter = new AddActionPlanActivitiesPresenter(sentReferral, serviceCategories, actionPlan)
+
+        expect(presenter.text.referredOutcomesHeader).toEqual('Referred outcomes for Jenny')
+      })
+    })
+  })
+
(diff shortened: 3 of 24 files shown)


server/routes/serviceProviderReferrals/addActionPlanActivitiesPresenter.test.ts
```diff
@@ -46,16 +46,45 @@ describe(AddActionPlanActivitiesPresenter, () => {
...
+        })
+      })
 
-        expect(presenter.text.title).toEqual('Social inclusion - create action plan')
+      describe('when 1 activity has been added', () => {
+        it('specifies the number of the activity to be added', () => {
+          const socialInclusionServiceCategory = serviceCategoryFactory.build({ name: 'social inclusion' })
+          const actionPlan = actionPlanFactory.justCreated(sentReferral.id).build({
+            activities: [
+              {
```
lawrence-forooghian: This `Activity` object contains a lot of data that's not relevant for this test case, which is only concerned with the number of activities in the action plan.

It might be a good idea to create an `actionPlanActivityFactory`, so that we can create valid instances of `Activity` in a single line. We'd also want to update `actionPlanFactory` to call this new factory when creating activity instances.

server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
```diff
@@ -56,6 +55,7 @@ import EndOfServiceReportConfirmationView from './endOfServiceReportConfirmation
 import ControllerUtils from '../../utils/controllerUtils'
 import AuthUserDetails from '../../models/hmppsAuth/authUserDetails'
 import ServiceCategory from '../../models/serviceCategory'
+import FinaliseActionPlanActivitiesForm from './finaliseActionPlanActivitiesForm'
```
lawrence-forooghian, line 58: Why has this import moved?

server/routes/serviceProviderReferrals/addActionPlanActivitiesForm.test.ts
```diff
@@ -58,33 +58,28 @@ describe(AddActionPlanActivitiesForm, () => {
...
         } as Request)
 
-        expect(form.errors).toEqual([])
+        expect(form.error).toEqual(null)
       })
     })
 
     describe('when there is an empty string for "description"', () => {
-      it('returns an array with an error for the relevant desired outcome ID', async () => {
+      it('returns an error for the relevant desired outcome ID', async () => {
```
lawrence-forooghian, line 66: The "for the relevant desired outcome ID" no longer applies, right?

server/routes/serviceProviderReferrals/addActionPlanActivitiesPresenter.test.ts
```diff
@@ -46,205 +46,159 @@ describe(AddActionPlanActivitiesPresenter, () => {
...
 
-    describe('subTitle', () => {
+    describe('referredOutcomesHeader', () => {
       it('includes the name of the service user', () => {
         const actionPlan = actionPlanFactory.justCreated(sentReferral.id).build()
 
         const presenter = new AddActionPlanActivitiesPresenter(sentReferral, serviceCategories, actionPlan)
 
-        expect(presenter.text.subTitle).toEqual('Referred outcomes for Jenny')
+        expect(presenter.text.referredOutcomesHeader).toEqual('Referred outcomes for Jenny')
```
lawrence-forooghian, line 81: Similarly, could this be in the previous commit?

assets/sass/components/_inset-text.scss
```diff
@@ -0,0 +1,4 @@
+.govuk-inset-text--grey {
```
lawrence-forooghian, line 1: The [suggested naming for custom modifiers to GOV.UK Frontend components](https://design-system.service.gov.uk/get-started/extending-and-modifying-components/#small-modifications-to-components) I think would make this `.app-inset-text--grey`. I might have misunderstood the guidance though; does your reading of it match?

assets/sass/components/_review-action-plan.scss
```diff
@@ -0,0 +1,5 @@
+.review-action-plan--bordered {
```
lawrence-forooghian, line 1: A couple of thoughts on this:

- The `--` suffix is used for modifiers to an existing component; since we don't have a `review-action-plan` component to modify I don't think it makes sense. We should probably either call it `review-action-plan` or `review-action-plan-bordered`.
- The [suggested naming for custom components](https://design-system.service.gov.uk/get-started/extending-and-modifying-components/#use-a-unique-prefix-for-component-names) I think would suggest an `app-` prefix here.

I also wonder whether a [section break](https://design-system.service.gov.uk/styles/typography/#section-break) might be a more appropriate thing to use here? It _seems_ like we have a "thematic break between sections of content".

server/views/serviceProviderReferrals/endOfServiceReport/formTemplate.njk
```diff
@@ -13,10 +13,6 @@
 
       <h1 class="govuk-heading-xl">{{ presenter.formPagePresenter.text.title }}</h1>
 
-      {% if presenter.formPagePresenter.text.pageNumber %}
```
lawrence-forooghian, line 16: Was removing the page numbers from end of service report an accident?

lawrence-forooghian (commented):
Looks great, lots of work here! A few small comments.

server/views/serviceProviderReferrals/endOfServiceReport/formTemplate.njk
Gweaton: Yep, have fixed this.

assets/sass/components/_review-action-plan.scss
Gweaton: I knew "section break" was a thing, but I couldn't find it in the GOV components section, didn't realise it was in Typography - thanks for the info, I think that's exactly what we need - have replaced it (also solves the issue of css naming)

assets/sass/components/_inset-text.scss
Gweaton: I didn't know this, but I think you're right. Have changed this.

server/routes/serviceProviderReferrals/addActionPlanActivitiesPresenter.test.ts
Gweaton: I've lazily squashed those two commits together, it kind of makes sense to do them in one go.

server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
Gweaton: I removed it then put it back in and it auto-added here. Have fixed this up.

server/routes/serviceProviderReferrals/addActionPlanActivitiesPresenter.test.ts
Gweaton: Good spot, I've added an `activityFactory` for this and used it in those tests.

lawrence-forooghian (approved):

PR #369, opened 2021-06-03
IC-1625: Add new enforceable days page to Refer journey and remove old RAR days page

Gweaton:

- Switches `draftReferrals` to use new `maximumEnforceableDays`, rather than `maximumRarDays` field and removes that and `usingRarDays` fields.
- Adds new page for setting maximum enforceable days.
- Removes deprecated fields on the `draftReferral` object.
- Updates all points that expected RAR days field to use `maximumEnforceableDays`

This will replace the current `maximumRarDays` and `usingRarDays` fields, now that enforceable days is not an optional field, as PPs always have to specify a number of enforceable days.

## Screenshot

image


server/models/draftReferral.ts
```diff
@@ -23,8 +23,9 @@ export interface ReferralFields {
   whenUnavailable: string | null
   serviceUser: ServiceUser
   additionalRiskInformation: string
-  usingRarDays: boolean
-  maximumRarDays: number | null
+  usingRarDays: boolean // deprecated
+  maximumRarDays: number | null // deprecated
+  maximumEnforceableDays: number | null
```
lawrence-forooghian, line 28: I reckon this will no longer be nullable, now that the user has to enter a number of days?

Gweaton: Doh, I was getting confused with where it starts out as `null` but that's where we extend it with nullable fields. Will fix this, thanks.

lawrence-forooghian (approved):

Gweaton:
Marking as draft, as we've decided to do everything in this one PR.

tomsmyers (approved):
brilliant - thank you !

PR #378, opened 2021-06-04
IC-1854: remove hardcoded and unavailable risk info from referral details page

lawrence-forooghian:

Removes hardcoded risk info from the referral details page, and adds a message explaining that the risk info is not yet available.

To make sure we don't show inaccurate risk information, and to explain when risk information will become available.

## Screenshot

Screenshot 2021-06-04 at 11-43-10 HMPPS Interventions - GOV UK


tomsmyers (approved):

lawrence-forooghian:
Yikes, looks like I committed the wrong text (i.e. not that which is in the screenshots). I'll do another PR 🤦

PR #380, opened 2021-06-04
IC-1857: Hide intervention details until service categories are selected

andrewburrowsmoj:

Hides service category related details section for cohort referrals until user has selected them

image

Ensure that the page doesn't show incomplete sections

### DIFF
--- a/server/routes/referrals/referralFormPresenter.ts
+++ b/server/routes/referrals/referralFormPresenter.ts
@@ -338,15 +338,15 @@ class TaskValues {
     ]
   }
 
-  get relevantSentence(): DraftReferralValues {
-    return [this.referral.relevantSentenceId]
-  }
-
   get cohortServiceCategories(): DraftReferralValues {
-    if (this.referral.serviceCategoryIds === null) {
+    if (this.referral.serviceCategoryIds === null || this.referral.serviceCategoryIds.length === 0) {
       return [null]
     }
     return this.referral.serviceCategoryIds
+  }
+
+  get relevantSentence(): DraftReferralValues {
+    return [this.referral.relevantSentenceId]
   }
 
   get allComplexityLevels(): DraftReferralValues {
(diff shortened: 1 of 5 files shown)


server/routes/referrals/referralFormPresenter.ts
```diff
@@ -324,17 +338,17 @@ class TaskValues {
     ]
   }
 
-  get relevantSentence(): DraftReferralValues {
-    return [this.referral.relevantSentenceId]
-  }
-
   get cohortServiceCategories(): DraftReferralValues {
-    if (this.referral.serviceCategoryIds === null) {
+    if (this.referral.serviceCategoryIds === null || this.referral.serviceCategoryIds.length === 0) {
```
andrewburrowsmoj, line 342: even though we expect serviceCategoryIds to be null - the API appears to return an emptyList

sldblog: that sounds like something that'd be great in a contract

sldblog:
I'm a bit out of touch here -- could we have a before/after screenshot, if that makes sense, please?

sldblog (approved):
👍 LGTM, but I'm not the resident expert

server/routes/referrals/referralFormPresenter.ts
sldblog: great to have a test added for this 👏

andrewburrowsmoj:
Before:

image

After:
The screenshot in the main comment

server/routes/referrals/referralFormPresenter.ts
andrewburrowsmoj: I did add a test :-) it's in referralFormPresenter.test.ts
it tests if it's both null or empty

sldblog: yeah, sorry, I was happy that it was added :)

PR #385, opened 2021-06-04
Fixes CWE-99 and CWE-117 flagged by veracode

sldblog:

CWE-99: Improper Control of Resource Identifiers is flagged on `app.listen(app.get('port')...`
CWE-117: Improper Output Neutralization for Logs is flagged on ```log.info(`Server listening on port ${app.get('port')}`)```

The code switches to use `config.port`, which is then sanitised by converting the env variable to a number, defaulting to a sane value

Reduce medium CWEs open on the app.

To exploit these, someone would need to override the `PORT` variable on a production setting. Tbh if someone can do that, they can probably do much worse. 🤷

### DIFF
--- a/server/config.ts
+++ b/server/config.ts
@@ -37,7 +37,7 @@ export default {
   version: Date.now().toString(),
   production,
   testMode: process.env.NODE_ENV === 'test',
-  port: get('PORT', 3000),
+  port: Number(get('PORT', 3000)),
   https: production,
   staticResourceCacheDuration: 20,
   deploymentEnvironment: get('DEPLOYMENT_ENV', 'local', requiredInProduction),
--- a/server.ts
+++ b/server.ts
@@ -9,10 +9,11 @@ initialiseAppInsights()
 buildAppInsightsClient()
 
 import app from './server/index'
+import config from './server/config'
 import log from './log'
 
-app.listen(app.get('port'), async () => {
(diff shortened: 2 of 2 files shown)


server/config.ts
```diff
@@ -37,7 +37,7 @@ export default {
   version: Date.now().toString(),
   production,
   testMode: process.env.NODE_ENV === 'test',
-  port: get('PORT', 3000),
+  port: Number(process.env.PORT) || 3000,
```
tomsmyers, line 40: what's the reason for this change? the `get` function pulls the port from the environment in the same way, but with a bunch more checks and error handling

sldblog: hmm, the compiled javascript for `get` (for me) is
```
function get(name, fallback, options = { requireInProduction: false }) {
    const fromEnv = process.env[name];
    if (fromEnv !== undefined) {
        return fromEnv;
    }
```
which doesn't seem to explicitly sanitise the input -- was trying to ensure the string coming from the environment is explicitly treated as a number and any non-numeric values are thrown away

would `Number(get('PORT', 3000))` make sense in this context?

note: `Number(process.env.REDIS_PORT) || 6379,` is also a candidate to a similar change

sldblog: done in a269aedd7e5391afed2a01707b7fcecb7d37558b

tomsmyers (approved):

PR #391, opened 2021-06-07
IC-1832: Allow trailing and leading whitespace when entering CRN for …

andrewburrowsmoj:
…make a referral journey

Removes trailing and leading whitespace on CRN when being entered by user for make a referral journey.

image

Make user experience more enjoyable when copy and pasting CRNs


tomsmyers:
nice andrew! 

please can we also take this opportunity to upper case the CRN also. we have had it confirmed that all CRNs are upper case in delius.

cc @sldblog

sldblog:
yay user quality of life changes 🎉

tomsmyers (approved):

PR #392, opened 2021-06-07
Load instrumentation key on target environments

sldblog:

Load instrumentation key on target environments

`APPINSIGHTS_INSTRUMENTATIONKEY` was always unset on all environments -- this is now read from the same secret that the service uses

Switched to `APPLICATIONINSIGHTS_CONNECTION_STRING` as the service uses that same mechanism

This will enable tracing across all interventions apps

### DIFF
--- a/helm_deploy/hmpps-interventions-ui/templates/_envs.tpl
+++ b/helm_deploy/hmpps-interventions-ui/templates/_envs.tpl
@@ -55,6 +55,12 @@ env:
         name: sentry
         key: ui_dsn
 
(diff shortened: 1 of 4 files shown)


tomsmyers:
why not use the same variable - "connection string" - instead of instrumentation key in this deploy?

sldblog:
that's much better 👍 let me do that

sldblog:
@tomsmyers done in 5cb9d6f8e5eb7d26ea9a2142a902430a88d3f0e8

tomsmyers (approved):
nice - thanks!

sldblog:
(forgot the secrets example -- no other changes)

PR #396, opened 2021-06-07
IC-1859: Stop linking Activities to Desired Outcomes

Gweaton:

Removes `desiredOutcomes` field from an Action Plan `Activity`.

EDIT: change by @tomsmyers means that we still do pass the desired outcome id with the action plan activity request. but we no longer expect this in the response. doing this allows the API to update without causing any breaking contract changes. we will remove the unneeded extra request parameter from interventionsService.ts at a later date.

In the previous designs for the Action Plan activity form, we required
users to enter at least one activity per desired outcome, so it made
sense to link them.

After user feedback, and partly due to cohort referrals possibly having
many desired outcomes selected, we've made the decision to keep
activities separate from specific desired outcomes.

### DIFF
--- a/server/services/interventionsService.test.ts
+++ b/server/services/interventionsService.test.ts
@@ -1873,6 +1873,7 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
           body: {
             newActivity: {
               description: 'Attend training course',
+              desiredOutcomeId: '301ead30-30a4-4c7c-8296-2768abfb59b5',
             },
           },
         },
@@ -1899,6 +1900,7 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
       const draftActionPlan = await interventionsService.updateDraftActionPlan(token, draftActionPlanId, {
         newActivity: {
           description: 'Attend training course',
+          desiredOutcomeId: '301ead30-30a4-4c7c-8296-2768abfb59b5',
         },
       })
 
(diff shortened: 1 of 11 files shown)


server/views/serviceProviderReferrals/actionPlan/addActivities.njk
```diff
@@ -19,7 +19,6 @@
 
   <form method="post" action="{{ presenter.addActivityAction }}">
     <input type="hidden" name="_csrf" value="{{csrfToken}}">
-    <input type="hidden" name="desired-outcome-id" value="{{ presenter.firstDesiredOutcomeId }}">
```
lawrence-forooghian, line 22: Can we get rid of this property on the presenter?

lawrence-forooghian (approved):

server/views/serviceProviderReferrals/actionPlan/addActivities.njk
Gweaton: Oops forgot - thanks

server/services/interventionsService.test.ts
```diff
@@ -1886,11 +1875,6 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
               {
```
sldblog, line 1875: _nit_ this `activities` section is irrelevant anyway from the point of "PATCH request to set the number of sessions on the action plan" right? might be good to review later to cut off expectations that are not relevant

Gweaton: I suppose so, I guess I quite like having the whole response in the test so it's clear on our side exactly what we'll get, but maybe I should just be checking the swagger documentation for that?

tomsmyers (approved):

PR #397, opened 2021-06-07
IC-1856: Added report a problem page

andrewburrowsmoj:

Adds a new page that tells users how to report a problem.

Provide user with instructions on how to report a problem

Footer on all pages now looks like this:

image

When the user clicks on it as a PP they see this:

image

When the user clicks on it as an SP user they see this:

image


server/routes/common/reportAProblemPresenter.ts
```diff
@@ -0,0 +1,13 @@
+export default class ReportAProblemPresenter {
+  constructor(private readonly ppUser: boolean) {}
+
+  get contentHTML(): string {
+    if (this.ppUser) {
```
lawrence-forooghian, line 5: Creating HTML is a view responsibility, not a presenter one - I reckon expose the flag on the presenter and let the template do the conditional. (Ideally, the presenter would be able to expose the text, so that it could be tested in a unit test, but that's a bit faffy when you've got those embedded links in there and I'm not sure it's worth trying to solve. Looks like the integration tests have it covered anyway.)

server/routes/common/commonController.ts
```diff
@@ -0,0 +1,13 @@
+import { Request, Response } from 'express'
+import ReportAProblemPresenter from './reportAProblemPresenter'
+import ReportAProblemView from './reportAProblemView'
+import ControllerUtils from '../../utils/controllerUtils'
+
+export default class CommonController {
```
lawrence-forooghian, line 6: Can we have a unit test for this class, please?

andrewburrowsmoj: sure, I've added a unit test

server/routes/common/reportAProblemPresenter.ts
andrewburrowsmoj: Sure, I've moved the logic to the View. The commonController.test should cover the unit test scenario that you mention.

tomsmyers (approved):
lgtm

PR #398, opened 2021-06-07
IC-1861: Display real relevant sentence info on referral

Gweaton:

Displays real relevant sentence information on referral details for PP/SP.

This was previously "Not currently set", due to the fact that we hadn't
managed to fetch the relevant sentence from the Community API at that
point in development.

It looks like we forgot to update this once we did fetch that data.

To show the SP/PP the SU's sentence at any stage in the process.

## Screenshot

image

### DIFF
--- a/integration_tests/integration/probationPractitionerReferrals.spec.js
+++ b/integration_tests/integration/probationPractitionerReferrals.spec.js
@@ -4,6 +4,7 @@ import endOfServiceReportFactory from '../../testutils/factories/endOfServiceRep
 import deliusServiceUserFactory from '../../testutils/factories/deliusServiceUser'
 import interventionFactory from '../../testutils/factories/intervention'
 import deliusUserFactory from '../../testutils/factories/deliusUser'
+import deliusConvictionFactory from '../../testutils/factories/deliusConviction'
 
 describe('Probation practitioner referrals dashboard', () => {
   beforeEach(() => {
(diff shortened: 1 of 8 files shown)


integration_tests/integration/probationPractitionerReferrals.spec.js
```diff
@@ -169,12 +170,43 @@ describe('Probation practitioner referrals dashboard', () => {
...
+        },
+        {
+          mainOffence: false,
+          detail: {
+            mainCategoryDescription: 'Common and other types of assault',
+            subCategoryDescription: 'Common assault and battery',
+          },
+        },
+      ],
+      sentence: {
```
lawrence-forooghian, line 192: The only field on the sentence which affects this class's behaviour is the `expectedSentenceEndDate`, so can we use the `deliusSentence` factory to create this object and omit the other fields? Ditto for the other integration test.

server/routes/shared/showReferralPresenter.test.ts
```diff
@@ -20,6 +21,8 @@ describe(ShowReferralPresenter, () => {
     serviceCategories: cohortServiceCategories,
   })
 
+  const deliusConviction = deliusConvictionFactory.build()
```
lawrence-forooghian, line 24: We should be specific with the fields that affect the output of the class under test, and not rely on the factory's default values.

lawrence-forooghian (changes requested):
Looks good, a couple of comments.

Aside: I noticed whilst working on this page, having worked on "check your answers" recently, that we probably should have considered reusing the check your answers page as a referral details page, the same way we have for some other screens (forget which). Maybe one to bring up if we need further iterations on this page.

server/routes/shared/showReferralPresenter.test.ts
Gweaton: Good point - does this apply to the above too, or only in unit tests? You've only mentioned `expectedSentenceEndDate`, but we also display `category` and `subcategory`, so wanted to double check.

Gweaton: I think it should be in both, so I've added it to both and cut down the number of fields being explicitly set that we don't care about

lawrence-forooghian: Yep, applies to the above. But in the above I was just talking about the sentence (the category and subcategory are on the offence, not the sentence).

lawrence-forooghian (approved):

PR #399, opened 2021-06-07
Don't absorb appointment update errors

lawrence-forooghian:

Fixes a bug where non-409 appointment update response errors were being absorbed, causing the form to be re-rendered without any indication that something had gone wrong.

To make sure that users see an error message when something goes wrong.


server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
```diff
@@ -462,6 +462,8 @@ export default class ServiceProviderReferralsController {
                 },
               ],
             }
+          } else {
+            throw e
```
Gweaton, line 466: Oops, well spotted!

Gweaton (approved):

PR #400, opened 2021-06-07
IC-1871: Fix bug where desired outcomes link was inactive for cohort referrals after previous step filled in.

Gweaton:

Fixes a bug where the second desired outcomes selection link in the Refer journey wasn't accessible due to slightly incorrect logic. Before this change,  after filling out the referral details section, the "select desired outcomes" link was still inactive.

To make sure PPs can tell they're at the correct step in the journey, particularly after having filled out most of the referral form.

## Screenshot of bug

image

### DIFF
--- a/server/routes/referrals/referralFormPresenter.ts
+++ b/server/routes/referrals/referralFormPresenter.ts
@@ -208,7 +208,9 @@ class FormSectionBuilder {
                       `service-category/${serviceCat.id}/desired-outcomes`,
                       index === 0
                         ? this.taskValues.relevantSentence
-                        : this.taskValues.complexityLevel(this.intervention.serviceCategories[index - 1].id)
+                        : this.taskValues.complexityLevel(
+                            this.referral.serviceCategoryIds!.sort((a, b) => a.localeCompare(b))[index - 1]
+                          )
                     ),
                   },
                   {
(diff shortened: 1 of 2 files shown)


server/routes/referrals/referralFormPresenter.ts
```diff
@@ -208,7 +208,7 @@ class FormSectionBuilder {
                       `service-category/${serviceCat.id}/desired-outcomes`,
                       index === 0
                         ? this.taskValues.relevantSentence
-                        : this.taskValues.complexityLevel(this.intervention.serviceCategories[index - 1].id)
+                        : this.taskValues.complexityLevel(this.referral.serviceCategoryIds![index - 1])
```
andrewburrowsmoj, line 211: Although it works, perhaps we should also do a sort of this.referral.serviceCategoryIds?

Gweaton: I'm still a bit confused about the sorting logic 😕 would you mind explaining what that would do for us?

andrewburrowsmoj: it's only because the interventions is sorted based on the id.

the service category ids should already be sorted by the backend - but it provides additional protection if the contract with the back-end changes

Gweaton: Have pushed this change to sort them.

andrewburrowsmoj (approved):

PR #410, opened 2021-06-08
IC-1773: F/google analytics

johnnolan:

This PR adds Google Analytics to the site with a event delegation function to listen for clicks on the page and if there is an event data attribute on that element, it will collect the information and send to GA.

Enable basic tracking on the service with a way to add events to the markup.

I am going to put more comments in the PR on seperate lines to explain things further and drive some discussion/critique.

### DIFF
--- a/server/views/partials/layout.njk
+++ b/server/views/partials/layout.njk
@@ -44,7 +44,7 @@
     tag: {
       text: "feedback"
     },
-    html: '<a class="govuk-link" href="https://eu.surveymonkey.com/r/369L5GL">Give us your feedback</a> to help us improve this service.',
+    html: '<a class="govuk-link" data-ga-action="click" data-ga-category="Feedback" data-ga-label="Survey Monkey 369L5GL" data-ga-value="1" href="https://eu.surveymonkey.com/r/369L5GL">Give us your feedback</a> to help us improve this service.',
     classes: 'govuk-width-container'
   }) }}
 
--- a/server/views/partials/layout.njk
+++ b/server/views/partials/layout.njk
@@ -13,6 +13,11 @@
   <script src="/assets/js/html5shiv-3.7.3.min.js"></script>
   <![endif]-->
 
+  <script type="application/javascript">
+      window.gaConfig = {};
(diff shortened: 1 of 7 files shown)


server/views/partials/layout.njk
```diff
@@ -13,6 +13,11 @@
   <script src="/assets/js/html5shiv-3.7.3.min.js"></script>
   <![endif]-->
 
+  <script type="application/javascript">
+      window.gaConfig = {};
+      window.gaConfig.uaId = "{{ uaId }}";
```
johnnolan, line 18: As each environment will require its own UA- id, if possible, can this value be added at build time via a ENV in the CI build?

server/views/partials/layout.njk
```diff
@@ -39,7 +44,7 @@
     tag: {
       text: "feedback"
     },
-    html: '<a class="govuk-link" href="https://eu.surveymonkey.com/r/369L5GL">Give us your feedback</a> to help us improve this service.',
+    html: '<a class="govuk-link" data-ga-action="click" data-ga-category="Feedback" data-ga-label="Survey Monkey 369L5GL" data-ga-value="1" href="https://eu.surveymonkey.com/r/369L5GL">Give us your feedback</a> to help us improve this service.',
```
johnnolan, line 47: I have added the javascript event handler to look for these in any elements on the page and automatically send an event when clicked. Hopefully this helps cover a lot of use cases for now and makes it easy to add events as you need.

server/views/partials/layout.njk
johnnolan: yep that is correct and would work, exactly how we do it in our service here. Will take a go at adding that now, thanks

Gweaton (approved):
Thanks for this John - looks good to me. We tend to rebase on this project rather than use merge commits - I'm happy to pull, rebase and re-push this if you like?

johnnolan:
> Thanks for this John - looks good to me. We tend to rebase on this project rather than use merge commits - I'm happy to pull, rebase and re-push this if you like?

No problem and sorry, should have referenced your commits to take a look first, that would be ace thank you!

PR #411, opened 2021-06-09
increase the interventions-service webclient timeouts to allow long running community api requests on the backend to complete

tomsmyers:

increase the interventions-service webclient timeouts to allow long running community api requests on the backend to complete

stop referral send looking like it's failed on the front end when it's actually succeeded on the API

### DIFF
--- a/server/config.ts
+++ b/server/config.ts
@@ -77,9 +77,11 @@ export default {
     },
     interventionsService: {
       url: get('INTERVENTIONS_SERVICE_URL', 'http://localhost:9092', requiredInProduction),
+      // the interventions-service <-> community-api worst case timeout is 20s, this is intentionally just slightly higher
+      // fixme: this is too long
       timeout: {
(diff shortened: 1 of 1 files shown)


server/config.ts
```diff
@@ -77,9 +77,11 @@ export default {
     },
     interventionsService: {
       url: get('INTERVENTIONS_SERVICE_URL', 'http://localhost:9092', requiredInProduction),
+      // the interventions-service <-> community-api worst case timeout is 20s, this is intentionally just slightly higher
+      // fixme: this is too long
       timeout: {
-        response: 10000,
-        deadline: 10000,
+        response: 22000,
+        deadline: 22000,
```
sldblog, line 84: We can go up to `30s` I think as long as we put the end-to-end transaction on a KPI and set a target to bring it down

sldblog (approved):

PR #417, opened 2021-06-10
IC-1882: Changed initial find interventions page to use 'Service user' rather than 'Service User'

andrewburrowsmoj:

Change occurrences of 'Service User' to 'Service user'

Ensure that the user only sees 'Service user' as sentence case

Split into three parts:
3a23c1f: Changes to front end screen directly
be60a85: Change to factories
bd47688: Changes to unit tests

### DIFF
--- a/integration_tests/integration/probationPractitionerReferrals.spec.js
+++ b/integration_tests/integration/probationPractitionerReferrals.spec.js
@@ -266,15 +266,15 @@ describe('Probation practitioner referrals dashboard', () => {
 
     cy.contains('Accommodation service')
     cy.contains('LOW COMPLEXITY')
-    cy.contains('Service User has some capacity and means to secure')
-    cy.contains('All barriers, as identified in the Service User Action Plan')
-    cy.contains('Service User makes progress in obtaining accommodation')
+    cy.contains('Service user has some capacity and means to secure')
+    cy.contains('All barriers, as identified in the Service user action plan')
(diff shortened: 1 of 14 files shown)


andrewburrowsmoj:
There is an issue with github checks not updating to latest check when doing certain types of squashing... but as you can see from the individual commits the last commit had a passing circleCI run.

tomsmyers (approved):
thanks for this! 

looks like this requires changes in the reference data in the API. do we have to hold off on merging this until we have that change ready to go?

sldblog:
that wording is outside our control

sldblog:
the integration tests use a mock, right? can we change the mocks to not use the actual values we have to make it obvious it's operating on dummy data?

andrewburrowsmoj:
> thanks for this!
> 
> looks like this requires changes in the reference data in the API. do we have to hold off on merging this until we have that change ready to go?

Yes, that has been done in this PR: https://github.com/ministryofjustice/hmpps-interventions-service/pull/321

andrewburrowsmoj:
> the wording of the levels, outcomes is outside our control I think (we get it in a spreadsheet)

My understanding is that it is held in the database?

andrewburrowsmoj:
> the integration tests use a mock, right? can we change the mocks to not use the actual values we have to make it obvious it's operating on dummy data?

What do you mean by this?
You mean to use something like "Dummy Desired Outcome Description 1" "Dummy Desired Outcome Description 2"?

PR #418, opened 2021-06-10
IC-1828: Display supplementary risk information from ARN on Referral Details page

Gweaton:

Adds an Assess Risks and Needs Service to fetch supplementary risk information from the ARN API that was set in the Refer journey.

Refactors the `RestClient` to take an optional token, rather than requiring one on instantiation. This will make it possible to eventually refactor the other clients so we can write useful tests for them, e.g. we currently have no tests for the community API - see commit message in #b8ed580 for further information.

Displays risk information on the PP / SP referral details page.

image

To inform PP / SPs of Service User risks.

## Screenshot

### DIFF
--- a/mockApis/assessRisksAndNeedsService.ts
+++ b/mockApis/assessRisksAndNeedsService.ts
@@ -1,0 +1,21 @@
+import Wiremock from './wiremock'
+
+export default class AssessRisksAndNeedsServiceMocks {
+  constructor(private readonly wiremock: Wiremock) {}
+
+  stubGetSupplementaryRiskInformation = async (riskId: string, responseJson: unknown): Promise<unknown> => {
+    return this.wiremock.stubFor({
+      request: {
+        method: 'GET',
+        urlPattern: `/assess-risks-and-needs/risks/supplementary/${riskId}`,
+      },
+      response: {
+        status: 200,
+        headers: {
+          'Content-Type': 'application/json',
+        },
+        jsonBody: responseJson,
+      },
+    })
+  }
+}
--- a/server/config.ts
+++ b/server/config.ts
@@ -64,6 +64,14 @@ export default {
       },
       agent: new AgentConfig(),
     },
+    assessRisksAndNeedsApi: {
+      url: get('ASSESS_RISKS_AND_NEEDS_API_URL', 'http://localhost:8094', requiredInProduction),
+      timeout: {
+        response: Number(get('ASSESS_RISKS_AND_NEEDS_API_TIMEOUT_RESPONSE', 20000)),
+        deadline: Number(get('ASSESS_RISKS_AND_NEEDS_API_TIMEOUT_DEADLINE', 20000)),
+      },
+      agent: new AgentConfig(),
+    },
     hmppsAuth: {
       url: get('HMPPS_AUTH_URL', 'http://hmpps-auth:8090/auth', requiredInProduction),
       timeout: {
--- a/server/data/restClient.ts
+++ b/server/data/restClient.ts
@@ -1,6 +1,5 @@
 import superagent from 'superagent'
 import Agent, { HttpsAgent } from 'agentkeepalive'
-import { Readable } from 'stream'
 
 import Logger from 'bunyan'
 import sanitiseError from '../sanitisedError'
@@ -14,6 +13,7 @@ interface GetRequest {
   headers?: Record<string, string>
   responseType?: string
   raw?: boolean
+  token?: string | null
 }
 
 interface PostRequest {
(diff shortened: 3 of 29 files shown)


server/routes/probationPractitionerReferrals/probationPractitionerReferralsController.ts
```diff
@@ -106,6 +108,7 @@ export default class ProbationPractitionerReferralsController {
         sentReferral.referral.serviceUser.crn,
         sentReferral.referral.relevantSentenceId
       ),
+      this.assessRisksAndNeedsService.getSupplementaryRiskInformation(sentReferral.supplementaryRiskId),
```
Gweaton, line 111: *Question*: the Referral details page currently fails to load if the ARN api returns a 404 for the requested Risk ID. This should only happen if there's no Risk found in the ARN api, but it should be set in the Refer journey. Is it worth adding a `catch` statement and rendering "risk not found" in this scenario?

mockApis/assessRisksAndNeedsService.ts
```diff
@@ -0,0 +1,21 @@
+import Wiremock from './wiremock'
+
+export default class AssessRisksAndNeedsServiceMocks {
+  constructor(private readonly wiremock: Wiremock) {}
+
+  stubGetSupplementaryRiskInformation = async (riskId: string, responseJson: unknown): Promise<unknown> => {
+    return this.wiremock.stubFor({
+      request: {
+        method: 'GET',
+        urlPattern: `/assess-risks-and-needs/risks/supplementary/${riskId}`,
```
sldblog, line 10: _nit_ can be a new contract ;)

server/config.ts
```diff
@@ -63,6 +63,14 @@ export default {
       },
       agent: new AgentConfig(),
     },
+    assessRisksAndNeedsApi: {
+      url: get('ASSESS_RISKS_AND_NEEDS_API_URL', 'http://localhost:8094', requiredInProduction),
+      timeout: {
+        response: Number(get('ASSESS_RISKS_AND_NEEDS_API_TIMEOUT_RESPONSE', 10000)),
+        deadline: Number(get('ASSESS_RISKS_AND_NEEDS_API_TIMEOUT_DEADLINE', 10000)),
```
sldblog, line 70: given we saw a 8-9 second referral post to ARN, can we increase this to 20s, please?

server/routes/probationPractitionerReferrals/probationPractitionerReferralsController.ts
sldblog: it would be valuable I think, yes -- it may fail with other reasons too (network death, 503, actual bug that returns 404 on everything) and it'd be nice to display something on the screen next to the field about this

mockApis/assessRisksAndNeedsService.ts
Gweaton: A pact contract?

server/routes/probationPractitionerReferrals/probationPractitionerReferralsController.ts
Gweaton: I'm wondering if it's worth doing this as a separate PR just so we at least have this in for Monday next week? Will pick it up very soon but don't want to block this one!

Gweaton: We have a few places where we need to handle errors a bit more gracefully, so it could be done as part of that work.

server/models/assessRisksAndNeeds/supplementaryRiskInformation.ts
```diff
@@ -0,0 +1,10 @@
+export interface SupplementaryRiskInformation {
```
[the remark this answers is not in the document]
Gweaton: No, you're right - I can remove the other unused fields

server/data/restClient.ts
```diff
@@ -35,7 +36,7 @@ export default class RestClient {
 
   logger: Logger
 
-  constructor(private readonly name: string, private readonly config: ApiConfig, private readonly token: string) {
+  constructor(private readonly name: string, private readonly config: ApiConfig, private readonly token = '') {
```
tomsmyers, line 39: does this mean that if the caller of this doesn't specify token we explicitly send `Authorization: Bearer` with no token content?

would it be better to default to null and not set the auth header full stop?

tomsmyers (approved):
one comment about default token values in rest client 

nice work!

server/data/restClient.ts
Gweaton: That's a good point, I guess the caller could not pass in the token on initialization or when calling `get` so we'd sent that - I could either throw an error on an empty string or throw one if it's null  - what do you think would be better?

PR #421, opened 2021-06-11
Change the appointment update request DTO to ensure time and duration are always sent

tomsmyers:

Change the appointment update request DTO to ensure time and duration are always sent

fit new appointment model changes in the API.

### DIFF
--- a/server/routes/serviceProviderReferrals/editSessionForm.ts
+++ b/server/routes/serviceProviderReferrals/editSessionForm.ts
@@ -8,7 +8,7 @@ import errorMessages from '../../utils/errorMessages'
 export default class EditSessionForm {
   constructor(private readonly request: Request) {}
 
-  async data(): Promise<FormData<Partial<ActionPlanAppointmentUpdate>>> {
+  async data(): Promise<FormData<ActionPlanAppointmentUpdate>> {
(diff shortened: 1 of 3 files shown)


server/routes/serviceProviderReferrals/editSessionForm.ts
```diff
@@ -24,8 +24,8 @@ export default class EditSessionForm {
     return {
       error: null,
       paramsForUpdate: {
-        appointmentTime: dateResult.value.toISOString() ?? null,
-        durationInMinutes: durationResult.value.minutes ?? null,
+        appointmentTime: dateResult.value.toISOString(),
+        durationInMinutes: durationResult.value.minutes!,
```
tomsmyers, line 28: is this ok to do? does the validation step check that this will not be null?

Gweaton: The UI prevents form submission if any of those fields are empty, so it _should_ be safe

sldblog: there's a test with `'duration-minutes': '',` input and it seems to pass 🤷‍♂️

sldblog (approved):
LGTM from my perspective

PR #422, opened 2021-06-11
IC-1773: Pass in GA_ID from configuration

sldblog:

Fills in the missing link between configuration and layout locals to make sure GA tracking ID works.

I renamed `uaId` to `googleAnalyticsTrackingId` as it was weird passing it around out of context

Bugfix:

**Actual behaviour**: Currently, the tracking ID is empty on all environments.

**Expected behaviour**: Tracking ID is not empty

### DIFF
--- a/server/config.ts
+++ b/server/config.ts
@@ -41,7 +41,7 @@ export default {
   https: production,
   staticResourceCacheDuration: 20,
   deploymentEnvironment: get('DEPLOYMENT_ENV', 'local', requiredInProduction),
-  uaId: get('GA_ID', '', requiredInProduction),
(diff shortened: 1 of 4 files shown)


sldblog (commented):
Things I weren't sure about:

- Is this approach of getting the config and passing it down as view locals ok?
- Is the "set the config" approach to "mock" it in `server/utils/controllerUtils.test.ts` ok?

Gweaton (approved):
I think this approach is fine, thanks for fixing.

sldblog:
♻️ rebased to resolve conflicts

sldblog:
.. another conflcit

PR #423, opened 2021-06-11
WIP: Show friendly error to users when no convictions are found in Delius for the user being referred

Gweaton:

Renders an error on the "Select the relevant sentence" screen when no convictions are found for a user in Delius, rather than using the generic server error page we usually display when something goes wrong.

To stop the user journey from effectively breaking, and to signpost them back to where they can keep using the service.

## Note

This is waiting on feedback from the content / design team, but I've opened this as a PR to get feedback on the approach.

## Screenshot

image


Gweaton:
Closing this in favour of the new structure introduced in #427

PR #429, opened 2021-06-14
IC-1899: Display user friendly message when no convictions

Gweaton:

Displays 'No convictions found in nDelius for [CRN]' to a user when trying to select a relevant sentence for a service user without any active convictions.

To surface a more useful error to inform the user that something's not right with the Service User record they're trying to refer, so they know to contact us.

## Screenshot

image

### DIFF
--- a/server/routes/referrals/referralsController.ts
+++ b/server/routes/referrals/referralsController.ts
@@ -1,4 +1,5 @@
 import { Request, Response } from 'express'
+import createError from 'http-errors'
 import InterventionsService from '../../services/interventionsService'
 import { FormValidationError } from '../../utils/formValidationError'
 import createFormValidationErrorOrRethrow from '../../utils/interventionsFormError'
@@ -167,7 +168,9 @@ export default class ReferralsController {
     ])
 
     if (convictions.length < 1) {
-      throw new Error(`No active convictions found for service user ${referral.serviceUser.crn}`)
+      throw createError(500, `No active convictions found for service user ${referral.serviceUser.crn}`, {
+        userMessage: `No convictions were found in nDelius for ${referral.serviceUser.crn}.`,
+      })
     }
 
     const presenter = new RelevantSentencePresenter(referral, intervention, convictions)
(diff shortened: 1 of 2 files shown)


server/routes/referrals/referralsController.ts
```diff
@@ -167,7 +168,9 @@ export default class ReferralsController {
     ])
 
     if (convictions.length < 1) {
-      throw new Error(`No active convictions found for service user ${referral.serviceUser.crn}`)
+      throw createError(404, `No active convictions found for service user ${referral.serviceUser.crn}`, {
```
tomsmyers, line 171: i'm not convinced 404 is the correct response here. for requests in the browser 404 is typically reserved for invalid urls vs missing resources in rest apis

tomsmyers: i would go with a common and garden 500

tomsmyers:
just a question on the status code. 

nice that this fits neatly with the new error handler!

server/routes/referrals/referralsController.ts
sldblog: it's a client-input based problem (this person shouldn't be referred here), so 5xx does not seem appropriate

perhaps 400 Bad Request?

sldblog: 409 Conflict's description sounds best though

> The 409 (Conflict) status code indicates that the request could not
> be completed due to a conflict with the current state of the target
> resource.  This code is used in situations where the user might 
> able to resolve the conflict and resubmit the request

tomsmyers: i'm not sure this is a client error. there's nothing the client can do to fix this. it's an invalid referral.

mozilla says this about 500s "the server encountered an unexpected condition that prevented it from fulfilling the request", which sounds about right to me.

i don't think this is 400 or 409. 400s should be returned when the syntax or content of the request cannot be understood (e.g. malformed UUID). 409 is more like "you tried to create an supplier assessment for this referral but one already exists - try creating one for a different referral". 

in this case i'm not sure the client has anything to do with it. 

all this is kinda moot though, because we need to do this check on referral creation really soon. i can imagine this will happen a lot otherwise. also i could have probably just made a better use of my time than nitpicking response codes for temporary code paths 😛

tomsmyers (approved):

server/routes/referrals/referralsController.ts
Gweaton: Ok, I've gone for 500 as a temporary step so we can at least get this in if it's going to be replaced soon.

PR #431, opened 2021-06-15
IC-1673 refactor action plan presenter code so it can be reused

tomsmyers:

refactor action plan presenter code so it can be reused

allow more pages with action plan content to be built without copy pasting this code.

### DIFF
--- a/server/routes/shared/actionPlanDetailsPresenter.ts
+++ b/server/routes/shared/actionPlanDetailsPresenter.ts
@@ -1,0 +1,45 @@
+import ActionPlan from '../../models/actionPlan'
+import DateUtils from '../../utils/dateUtils'
+
+export default class ActionPlanDetailsPresenter {
+  constructor(
+    private readonly actionPlan: ActionPlan | null,
+    private readonly subNavUrlPrefix: 'service-provider' | 'probation-practitioner'
+  ) {}
+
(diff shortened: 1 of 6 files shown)


server/routes/shared/actionPlanDetailsPresenter.ts
```diff
@@ -0,0 +1,45 @@
...
+    }
+    return 'Not submitted'
+  }
+
+  get actionPlanCreated(): boolean {
+    return this.actionPlan !== null
+  }
+
+  get actionPlanUnderReview(): boolean {
+    return this.actionPlan?.submittedAt != null && !this.actionPlanApproved
```
Gweaton, line 39: Is this intentionally `!=` so it can check the `undefined`? If so, I think `return this.actionPlan?.submittedAt ...` is a bit easier to read.

server/routes/shared/actionPlanDetailsPresenter.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  get actionPlanApproved(): boolean {
+    return this.actionPlan?.approvedAt != null
```
Gweaton, line 43: same as above

server/routes/shared/actionPlanDetailsPresenter.ts
Gweaton: Just seen it's the same as before you moved it, but might be worth updating now?

Gweaton (approved):
Looks good, just one thing re `!=`

PR #432, opened 2021-06-15
IC-1773: Fix Google Analytics setup [take 3]

Gweaton:

Allows inline GA-variable-setting script to execute with our current CSP.

This allows us to set the GA `uaId` variable in views with a script, as we
attempted to do before in f10c703 - it was blocked due to our CSP
forbidding inline `<script>` execution. Supplying a whitelisted unique
nonce attribute to the script tag permits this execution.

See
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src
for more information.

To fix our GA setup 🤞 and track user behaviour.

### DIFF
--- a/server/app.ts
+++ b/server/app.ts
@@ -12,6 +12,7 @@ import flash from 'connect-flash'
 import redis from 'redis'
 import session from 'express-session'
 import connectRedis from 'connect-redis'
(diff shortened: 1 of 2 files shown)


tomsmyers (approved):
lgtm

PR #436, opened 2021-06-15
IC-1888: Provide front-end functionality to specify location details for an appointment

andrewburrowsmoj:

Provides an input form for the user to specify how an appointment will be delivered (phone call, video call, address) - and if they've chosen an address - where an appointment will take place.

Provide user with ability to edit and view address details of appointments.

The edit page looks like this:

image
image
image

If the user does not select a method:

image
image
If the user does not provide an address:

image

image
image

### DIFF
--- a/server/models/actionPlan.ts
+++ b/server/models/actionPlan.ts
@@ -6,16 +6,31 @@ export interface Activity {
   createdAt: string
 }
 
+export interface Address {
+  firstAddressLine: string
+  secondAddressLine: string | null
+  townOrCity: string
+  county: string
+  postCode: string
+}
 export interface ActionPlanAppointment {
   sessionNumber: number
   appointmentTime: string | null
   durationInMinutes: number | null
+  appointmentDeliveryType: AppointmentDeliveryType | null
+  appointmentDeliveryAddress: Address | null
   sessionFeedback: {
     attendance: AppointmentAttendance
     behaviour: AppointmentBehaviour
     submitted: boolean
   }
 }
+
+export type AppointmentDeliveryType =
+  | 'PHONE_CALL'
+  | 'VIDEO_CALL'
+  | 'IN_PERSON_MEETING_PROBATION_OFFICE'
+  | 'IN_PERSON_MEETING_OTHER'
 
 export type Attended = 'yes' | 'no' | 'late' | null
 
--- a/server/routes/serviceProviderReferrals/editSessionForm.ts
+++ b/server/routes/serviceProviderReferrals/editSessionForm.ts
@@ -16,14 +16,18 @@ export default class EditSessionForm {
     const [dateResult, durationResult, appointmentDeliveryType] = await Promise.all([
       new TwelveHourBritishDateTimeInput(this.request, 'date', 'time', errorMessages.editSession.time).validate(),
       new DurationInput(this.request, 'duration', errorMessages.editSession.duration).validate(),
-      new MeetingMethodInput(this.request, 'meeting-method').validate(),
+      new MeetingMethodInput(this.request, 'meeting-method', errorMessages.editSession.meetingMethod).validate(),
     ])
     let appointmentDeliveryAddress: FormValidationResult<Address | null> = { value: null, error: null }
     if (appointmentDeliveryType.value === 'IN_PERSON_MEETING_OTHER') {
-      appointmentDeliveryAddress = await new AddressInput(this.request, 'method-other-location').validate()
+      appointmentDeliveryAddress = await new AddressInput(
+        this.request,
+        'method-other-location',
+        errorMessages.editSession.address
+      ).validate()
     }
 
-    if (dateResult.error || durationResult.error || appointmentDeliveryType.error) {
+    if (dateResult.error || durationResult.error || appointmentDeliveryType.error || appointmentDeliveryAddress.error) {
       return {
         paramsForUpdate: null,
         error: {
@@ -31,6 +35,7 @@ export default class EditSessionForm {
             ...(dateResult.error?.errors ?? []),
             ...(durationResult.error?.errors ?? []),
             ...(appointmentDeliveryType.error?.errors ?? []),
+            ...(appointmentDeliveryAddress.error?.errors ?? []),
           ],
         },
       }
--- a/server/routes/serviceProviderReferrals/editSessionForm.ts
+++ b/server/routes/serviceProviderReferrals/editSessionForm.ts
@@ -4,20 +4,35 @@ import TwelveHourBritishDateTimeInput from '../../utils/forms/inputs/twelveHourB
 import { FormData } from '../../utils/forms/formData'
 import DurationInput from '../../utils/forms/inputs/durationInput'
 import errorMessages from '../../utils/errorMessages'
+import MeetingMethodInput from '../../utils/forms/inputs/meetingMethodInput'
+import AddressInput from '../../utils/forms/inputs/addressInput'
+import { FormValidationResult } from '../../utils/forms/formValidationResult'
+import { Address } from '../../models/actionPlan'
 
 export default class EditSessionForm {
   constructor(private readonly request: Request) {}
 
   async data(): Promise<FormData<ActionPlanAppointmentUpdate>> {
-    const [dateResult, durationResult] = await Promise.all([
+    const [dateResult, durationResult, appointmentDeliveryType] = await Promise.all([
       new TwelveHourBritishDateTimeInput(this.request, 'date', 'time', errorMessages.editSession.time).validate(),
       new DurationInput(this.request, 'duration', errorMessages.editSession.duration).validate(),
+      new MeetingMethodInput(this.request, 'meeting-method').validate(),
     ])
+    let appointmentDeliveryAddress: FormValidationResult<Address | null> = { value: null, error: null }
+    if (appointmentDeliveryType.value === 'IN_PERSON_MEETING_OTHER') {
+      appointmentDeliveryAddress = await new AddressInput(this.request, 'method-other-location').validate()
+    }
 
-    if (dateResult.error || durationResult.error) {
+    if (dateResult.error || durationResult.error || appointmentDeliveryType.error) {
       return {
         paramsForUpdate: null,
-        error: { errors: [...(dateResult.error?.errors ?? []), ...(durationResult.error?.errors ?? [])] },
+        error: {
+          errors: [
+            ...(dateResult.error?.errors ?? []),
+            ...(durationResult.error?.errors ?? []),
+            ...(appointmentDeliveryType.error?.errors ?? []),
+          ],
+        },
       }
     }
 
@@ -26,6 +41,8 @@ export default class EditSessionForm {
       paramsForUpdate: {
         appointmentTime: dateResult.value.toISOString(),
         durationInMinutes: durationResult.value.minutes!,
+        appointmentDeliveryType: appointmentDeliveryType.value!,
+        appointmentDeliveryAddress: appointmentDeliveryAddress.value,
       },
     }
   }
--- a/server/routes/serviceProviderReferrals/editSessionPresenter.test.ts
+++ b/server/routes/serviceProviderReferrals/editSessionPresenter.test.ts
@@ -108,12 +108,8 @@ describe(EditSessionPresenter, () => {
             hours: { value: '', hasError: false },
             minutes: { value: '', hasError: false },
           },
-          meetingMethod: null,
-          addressLine1: null,
-          addressLine2: null,
-          addressTownOrCity: null,
-          addressCounty: null,
-          addressPostCode: null,
+          meetingMethod: { value: null, errorMessage: null },
+          address: null,
         })
       })
     })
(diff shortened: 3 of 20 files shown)


server/routes/serviceProviderReferrals/editSessionPresenter.ts
```diff
@@ -45,5 +45,39 @@ export default class EditSessionPresenter {
...
+    meetingMethod: this.chosenAppointmentDeliveryType(this.appointment.appointmentDeliveryType),
+    addressLine1: this.getAddressFromIndex(0, this.appointment.appointmentDeliveryAddress),
+    addressLine2: this.getAddressFromIndex(1, this.appointment.appointmentDeliveryAddress),
+    addressTownOrCity: this.getAddressFromIndex(2, this.appointment.appointmentDeliveryAddress),
+    addressCounty: this.getAddressFromIndex(3, this.appointment.appointmentDeliveryAddress),
+    addressPostCode: this.getAddressFromIndex(4, this.appointment.appointmentDeliveryAddress),
+  }
+
+  private getAddressFromIndex(index: number, address: string[] | null): string | null {
+    if (address != null) {
```
Gweaton, line 57: should this be `!== null`?

server/routes/serviceProviderReferrals/editSessionPresenter.ts
```diff
@@ -45,5 +45,39 @@ export default class EditSessionPresenter {
...
     ),
+    meetingMethod: this.chosenAppointmentDeliveryType(this.appointment.appointmentDeliveryType),
+    addressLine1: this.getAddressFromIndex(0, this.appointment.appointmentDeliveryAddress),
+    addressLine2: this.getAddressFromIndex(1, this.appointment.appointmentDeliveryAddress),
+    addressTownOrCity: this.getAddressFromIndex(2, this.appointment.appointmentDeliveryAddress),
+    addressCounty: this.getAddressFromIndex(3, this.appointment.appointmentDeliveryAddress),
+    addressPostCode: this.getAddressFromIndex(4, this.appointment.appointmentDeliveryAddress),
+  }
+
+  private getAddressFromIndex(index: number, address: string[] | null): string | null {
```
Gweaton, line 56: Would array destructuring work here instead of this function, e.g. `const [addressLine1, addressLine2, addressTownOrCity, addressCounty, addressPostcode] = this.appintment.appointmentDeliveryAddress`? Might make things a little neater, as it's not that obvious from the indexes what each value represents.

You could then possibly return these values from a function and use them above e.g.

```
private getAddress(address: string[] | null): Address {
  const [addressLine1, addressLine2, addressTownOrCity, addressCounty, addressPostcode] = this.appointment.appointmentDeliveryAddress`

  return { addressLine1, addressLine2, addressTownOrCity, addressCounty, addressPostcode }
}
```

server/utils/forms/inputs/meetingMethodInput.ts
```diff
@@ -0,0 +1,28 @@
+import { Request } from 'express'
+import { FormValidationResult } from '../formValidationResult'
+import { AppointmentDeliveryType } from '../../../models/actionPlan'
+
+export default class MeetingMethodInput {
```
Gweaton, line 5: Would be good to add a test for this class (unless I've missed something!)

Gweaton (commented):
Nice work - just a couple of little things I've spotted, let me know if they don't make sense.

server/routes/serviceProviderReferrals/editSessionPresenter.ts
andrewburrowsmoj: Sorry this was a draft PR. It has been corrected in the latest commits.

server/utils/forms/inputs/meetingMethodInput.ts
andrewburrowsmoj: Yes, I've added this now (Sorry it was in a draft PR accidentally)

server/routes/serviceProviderReferrals/editSessionPresenter.ts
andrewburrowsmoj: Not required anymore (was a draft PR)

Gweaton (commented):
Need to take a bit more of a detailed look but here are things I've spotted so far

server/utils/forms/inputs/addressInput.ts
```diff
@@ -15,19 +21,44 @@ export default class AddressInput {
...
-      this.request.body[this.keys.postCode],
+      ExpressValidator.body(this.keys.addressLine1)
+        .notEmpty({ ignore_whitespace: true })
+        .withMessage('Enter a Building and street'),
+
+      ExpressValidator.body(this.keys.townOrCity)
+        .notEmpty({ ignore_whitespace: true })
+        .withMessage('Enter a town or city'),
+
+      ExpressValidator.body(this.keys.county).notEmpty({ ignore_whitespace: true }).withMessage('Enter a county'),
```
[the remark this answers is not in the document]
andrewburrowsmoj: Thanks. I've done changes to include this.

server/utils/forms/inputs/addressInput.ts
```diff
@@ -15,19 +21,44 @@ export default class AddressInput {
...
+  private get validations(): ExpressValidator.ValidationChain[] {
     return [
-      this.request.body[this.keys.addressLine1],
-      this.request.body[this.keys.addressLine2] ?? '',
-      this.request.body[this.keys.townOrCity],
-      this.request.body[this.keys.county],
-      this.request.body[this.keys.postCode],
+      ExpressValidator.body(this.keys.addressLine1)
+        .notEmpty({ ignore_whitespace: true })
+        .withMessage('Enter a Building and street'),
```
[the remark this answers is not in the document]
andrewburrowsmoj: Thanks I've changed it to lower.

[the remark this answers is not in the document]
andrewburrowsmoj: It is now "Enter a value for address line 1" based on feedback from standup demo and discussion with Ming

server/models/actionPlan.ts
```diff
@@ -12,13 +12,21 @@ export interface ActionPlanAppointment {
   sessionNumber: number
   appointmentTime: string | null
   durationInMinutes: number | null
+  appointmentDeliveryType: AppointmentDeliveryType | null
```
tomsmyers, line 15: isn't this non null on the backend model?

server/routes/serviceProviderReferrals/editSessionForm.ts
```diff
@@ -26,6 +45,8 @@ export default class EditSessionForm {
       paramsForUpdate: {
         appointmentTime: dateResult.value.toISOString(),
         durationInMinutes: durationResult.value.minutes!,
+        appointmentDeliveryType: appointmentDeliveryType.value!,
```
tomsmyers, line 48: does the validator handle the case where this is null of undefined?

server/routes/serviceProviderReferrals/editSessionPresenter.test.ts
```diff
@@ -15,29 +16,16 @@ describe(EditSessionPresenter, () => {
     describe('when a server error is passed in', () => {
       it('displays the message from the server error', () => {
         const appointment = actionPlanAppointmentFactory.build({ sessionNumber: 1 })
-        const presenter = new EditSessionPresenter(
-          appointment,
-          {
-            errors: [
-              {
-                errorSummaryLinkedField: 'date-day',
```
tomsmyers, line 23: what happened to this error?

server/routes/serviceProviderReferrals/editSessionView.ts
```diff
@@ -134,4 +140,104 @@ export default class EditSessionView {
       ],
     }
   }
+
+  private get otherLocationFirstAddressLineInputArgs(): InputArgs {
```
tomsmyers, line 144: would be good to have these in a reusable location so we can use for supplier assessments

server/routes/serviceProviderReferrals/editSessionView.ts
```diff
@@ -134,4 +140,104 @@ export default class EditSessionView {
...
+          text: 'Phone call',
+          checked: this.presenter.fields.meetingMethod.value === 'PHONE_CALL',
+        },
+        {
+          value: 'VIDEO_CALL',
+          text: 'Video call',
+          checked: this.presenter.fields.meetingMethod.value === 'VIDEO_CALL',
+        },
+        {
+          value: 'IN_PERSON_MEETING_OTHER',
```
tomsmyers, line 233: i'm assuming we are holding off on allowing NPS office locations until we figure all that out?

server/utils/address.ts
```diff
@@ -0,0 +1,29 @@
...
+  ) {}
+
+  static fromArray(address: string[] | null): Address | null {
+    if (address === null) {
+      return null
+    }
+    if (address.length !== 5) {
+      return null
+    }
+    return new Address(address[0], address[1], address[2], address[3], address[4])
```
tomsmyers, line 17: this slightly worries me. i do actually like that the main portion of the address is unstructured, but then unpacking an array into a structured data class like this seems error prone. perhaps i missed the discussions on this, but doesn't this tightly couple the UI implementation to the data model?

tomsmyers: i've just continued to read below and i see that the address form _is_ structured, so perhaps i'm just not grepping the full flow end to end. but it all looks fine where the address form is parsed.

server/utils/forms/inputs/addressInput.ts
```diff
@@ -0,0 +1,80 @@
...
+        .withMessage(this.errorMessages.townOrCityEmpty),
+
+      ExpressValidator.body(this.keys.county)
+        .notEmpty({ ignore_whitespace: true })
+        .withMessage(this.errorMessages.countyEmpty),
+
+      ExpressValidator.body(this.keys.postCode)
+        .notEmpty({ ignore_whitespace: true })
+        .withMessage(this.errorMessages.postCodeEmpty),
+      ExpressValidator.body(this.keys.postCode).isLength({ max: 8 }).withMessage(this.errorMessages.postCodeInvalid),
```
tomsmyers, line 65: could we apply some basic regex to this? 

i know it's a thorny subject but there's some examples here we could use https://stackoverflow.com/questions/164979/regex-for-matching-uk-postcodes

```
^([Gg][Ii][Rr] 0[Aa]{2})|((([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})|(([AZa-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9]?[A-Za-z]))))[0-9][A-Za-z]{2})$
```

tomsmyers (approved):
this all looks great andrew - nice work. 

i don't have any issues. couple of smaller comments primarily on postcode validation and the address as string array thing that i'm still not fully comprehending.

also on making sure this is readily reusable for intial assessments.

server/models/actionPlan.ts
andrewburrowsmoj: It is nullable to be consistent with the fields `appointmentTime` and `durationInMinutes` also being nullable.

It needs to be nullable because when the user clicks on the edit appointment details page these specific fields will be null on the Appointment object but the Appointment itself will exist.

server/routes/serviceProviderReferrals/editSessionForm.ts
andrewburrowsmoj: It's impossible for it to be undefined because the FormValidation will cause an error to be created if it is null. This is following on the conventions for `durationInMinutes` and `appointmentTime` which also use `!` despite it being impossible.

In my opinion the `!` is not necessary because the underlying value is not defined as `undefinable`but is an additional assertation that if it's undefined then something really bad has gone on. Would you like me to remove it?

server/routes/serviceProviderReferrals/editSessionPresenter.test.ts
andrewburrowsmoj: I will restore this.

server/routes/serviceProviderReferrals/editSessionView.ts
andrewburrowsmoj: Sure, that would be nice.

I don't think we have used this specific pattern before except possibly with LayoutView. The other shared Views have an inheritance relation rather than composition.

So my plan would be to  create an `addressFormView.ts` and an `addressFormPresenter.ts`. Does this sound agreeable to you?

server/routes/serviceProviderReferrals/editSessionView.ts
andrewburrowsmoj: That is correct.

server/utils/address.ts
andrewburrowsmoj: We had a discussion about this a while ago and we decided that we wanted to make the address fields agnostic to the type for instance, npsOfficeCode, otherLocationsAddress and videoCallUrl (which has now been removed).

This means that for the JSON itself sending to the server will be agnostic.

I have the issue whereby once I have a `requirement` defined in my mind I don't question it as heavily; I think perhaps this requirement was super-ceded with another design.

I do prefer to have explicit fields for the address rather than an array. If you are in agreement, I will change to using fields rather than array.

server/utils/forms/inputs/addressInput.ts
andrewburrowsmoj: Sure. I can do that. Leigh indicated that validation on the fields is not important; but this seems like a worthwhile validation to add in.

PR #438, opened 2021-06-16
IC-1866 add more fine grained control over ignored and unsuccessful requests in app insights

tomsmyers:

add more fine grained control over ignored and unsuccessful requests in app insights
 - allows requests matching patterns to be ignored
 - allows customization of which status codes are treated as errors in the UI

make app insights more useable

### DIFF
--- a/server/azureAppInsights.ts
+++ b/server/azureAppInsights.ts
@@ -4,14 +4,20 @@ import applicationVersion from './applicationVersion'
 import config from './config'
 import logger from '../log'
 
-function ignoreStaticAssetsProcessor(
+function ignoreExcludedRequestsProcessor(
   envelope: Contracts.EnvelopeTelemetry,
   _contextObjects: { [name: string]: unknown } | undefined
 ): boolean {
   if (envelope.data.baseType === Contracts.TelemetryTypeString.Request) {
     const requestData = envelope.data.baseData
-    if (requestData instanceof Contracts.RequestData && requestData.name.startsWith('GET /assets/')) {
-      return false
+    if (requestData instanceof Contracts.RequestData) {
+      const { excludedRequests } = config.applicationInsights
+      for (let i = 0; i < excludedRequests.length; i += 1) {
+        const pattern = excludedRequests[i]
+        if (requestData.name.match(pattern) !== null) {
+          return false
+        }
+      }
     }
   }
   return true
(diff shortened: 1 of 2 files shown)


server/azureAppInsights.ts
```diff
@@ -4,14 +4,20 @@ import applicationVersion from './applicationVersion'
...
   envelope: Contracts.EnvelopeTelemetry,
   _contextObjects: { [name: string]: unknown } | undefined
 ): boolean {
   if (envelope.data.baseType === Contracts.TelemetryTypeString.Request) {
     const requestData = envelope.data.baseData
-    if (requestData instanceof Contracts.RequestData && requestData.name.startsWith('GET /assets/')) {
-      return false
+    if (requestData instanceof Contracts.RequestData) {
+      const { excludedRequests } = config.applicationInsights
+      for (let i = 0; i < excludedRequests.length; i += 1) {
```
Gweaton, line 15: This doesn't really matter, and I might be missing something or could this be an ES6 `for...of`? Is the index important?

server/azureAppInsights.ts
```diff
@@ -31,6 +37,21 @@ function addUsernameProcessor(
   return true
 }
 
+function errorStatusCodeProcessor(
+  envelope: Contracts.EnvelopeTelemetry,
+  _contextObjects: { [name: string]: unknown } | undefined
+): boolean {
+  if (envelope.data.baseType === Contracts.TelemetryTypeString.Request) {
+    if (envelope.data.baseData !== undefined) {
```
Gweaton, line 45: can this be part of the above `if` as an `&&` condition so we don't have the nested `if`?

Gweaton (approved):
I'm not sure I understand all the config changes here, but have left a couple of pedantic JS comments - enjoy!

server/azureAppInsights.ts
tomsmyers: intellij really doesn't like you using that thing... the reasons are compelling but boring

server/azureAppInsights.ts
tomsmyers: yes 👍

server/azureAppInsights.ts
```diff
@@ -25,12 +31,25 @@ function addUsernameProcessor(
...
     if (userId) {
       // eslint-disable-next-line no-param-reassign
-      envelope.tags['ai.user.authUserId'] = userId
+      envelope.tags[defaultClient.context.keys.userAuthUserId] = userId
     }
   }
   return true
 }
 
+// function errorStatusCodeProcessor(
```
sldblog, line 40: Can these commented blocks be removed until they're ready to use?

sldblog (commented):
So if I understand correctly, the core aim of this is to ignore requests that `startswith "GET /health"`?

👍 Thanks

PR #439, opened 2021-06-16
IC-1874: Allow users to go back to EOSR creation page after clicking the "back" button.

Gweaton:

Ensures "Create End of Service Report" link is displayed if a user clicks the "back" button from that page.

As part of this, we want to ensure that we don't try and create a new End of Service report here when clicking this button if one already exists - we could implement this logic on the backend POST endpoint to return the existing End of Service report, so that it won't try and create a duplicate EOSR, but this felt like an easy approach for now.

To stop users being stuck and unable to create an end of service report.

### DIFF
--- a/server/routes/serviceProviderReferrals/interventionProgressPresenter.ts
+++ b/server/routes/serviceProviderReferrals/interventionProgressPresenter.ts
@@ -185,11 +185,19 @@ export default class InterventionProgressPresenter {
     ? 'active'
     : 'inactive'
 
-  private get endOfServiceReportSubmitted() {
-    return this.referral.endOfServiceReport !== null && this.referral.endOfServiceReport.submittedAt !== null
+  get endOfServiceReportSubmitted(): boolean {
+    return !!this.referral.endOfServiceReport?.submittedAt
   }
 
-  get allowEndOfServiceReportCreation(): boolean {
-    return this.referral.endOfServiceReport === null
+  private get endOfServiceReportStarted(): boolean {
+    return this.referral.endOfServiceReport !== null && !this.endOfServiceReportSubmitted
+  }
+
+  get endOfServiceReportButtonActionText(): string | null {
+    if (this.endOfServiceReportSubmitted) {
+      return null
+    }
+
+    return this.endOfServiceReportStarted ? 'Continue' : 'Create'
   }
 }
--- a/server/routes/serviceProviderReferrals/interventionProgressPresenter.ts
+++ b/server/routes/serviceProviderReferrals/interventionProgressPresenter.ts
@@ -189,6 +189,10 @@ export default class InterventionProgressPresenter {
     return !!this.referral.endOfServiceReport?.submittedAt
   }
 
+  get canSubmitEndOfServiceReport(): boolean {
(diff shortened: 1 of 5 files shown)


server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
```diff
@@ -678,12 +678,23 @@ export default class ServiceProviderReferralsController {
...
 
   async createDraftEndOfServiceReport(req: Request, res: Response): Promise<void> {
-    const draftEndOfServiceReport = await this.interventionsService.createDraftEndOfServiceReport(
-      res.locals.user.token.accessToken,
-      req.params.id
-    )
+    const { accessToken } = res.locals.user.token
+    const referralId = req.params.id
+
+    const referral = await this.interventionsService.getSentReferral(accessToken, referralId)
```
Gweaton, line 684: @tomsmyers do you think it's fine to do this on the Frontend, or should we update the API endpoint to return an existing EOSR if one already exists when we call `createDraftEndOfServiceReport`?

tomsmyers: we can do this here, but we should also not allow more than one EOSR to be created for the referral in the API. currently it just creates another.

server/routes/serviceProviderReferrals/interventionProgressPresenter.ts
```diff
@@ -185,11 +185,7 @@ export default class InterventionProgressPresenter {
     ? 'active'
     : 'inactive'
 
-  private get endOfServiceReportSubmitted() {
+  get endOfServiceReportSubmitted(): boolean {
     return this.referral.endOfServiceReport !== null && this.referral.endOfServiceReport.submittedAt !== null
```
tomsmyers, line 189: this can be simplified using `?.`

tomsmyers (commented):
i don't see the code here to allow the user to get back to the EOSR creation page?

tomsmyers:
ah, disregard my comment above. i understand now. 

should we follow the pattern we used with action plans here instead? whereby it's created, and once it's created the link no longer says "create", but "continue".

tomsmyers:
> @tomsmyers I've pushed a few changes to this for the time being e.g. "Continue" vs "create" - there's still some logic to work on for when all sessions have been completed etc. but as that's waiting for possible back-end changes, should we get this in as-is, as it's at least an improvement.

yeh sure thing - thanks

server/routes/serviceProviderReferrals/interventionProgressPresenter.ts
```diff
@@ -185,11 +185,23 @@ export default class InterventionProgressPresenter {
     ? 'active'
     : 'inactive'
 
-  private get endOfServiceReportSubmitted() {
-    return this.referral.endOfServiceReport !== null && this.referral.endOfServiceReport.submittedAt !== null
+  get endOfServiceReportSubmitted(): boolean {
+    return !!this.referral.endOfServiceReport?.submittedAt
```
tomsmyers, line 189: not sure i'm a fan of `!!` but i'll roll with it !(!)

tomsmyers (approved):

PR #440, opened 2021-06-16
add simple page for SP to view an under approval action plan

tomsmyers:

add simple page for SP to view an under approval action plan

this change is intentionally small and with limited reach. 

i'm gonna refactor various parts of this code to work across PP and SP pages as i go, but don't want a huge change set. 

this means once an SP has submitted the plan they can at least see it by clicking the link. which is kinda an MVP for 'view action plan' IMO.  

Screenshot 2021-06-16 at 12 17 25

### DIFF
--- a/assets/sass/components/_shaded-list.scss
+++ b/assets/sass/components/_shaded-list.scss
@@ -1,0 +1,10 @@
+.refer-and-monitor__shaded-list-item {
+  background-color: govuk-colour('light-grey');
+  border-left: 5px solid govuk-colour('blue');
+  margin-bottom: 30px;
+  padding: 15px;
+
+  .govuk-body:last-child {
+    margin-bottom: 5px;
+  }
+}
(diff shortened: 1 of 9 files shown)


assets/sass/components/_shaded-list.scss
```diff
@@ -0,0 +1,10 @@
+.refer-and-monitor__shaded-list-item {
```
Gweaton, line 1: You can use `app-inset-text--grey` here - I added it in my branch where I added the new action plan styles (see https://github.com/ministryofjustice/hmpps-interventions-ui/blob/3b48c13868b0dfa56d38f389a771ef9c64c4b6cc/server/routes/serviceProviderReferrals/reviewActionPlanView.ts#L17)

Gweaton (commented):
Looks good - just one thing re: css class

PR #442, opened 2021-06-16
PPD-299: Fixing npm security audit threshold

brightonsbox:
## What does this pull request do?

Corrects the threshold used to flag the nightly npm security audit as a pass or failure

## What is the intent behind these changes?

To make security vulnerabilities more visible

PR #443, opened 2021-06-16
DT-2208: 🐛 Fix audit so that actually runs

petergphillips:

_Required._

_Required._


sldblog:
Which one of #442 #443 should be the winner? 😁

PR #444, opened 2021-06-17
fix creation of action plans by ensuring the referral ID is always valid

tomsmyers:

fix creation of action plans by ensuring the referral ID is always valid

allow users to create action plans


sldblog:
> referral ID is always valid

Interesting, what situation would create an invalid state?

tomsmyers:
@sldblog i refactored this out of another class and somehow i ended up pulling the referral id out of a nullable action plan when trying to create the action plan in the first place

lawrence-forooghian (approved):

PR #448, opened 2021-06-17
IC-1471: Add tests for Community API

Gweaton:

 Add tests for Community API and stops it instantiating a new instance of `RestClient` with each request, as this made it difficult to test.

We didn't have any tests for this, and I'm about to introduce some new functionality to the Community API Service to do things with the Service User address, so wanted to make sure we had some tests in place at least.


tomsmyers (approved):
nice - does this mean we can remove the token param from the restlclient constructor?

Gweaton:
> nice - does this mean we can remove the token param from the restlclient constructor?

No, we're still using it in the `HMPPSAuthClient` and the `InterventionsService` (which we were able to test due to pact doing a lot of mocking for us), but hopefully there'll be time to refactor this later.

PR #458, opened 2021-06-21
IC-1829 PP can see SU's risk scores

tomsmyers:

This is the first part of this ticket which shows the 'ROSH analysis' table to the PP in the risk information part of the referral form.

Screenshot 2021-06-21 at 13 38 21

Screenshot 2021-06-21 at 13 38 29

show PP's the risk info from ARN.

### DIFF
--- a/server/models/assessRisksAndNeeds/riskSummary.ts
+++ b/server/models/assessRisksAndNeeds/riskSummary.ts
@@ -1,0 +1,10 @@
+export type RiskScore = 'LOW' | 'MEDIUM' | 'HIGH' | 'VERY_HIGH'
+
+export default interface RiskSummary {
+  whoIsAtRisk?: string | null
+  natureOfRisk?: string | null
+  riskImminence?: string | null
+  riskIncreaseFactors?: string | null
+  riskMitigationFactors?: string | null
+  riskInCommunity: Partial<Record<RiskScore, string[]>>
+}
--- a/server/routes/referrals/riskInformationView.ts
+++ b/server/routes/referrals/riskInformationView.ts
@@ -1,6 +1,7 @@
 import RiskInformationPresenter from './riskInformationPresenter'
 import ViewUtils from '../../utils/viewUtils'
-import { TextareaArgs } from '../../utils/govukFrontendTypes'
+import { TableArgs, TagArgs, TextareaArgs } from '../../utils/govukFrontendTypes'
+import utils from '../../utils/utils'
 
 export default class RiskInformationView {
   constructor(private readonly presenter: RiskInformationPresenter) {}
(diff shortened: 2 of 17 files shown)


server/models/assessRisksAndNeeds/riskSummary.ts
```diff
@@ -0,0 +1,10 @@
+export type RiskScore = 'LOW' | 'MEDIUM' | 'HIGH' | 'VERY_HIGH'
+
+export default interface RiskSummary {
+  whoIsAtRisk?: string | null
+  natureOfRisk?: string | null
+  riskImminence?: string | null
+  riskIncreaseFactors?: string | null
+  riskMitigationFactors?: string | null
+  riskInCommunity: Partial<Record<RiskScore, string[]>>
```
andrewburrowsmoj, line 9: do you think you have to cater for riskInCommunity being null? (given than all other fields are nullable), aslo perhaps annoyingly the string[] being nullable?

(not sure where the api is located)

tomsmyers: this is a `val riskInCommunity: Map<RiskLevel, List<String>>` in the API

server/routes/referrals/riskInformationView.ts
```diff
@@ -23,13 +24,43 @@ export default class RiskInformationView {
...
 
+  roshAnalysisTableArgs(tagMacro: (args: TagArgs) => string): TableArgs {
+    return {
+      head: this.presenter.roshAnalysisHeaders.map((header: string) => {
+        return { text: header }
+      }),
+      rows: this.presenter.roshAnalysisRows.map(row => {
+        return [
+          { text: utils.convertToProperCase(row.riskTo) },
+          { text: tagMacro({ text: row.riskScore, classes: this.tagClassForRiskScore(row.riskScore) }) },
```
andrewburrowsmoj, line 35: (minor suggestion): Perhaps can reuse PresenterUtils.complexityLevelTagArgs?
Seems that this can be considered a UI component independent of underlying field use (risk vs complexity)?

server/views/referrals/riskInformation.njk
```diff
@@ -28,6 +29,25 @@
 
       {{ govukTable(roshAnalysisTableArgs(govukTag)) }}
 
+      {{ govukDetails({
+          summaryText: "Definitions of risk levels",
+          html:
+            '<ul class="govuk-list govuk-list--bullet">
```
Gweaton, line 35: Not a blocker, but we tend to construct this sort of thing in the `view` layer as `govUkDetailsArgs` and pass it to the macro to keep the template a bit smaller.

andrewburrowsmoj (approved):
lgtm (just made a minor suggestion - not necessary unless you agree)

server/services/assessRisksAndNeedsService.ts
```diff
@@ -6,13 +6,16 @@ import { SupplementaryRiskInformation } from '../models/assessRisksAndNeeds/supp
...
-
+  async getSupplementaryRiskInformation(
+    riskId: string,
+    token: string | null = null
+  ): Promise<SupplementaryRiskInformation> {
     logger.info({ riskId }, 'getting supplementary risk information')
     return (await this.restClient.get({
       path: `/risks/supplementary/${riskId}`,
-      token,
+      token: token || (await this.hmppsAuthService.getApiClientToken()),
```
Gweaton, line 17: Am I missing something here? What's this change for? Is it to do with the 2 different types of credentials? Would you mind adding a bit of a description to the commit message as to why it's necessary?

Gweaton (commented):
Couple of small qs

server/views/referrals/riskInformation.njk
tomsmyers: 👍

server/services/assessRisksAndNeedsService.ts
tomsmyers: yeh there are some changes coming to ARN to use the user tokens for all the endpoints. will mention in commit message.

Gweaton (approved):
Looks good - nice work on the feature flag. As mentioned on slack, would be good to add a test for `getRiskSummary before merging.

PR #459, opened 2021-06-21
IC-1902 & IC-1596: Fetch SU address from Community API and display throughout the service

Gweaton:

Adds call to the community API endpoint for `/all` to fetch additional SU data, particularly the SU's address history.

We don't want to use this endpoint everywhere, as the payload might be huge, so this replaces the usual call for SU details on the pages we display the address.

To display the SU address to PPs and SPs so they know where the referred user is located.

### Refer journey: Confirm service user details
image

### Monitor/record journey: Show referral page
image

### DIFF
--- a/mockApis/communityApi.ts
+++ b/mockApis/communityApi.ts
@@ -19,6 +19,22 @@ export default class CommunityApiMocks {
     })
   }
 
+  stubGetExpandedServiceUserByCRN = async (crn: string, responseJson: unknown): Promise<unknown> => {
+    return this.wiremock.stubFor({
+      request: {
+        method: 'GET',
+        urlPattern: `/community-api/secure/offenders/crn/${crn}/all`,
+      },
+      response: {
+        status: 200,
+        headers: {
+          'Content-Type': 'application/json',
+        },
+        jsonBody: responseJson,
+      },
+    })
+  }
+
   stubGetUserByUsername = async (username: string, responseJson: unknown): Promise<unknown> => {
     return this.wiremock.stubFor({
       request: {
--- a/server/decorators/expandedDeliusServiceUserDecorator.test.ts
+++ b/server/decorators/expandedDeliusServiceUserDecorator.test.ts
@@ -1,0 +1,292 @@
+import expandedDeliusServiceUserFactory from '../../testutils/factories/expandedDeliusServiceUser'
+import ExpandedDeliusServiceUserDecorator from './expandedDeliusServiceUserDecorator'
+
+describe(ExpandedDeliusServiceUserDecorator, () => {
+  describe('address', () => {
+    describe('selecting the current address', () => {
+      describe('when there is only one address returned from nDelius', () => {
+        it('uses that address as the current address', () => {
+          const serviceUser = new ExpandedDeliusServiceUserDecorator(
+            expandedDeliusServiceUserFactory.build({
+              contactDetails: {
+                addresses: [
+                  {
+                    addressNumber: 'Flat 2',
+                    buildingName: null,
+                    streetName: 'Test Walk',
+                    postcode: 'SW16 1AQ',
+                    town: 'London',
+                    district: 'City of London',
+                    county: 'Greater London',
+                    from: '2019-01-01',
+                    to: null,
+                    noFixedAbode: false,
+                  },
+                ],
+              },
+            })
+          )
+
+          expect(serviceUser.address).toEqual([
+            'Flat 2 Test Walk',
+            'London',
+            'City of London',
+            'Greater London',
+            'SW16 1AQ',
+          ])
+        })
+      })
+
+      describe('when there are multiple addresses returned from nDelius', () => {
+        describe('when all addresses have a non-null "from" field', () => {
+          it('uses the address with the most recent "from" date', () => {
+            const oldAddress = {
+              addressNumber: 'Flat 2',
+              buildingName: null,
+              streetName: 'Test Walk',
+              postcode: 'SW16 1AQ',
+              town: 'London',
+              district: 'City of London',
+              county: 'Greater London',
+              from: '2019-01-01',
+              to: null,
+              noFixedAbode: false,
+            }
+
+            const newAddress = {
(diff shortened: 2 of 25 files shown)


mockApis/communityApi.ts
```diff
@@ -7,7 +7,7 @@ export default class CommunityApiMocks {
     return this.wiremock.stubFor({
       request: {
         method: 'GET',
-        urlPattern: `/community-api/secure/offenders/crn/${crn}`,
+        urlPattern: `/community-api/secure/offenders/crn/${crn}/all`,
```
Gweaton, line 10: Have just realised this response might be massive - am going to make this an additional endpoint so we're not calling it on every single page for the SU banner.

Gweaton: Have reworked this.

server/models/delius/deliusServiceUser.ts
```diff
@@ -11,11 +11,28 @@ interface OffenderLanguages {
   primaryLanguage: string
 }
 
+interface Address {
```
tomsmyers: when deserializing responses, does it matter if the field is missing? will it just cast it to null?

server/services/communityApiService.ts
```diff
@@ -22,6 +22,13 @@ export default class CommunityApiService {
     return (await this.restClient.get({ path: `/secure/offenders/crn/${crn}`, token })) as DeliusServiceUser
   }
 
+  async getExpandedServiceUserByCRN(crn: string): Promise<ExpandedDeliusServiceUser> {
+    const token = await this.hmppsAuthService.getApiClientToken()
+
+    logger.info({ crn }, 'getting all details for offender')
+    return (await this.restClient.get({ path: `/secure/offenders/crn/${crn}/all`, token })) as ExpandedDeliusServiceUser
```
tomsmyers, line 29: could we wrap rest client errors with a user message that says something about 'could not get service user details from nDelius'?

it would be

try {
  ... 
} catch (err) {
  throw createError(err.status, err, { userMessage: '...' })
}

server/decorators/expandedDeliusServiceUserDecorator.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (!addresses) {
+      return ['Not found']
+    }
+
+    if (addresses.length === 1) {
+      return this.deliusAddressToArray(addresses[0])
+    }
+
+    // If we have no "from" date, how do we know which is the most recent?
+    if (addresses.every(address => address.from === null)) {
```
tomsmyers, line 18: agree this is a bit iffy. can we log a warning then we know if anyone hits this in prod?

server/decorators/expandedDeliusServiceUserDecorator.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    const mostRecentAddress = addresses.sort((a, b) => new Date(b.from!).getTime() - new Date(a.from!).getTime())[0]
+
+    return this.deliusAddressToArray(mostRecentAddress)
+  }
+
+  private deliusAddressToArray(address: Address): string[] {
+    // required to force removal of null from (string | null)[]
+    function notEmpty(value: string | null | undefined): value is string {
```
tomsmyers, line 29: what does this return type mean? isn't it boolean?

tomsmyers (approved):
couple of comments but great work on short notice

server/decorators/expandedDeliusServiceUserDecorator.test.ts
```diff
@@ -0,0 +1,293 @@
...
+        it('returns "Not found"', () => {
+          const serviceUser = new ExpandedDeliusServiceUserDecorator(
+            expandedDeliusServiceUserFactory.build({
+              contactDetails: {
+                addresses: [],
+              },
+            })
+          )
+
+          expect(serviceUser.address).toEqual(['Not found'])
```
lawrence-forooghian, line 194: I reckon that deciding to display a "not found" string is beyond the responsiblities of a model. How about returning `null` from this method in the case where there's no address, and then the presenter can decide what to do with that?

server/models/delius/deliusServiceUser.ts
lawrence-forooghian: Yeah, if the field might not be present at all in the response then these should be represented as `addressNumber?` etc, I think.

server/decorators/expandedDeliusServiceUserDecorator.ts
Gweaton: Good idea. Do we want it at `warn` or `error` level?

server/services/communityApiService.ts
Gweaton: Yep, sure.

server/decorators/expandedDeliusServiceUserDecorator.ts
tomsmyers: erm. yeh error for now.

server/models/delius/deliusServiceUser.ts
Gweaton: Have updated this - I think we used ` | null` for most of the values in the past, but it probably should be `undefined` instead if it's not in the response at all (which is the case).

server/decorators/expandedDeliusServiceUserDecorator.ts
Gweaton: It's essentially asserting that the return value is a string, and can't be `null / undefined` because the compiler doesn't believe that we're handling `null` here. I've added a comment for this and a link to the stack overflow where I got this information from.

server/decorators/expandedDeliusServiceUserDecorator.test.ts
Gweaton: Fair point - have changed this and started handling in the presenter.

server/decorators/expandedDeliusServiceUserDecorator.ts
Gweaton: Done

PR #461, opened 2021-06-21
only propagate 403 response codes if they are likely to have come from interventions-service

tomsmyers:

makes the error handler a bit more picky about which 403 responses it propagates in its response

stop 403 errors from downstream services being missed by our monitoring tools


sldblog (approved):
LGTM 👍 

I'm a bit surprised this doesn't need any test changes though, this feels like an important assumption about structure and its outcomes (5xx page or not 5xx page)

PR #466, opened 2021-06-22
Temporarily comment out appointment feedback service calls

lawrence-forooghian:
I added these prematurely alongside the service calls for scheduling
appointments. The interventions service hasn’t added this functionality
yet and we don’t want failing contract tests.

I’m commenting out instead of removing, in order to reduce diff noise.
We _will_ be adding this functionality back.


sldblog:
I'm more in favour of removing code and using revert commits than comments -- but whatever works :)

sldblog (approved):

PR #469, opened 2021-06-22
Revert "include more ROSH risk information about the SU in the 'risk information' page"

tomsmyers:
Revert "include more ROSH risk information about the SU in the 'risk information' page"

This reverts commit ef3fd11e61111018b13caf026ca82a4f860d5489.

there is some back and forth about if this conent is going to be included for day 1. right now the concensus is "no".


Gweaton (approved):

sldblog:
👍 Thanks for the context

PR #475, opened 2021-06-23
IC-1830 Show risk summary in referral details

tomsmyers:

Show risk summary in referral details for SPs and PPs

Screenshot 2021-06-23 at 16 44 08

tie up the new risk summary info into the referral details screens.

### DIFF
--- a/server/routes/shared/riskPresenter.ts
+++ b/server/routes/shared/riskPresenter.ts
@@ -1,0 +1,27 @@
+import RiskSummary from '../../models/assessRisksAndNeeds/riskSummary'
+import config from '../../config'
+
+export interface RoshAnalysisTableRow {
+  riskTo: string
+  riskScore: string
+}
+
+export default class RiskPresenter {
+  constructor(private readonly riskSummary: RiskSummary | null) {}
+
+  readonly riskSummaryEnabled = config.apis.assessRisksAndNeedsApi.riskSummaryEnabled
+
+  get roshAnalysisHeaders(): string[] {
+    return ['Risk to', 'Risk in community']
+  }
+
+  get roshAnalysisRows(): RoshAnalysisTableRow[] {
+    if (this.riskSummary === null) return []
+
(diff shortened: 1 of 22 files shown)


server/routes/shared/riskPresenter.ts
```diff
@@ -0,0 +1,26 @@
...
+  get roshAnalysisRows(): RoshAnalysisTableRow[] {
+    if (this.riskSummary === null) return []
+
+    return Object.entries(this.riskSummary.riskInCommunity).flatMap(([riskScore, riskGroups]) => {
+      return riskGroups.map(riskTo => {
+        return { riskTo, riskScore }
+      })
+    })
+  }
+}
```
andrewburrowsmoj, line 26: Do you have tests being performed on these lines of code? I tried searching but couldn't find anything.

server/routes/shared/riskView.ts
```diff
@@ -0,0 +1,62 @@
+import { DetailsArgs, TableArgs, TagArgs } from '../../utils/govukFrontendTypes'
+import utils from '../../utils/utils'
+import RiskPresenter from './riskPresenter'
+
+export default class RiskView {
```
andrewburrowsmoj, line 5: Do you think it might be good to have some way to separate classes which are used to render a page (i.e. extend PageContentView) and classes which are used for composition of page rendering classes?

I had a similar situation with addresses and set it as a Component class rather than View class `addressFormComponent.ts`

tomsmyers: yeh that's a good idea. we have a few like this so i think we should get together and come up with a pattern we can apply across all the shared view / presenter classes.

andrewburrowsmoj (approved):
Just left two comments

server/routes/shared/riskPresenter.ts
tomsmyers: no but there should be. i'll add something now.

PR #477, opened 2021-06-23
IC-1962: Display time-sensitive message to all users

Gweaton:

Adds a yellow notification banner to all screens for all users between the hours of 9am and 9pm on Saturday 26th June to let them know that Delius is in read-only mode during scheduled maintenance.

The functionality should be fairly easy to extend if we want to do this again in future.

To notify users that Delius is down for scheduled maintenance so they know that the service won't function as expected.

## Screenshot

image

### DIFF
--- a/server/app.ts
+++ b/server/app.ts
@@ -25,6 +25,7 @@ import HmppsAuthService from './services/hmppsAuthService'
 import passportSetup from './authentication/passport'
 import AssessRisksAndNeedsService from './services/assessRisksAndNeedsService'
 import ControllerUtils from './utils/controllerUtils'
+import broadcastMessageConfig from './broadcast-message-config.json'
 
 const RedisStore = connectRedis(session)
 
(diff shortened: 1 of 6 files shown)


server/app.ts
```diff
@@ -25,6 +25,7 @@ import HmppsAuthService from './services/hmppsAuthService'
 import passportSetup from './authentication/passport'
 import AssessRisksAndNeedsService from './services/assessRisksAndNeedsService'
 import ControllerUtils from './utils/controllerUtils'
+import serviceUnavailableConfig from './service-unavailable-config.json'
```
sldblog, line 28: given it's a generic timed message, should we use a more generic name, e.g. broadcastMessages?

server/service-unavailable-config.json
```diff
@@ -0,0 +1,5 @@
+{
+  "start": "2021-06-26T07:00:00+01:00",
+  "end": "2021-06-26T21:00:00+01:00",
+  "message": "Some parts of Refer and monitor an intervention are unavailable until 9:00pm. You can read referral and intervention information but any updates you make will not be recorded."
```
sldblog, line 4: _not a blocker_ 🙋‍♂️ we could pull this directly from somewhere at startup (to make it changeable without deploy)

sldblog: should we have a probationMessage and a providerMessage?

server/app.ts
Gweaton: Yeah, I was really struggling to come up with a generic name (initially went for `userNotificationBanner/Message` but that felt like it clashed with our usual `notificationBanner` - I think `broadcastMessage` sounds quite nice 👍

server/service-unavailable-config.json
Gweaton: It might be useful in future, but for its current purpose we'll be displaying the same message to both sets of users. Shall we see if we need it further down the line?

tomsmyers (approved):
LGTM

PR #478, opened 2021-06-23
Merge `main` (which adds appointment method / address) into `feature/schedule-supplier-assessment`

lawrence-forooghian:

Merges `main` into the `feature/schedule-supplier-assessments` branch.

The merge commit here actually introduces a functionality change and a contract change - namely, it starts to add the ability to specify appointment delivery method / address on a supplier assessment appointment. See the commit message for https://github.com/ministryofjustice/hmpps-interventions-ui/commit/69b0bc261e62e31071c0ef735570066b63d395c5 for more details.

To bring the supplier assessments branch in line with `main` (fixing merge conflicts).

### DIFF
--- a/testutils/factories/appointment.ts
+++ b/testutils/factories/appointment.ts
@@ -1,6 +1,7 @@
 import { Factory } from 'fishery'
 import Appointment from '../../server/models/appointment'
 import { Attended } from '../../server/models/appointmentAttendance'
+import { AppointmentDeliveryType } from '../../server/models/appointmentDeliveryType'
 
 class AppointmentFactory extends Factory<Appointment> {
   newlyBooked() {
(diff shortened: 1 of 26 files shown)


testutils/factories/appointment.ts
```diff
@@ -24,10 +25,15 @@ class AppointmentFactory extends Factory<Appointment> {
...
 }
 
+const defaultAppointmentDeliveryType: AppointmentDeliveryType = 'VIDEO_CALL'
+
 export default AppointmentFactory.define(({ sequence }) => ({
   id: sequence.toString(),
   appointmentTime: new Date().toISOString(),
   durationInMinutes: 60,
+  // For some reason the compiler complains if I write 'VIDEO_CALL' inline
+  appointmentDeliveryType: defaultAppointmentDeliveryType,
```
andrewburrowsmoj, line 35: I think it assumes that 'VIDEO_CALL' is a string rather than a AppointmentDeliveryType, because the type is defined as `any` in the factory. I think your solution is the only way around it.

andrewburrowsmoj (approved):
The merge looks good to me.

PR #480, opened 2021-06-23
use user token to make calls to ARN supplementary risk API

tomsmyers:
**DO NOT MERGE YET**

use user token to make calls to ARN supplementary risk API instead of the client credentials token

ARN API is moving to a different auth model, where permissions are based on user tokens, not client tokens.

### DIFF
--- a/server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
+++ b/server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
@@ -93,7 +93,7 @@ export default class ServiceProviderReferralsController {
       this.communityApiService.getUserByUsername(sentReferral.sentBy.username),
       this.communityApiService.getExpandedServiceUserByCRN(crn),
       this.communityApiService.getConvictionById(crn, sentReferral.referral.relevantSentenceId),
-      this.assessRisksAndNeedsService.getSupplementaryRiskInformation(sentReferral.supplementaryRiskId),
+      this.assessRisksAndNeedsService.getSupplementaryRiskInformation(sentReferral.supplementaryRiskId, accessToken),
       this.assessRisksAndNeedsService.getRiskSummary(crn, accessToken),
     ])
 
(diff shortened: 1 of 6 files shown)


sldblog (approved):
LGTM

Does this need verification on dev? As in, does it block the pipeline until that's done?

server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
```diff
@@ -84,23 +84,18 @@ export default class ServiceProviderReferralsController {
...
       this.communityApiService.getExpandedServiceUserByCRN(sentReferral.referral.serviceUser.crn),
       this.communityApiService.getConvictionById(
         sentReferral.referral.serviceUser.crn,
         sentReferral.referral.relevantSentenceId
       ),
-      this.assessRisksAndNeedsService.getSupplementaryRiskInformation(sentReferral.supplementaryRiskId),
+      this.assessRisksAndNeedsService.getSupplementaryRiskInformation(sentReferral.supplementaryRiskId, accessToken),
     ])
 
     const assignee =
```
adelaidaa, line 101: looks good, but those 2 controllers seems to have identical code, what's different? I think the shared code could be extracted and reused to avoid duplication and having to test in both places.

tomsmyers:
@sldblog ARN are accepting both types of tokens whilst we do the switch over. we will validate in dev before pushing to prod both when rolling this out and when making the final switch in ARN.

server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
tomsmyers: agreed! we have quite a few instances of this... needs some tidy up. if only we had the luxury of time !

adelaidaa: i feel you!

PR #482, opened 2021-06-23
IC-1975: Use intervention title rather than contract type in SP & PP dashboards

Gweaton:

Uses intervention title rather than contract type in SP & PP dashboards

Intervention titles often include the region, which should make it easier for users to see, at a glance, where each referral is for.

## Screenshot

image

### DIFF
--- a/integration_tests/integration/probationPractitionerReferrals.spec.js
+++ b/integration_tests/integration/probationPractitionerReferrals.spec.js
@@ -18,8 +18,14 @@ describe('Probation practitioner referrals dashboard', () => {
 
   it("user logs in and sees 'My cases' screen with list of sent referrals", () => {
     const serviceCategory = serviceCategoryFactory.build()
-    const accommodationIntervention = interventionFactory.build({ contractType: { name: 'accommodation' } })
-    const womensServicesIntervention = interventionFactory.build({ contractType: { name: "womens' services" } })
+    const accommodationIntervention = interventionFactory.build({
+      contractType: { name: 'accommodation' },
+      title: 'accommodation services - west midlands',
+    })
+    const womensServicesIntervention = interventionFactory.build({
+      contractType: { name: "women's services" },
+      title: "women's services - west midlands",
+    })
 
     const sentReferrals = [
       sentReferralFactory.build({
@@ -80,7 +86,7 @@ describe('Probation practitioner referrals dashboard', () => {
         {
           Referral: 'ABCABCA1',
           'Service user': 'George Michael',
-          'Service category': 'Accommodation',
+          'Intervention type': 'Accommodation Services - West Midlands',
           Provider: 'Harmony Living',
           Caseworker: 'Unassigned',
           Action: 'View',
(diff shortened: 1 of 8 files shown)


integration_tests/integration/probationPractitionerReferrals.spec.js
```diff
@@ -80,15 +86,15 @@ describe('Probation practitioner referrals dashboard', () => {
         {
           Referral: 'ABCABCA1',
           'Service user': 'George Michael',
-          'Service category': 'Accommodation',
+          'Service category': 'Accommodation Services - West Midlands',
```
sldblog, line 89: Could we change the key to "contract title", "intervention type" or similar?

Gweaton: Good call - will do

Gweaton: Maybe just "intervention"?

server/routes/serviceProviderReferrals/dashboardPresenter.test.ts
```diff
@@ -41,15 +41,19 @@ describe(DashboardPresenter, () => {
           { text: '26 Jan 2021', sortValue: '2021-01-26', href: null },
           { text: 'ABCABCA1', sortValue: null, href: null },
           { text: 'George Michael', sortValue: 'michael, george', href: null },
-          { text: 'Accommodation Services - West Midlands', sortValue: null, href: null },
+          {
+            text: 'Accommodation Services - West Midlands',
+            sortValue: 'accommodation services - west midlands',
```
lawrence-forooghian, line 46: > I believe it's sorted by the text value if this is null

Yep, that's right.

> I wanted to at least keep it consistent with the PP dashboard

Good idea, but with the above in mind it would perhaps be better to make the `sortValue` null on the PP dashboard instead?

integration_tests/integration/serviceProviderReferrals.spec.js
```diff
@@ -25,11 +25,13 @@ describe('Service provider referrals dashboard', () => {
 
     const personalWellbeingIntervention = interventionFactory.build({
       contractType: { code: 'PWB', name: 'Personal wellbeing' },
+      title: 'personal wellbeing - west midlands',
```
lawrence-forooghian, line 28: Is this what the data actually looks like in the interventions service? Seems odd that they're writing "west midlands" in lower case in the first place…

lawrence-forooghian (commented):
Looks good, just a couple of questions.

integration_tests/integration/serviceProviderReferrals.spec.js
Gweaton: I'm not sure to be honest, this was more to test that the uppercasing works.

server/routes/serviceProviderReferrals/dashboardPresenter.test.ts
Gweaton: Yeah, you're right - I did this the wrong way round... have just removed this commit and pushed a replacement.

lawrence-forooghian (approved):

PR #483, opened 2021-06-23
change ARN risk summary endpoint to retrieve rosh summary _and_ risk to self

tomsmyers:

returns more information in the risk summary endpoint so we don't have to call another endpoint later to get risk to self data.

prepare for changes which add 'risk to self'

### DIFF
--- a/server/models/assessRisksAndNeeds/riskSummary.ts
+++ b/server/models/assessRisksAndNeeds/riskSummary.ts
@@ -1,10 +1,12 @@
 export type RiskScore = 'LOW' | 'MEDIUM' | 'HIGH' | 'VERY_HIGH'
 
 export default interface RiskSummary {
-  whoIsAtRisk?: string | null
-  natureOfRisk?: string | null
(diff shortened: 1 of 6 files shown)


server/models/assessRisksAndNeeds/riskSummary.ts
```diff
@@ -1,10 +1,12 @@
 export type RiskScore = 'LOW' | 'MEDIUM' | 'HIGH' | 'VERY_HIGH'
 
 export default interface RiskSummary {
-  whoIsAtRisk?: string | null
-  natureOfRisk?: string | null
-  riskImminence?: string | null
-  riskIncreaseFactors?: string | null
-  riskMitigationFactors?: string | null
-  riskInCommunity: Partial<Record<RiskScore, string[]>>
+  summary: {
```
Gweaton, line 4: I left a comment about this on your other PR,  but it should have come in this one - should this interface now be just `Risk(s?)` rather than `RiskSummary` if it has a `summary` field on it?

PR #484, opened 2021-06-23
IC-1829 include further risk information for PPs 

tomsmyers:

Adds the following section between 'ROSH analysis' and 'Additional referral information' on the PP facing risk and referral details screens. 

There is some setup code in the first two commits so review commit by commit is recommended. 

Screenshot 2021-06-23 at 21 19 52

### DIFF
--- a/server/models/assessRisksAndNeeds/riskSummary.ts
+++ b/server/models/assessRisksAndNeeds/riskSummary.ts
@@ -1,10 +1,12 @@
 export type RiskScore = 'LOW' | 'MEDIUM' | 'HIGH' | 'VERY_HIGH'
 
 export default interface RiskSummary {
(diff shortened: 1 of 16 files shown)


server/models/assessRisksAndNeeds/riskSummary.ts
```diff
@@ -1,10 +1,12 @@
 export type RiskScore = 'LOW' | 'MEDIUM' | 'HIGH' | 'VERY_HIGH'
 
 export default interface RiskSummary {
```
Gweaton, line 3: Not that big a deal, but should this be a top level `Risk(s)` interface rather than `RiskSummary`, now that it's got a summary field)

Gweaton (approved):
Looks good - just one comment re: naming

PR #493, opened 2021-06-24
IC-1830 do not error on missing risk data

tomsmyers:

instead of returning an error page for 404s, show some copy to the PP in the make screen saying the risk information doesn't exist, and then just don't show it in the referral details screens.

after discussing this with ARN team, we decided that it is not ok to continue as normal on unexpected errors from ARN (e.g. oasys down, or duplicate CRNs in oasys etc). these errors could mask serious risks. the case we want to handle is that the risk data is just not present or out of date. and in this case we show the PP (only) a message informing them as such. they can then enter the additional risk information and continue the referral as normal.

Allow referrals to be made for SUs with out of date or otherwise missing risk data.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -93,6 +93,21 @@ services:
     environment:
       - SERVER_PORT=8080
       - OAUTH_ENDPOINT_URL=http://hmpps-auth:8090/auth
+      - ASSESSMENTAPI_BASEURL=http://offender-assessments-api:8080
+
+  offender-assessments-api:
+    image: quay.io/hmpps/offender-assessments-api:latest
+    networks:
+      - hmpps
+    container_name: offender-assessments-api
+    restart: always
+    ports:
+      - "8095:8080"
+    healthcheck:
+      test: [ "CMD", "curl", "-f", "http://localhost:8080/health" ]
+    environment:
+      - SERVER_PORT=8080
+      - OAUTH_ENDPOINT_URL=http://hmpps-auth:8090/auth
 
 networks:
   hmpps:
--- a/server/routes/referrals/riskInformationPresenter.ts
+++ b/server/routes/referrals/riskInformationPresenter.ts
@@ -32,16 +32,16 @@ export default class RiskInformationPresenter {
   }
 
   get additionalRiskInformation(): Record<string, string | null> {
-    return this.riskPresenter.riskSummaryEnabled
+    return this.riskPresenter.riskSummaryNotFound
       ? {
-          label: 'Additional risk information',
-          labelClasses: 'govuk-label--l',
+          label: `Information for the service provider about ${this.referral.serviceUser?.firstName}’s risks`,
+          labelClasses: 'govuk-label--s',
           hint: 'Give any other information that is relevant to this referral. Do not include sensitive information about the individual or third parties.',
           errorMessage: PresenterUtils.errorMessage(this.error, 'additional-risk-information'),
         }
       : {
-          label: `Information for the service provider about ${this.referral.serviceUser?.firstName}’s risks`,
-          labelClasses: 'govuk-label--s',
+          label: 'Additional risk information',
+          labelClasses: 'govuk-label--l',
           hint: 'Give any other information that is relevant to this referral. Do not include sensitive information about the individual or third parties.',
           errorMessage: PresenterUtils.errorMessage(this.error, 'additional-risk-information'),
         }
--- a/server/routes/shared/riskPresenter.ts
+++ b/server/routes/shared/riskPresenter.ts
@@ -1,4 +1,5 @@
 import RiskSummary from '../../models/assessRisksAndNeeds/riskSummary'
+import config from '../../config'
 
 export interface RoshAnalysisTableRow {
   riskTo: string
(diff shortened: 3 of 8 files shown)


server/routes/referrals/riskInformationPresenter.ts
```diff
@@ -32,16 +32,16 @@ export default class RiskInformationPresenter {
   }
 
   get additionalRiskInformation(): Record<string, string | null> {
-    return this.riskPresenter.riskSummaryEnabled
+    return this.riskPresenter.riskSummaryNotFound
       ? {
-          label: 'Additional risk information',
```
tomsmyers, line 37: the changes here are swapping the order of the ternary, `since riskSummaryEnabled ~== !riskSummaryNotFound`

server/views/referrals/riskInformation.njk
```diff
@@ -25,15 +25,23 @@
 
       {% if presenter.riskPresenter.riskSummaryEnabled %}
 
-        <p class="govuk-body">This shows the current ROSH levels and risk to self information from OASys. You can add information for the service provider.</p>
+        {% if presenter.riskPresenter.riskSummaryNotFound %}
 
-        <h2 class="govuk-heading-l">ROSH analysis</h2>
+          <p class="govuk-body">Up to date risk information for this service user was not found in OASys. You can still add additional risk information for the service provider.</p>
```
tomsmyers, line 30: this is the crux of the content changes

server/routes/shared/riskPresenter.ts
```diff
@@ -8,7 +9,9 @@ export interface RoshAnalysisTableRow {
 export default class RiskPresenter {
   constructor(private readonly riskSummary: RiskSummary | null) {}
 
-  readonly riskSummaryEnabled = this.riskSummary !== null
+  readonly riskSummaryEnabled = config.apis.assessRisksAndNeedsApi.riskSummaryEnabled
+
+  readonly riskSummaryNotFound = this.riskSummary === null
```
tomsmyers, line 14: this legitimizes `null` as a value for `riskSummary`. therefore i changed the enabled flag to be based on the config only.

docker-compose.yml
```diff
@@ -93,6 +93,21 @@ services:
     environment:
       - SERVER_PORT=8080
       - OAUTH_ENDPOINT_URL=http://hmpps-auth:8090/auth
+      - ASSESSMENTAPI_BASEURL=http://offender-assessments-api:8080
```
sldblog, line 96: transitive dependencies? ouch

server/services/assessRisksAndNeedsService.test.ts
```diff
@@ -40,14 +40,20 @@ describe(AssessRisksAndNeedsService, () => {
...
         await assessRisksAndNeedsService.getRiskSummary('crn123', 'token')
       } catch (err) {
-        expect(err.status).toBe(404)
-        expect(err.userMessage).toBe('Could not get service user risk scores from OASys.')
+        expect(err.status).toBe(409)
+        expect(err.userMessage).toBe("Could not get service user's risk scores from OASys.")
       }
     })
+
+    it('return null when encountering 404 errors', async () => {
```
sldblog, line 52: What do you think about describing what 404 means in the test name?

Something similar to
```suggestion
    it('return null when the risk information does not exist', async () => {
```

server/services/assessRisksAndNeedsService.test.ts
```diff
@@ -40,14 +40,20 @@ describe(AssessRisksAndNeedsService, () => {
     })
 
     it('provides a userMessage on failure', async () => {
-      restClientMock.get.mockRejectedValue(createError(404))
+      restClientMock.get.mockRejectedValue(createError(409))
```
sldblog, line 43: is it worth adding a test for a 5xx branch as well?

server/views/shared/referralDetails.njk
```diff
@@ -31,7 +31,7 @@
 
       <h2 class="govuk-heading-m">Service user's risk information</h2>
 
-      {% if presenter.riskPresenter.riskSummaryEnabled %}
+      {% if presenter.riskPresenter.riskSummaryEnabled and not presenter.riskPresenter.riskSummaryNotFound %}
```
sldblog, line 34: _nit_ "NotFound" seems to be used in negatives across the board. Would it be simpler to use `riskSummaryAvailable`?

sldblog (commented):
no blockers from me, but I'll defer approval to people with better grasp at reality

docker-compose.yml
tomsmyers: sadly, we have this already all over the place. e.g. hmpps auth requires community api to allow delius logins.

server/views/shared/referralDetails.njk
tomsmyers: i count this useage in the negative to two useages without negation 😛

docker-compose.yml
sldblog: yeah, but if this keeps up, we won't be able to create a working dev env or start getting circular dependencies... need to take this forward with tooling

Gweaton (approved):
Looks good to me - I agree with David's comments re: testing though!

server/services/assessRisksAndNeedsService.test.ts
tomsmyers: done

server/services/assessRisksAndNeedsService.test.ts
tomsmyers: done

docker-compose.yml
tomsmyers: agree

PR #494, opened 2021-06-24
IC-1575: Provide PP team email and telephone number on referral details page

andrewburrowsmoj:

Gets the team details from Community API to then display on the screen for referral details.

Provides alternative contact details if PP is unavailable

## Note

image

### DIFF
--- a/server/models/delius/deliusStaffDetails.ts
+++ b/server/models/delius/deliusStaffDetails.ts
@@ -1,0 +1,9 @@
+export interface DeliusStaffDetails {
+  username: string
+  teams: DeliusTeam[]
+}
+
+export interface DeliusTeam {
+  telephone: string
+  emailAddress: string
+}
--- a/server/models/delius/deliusStaffDetails.ts
+++ b/server/models/delius/deliusStaffDetails.ts
@@ -1,9 +1,11 @@
 export interface DeliusStaffDetails {
   username: string
-  teams: DeliusTeam[]
+  teams?: DeliusTeam[]
 }
 
 export interface DeliusTeam {
-  telephone: string
-  emailAddress: string
+  telephone?: string | null
+  emailAddress?: string | null
+  startDate: string
+  endDate?: string | null
 }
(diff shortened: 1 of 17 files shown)


andrewburrowsmoj:
NOTE:
Do not merge yet - community API work still has to be merged

server/models/delius/deliusStaffDetails.ts
```diff
@@ -0,0 +1,9 @@
+export interface DeliusStaffDetails {
```
Gweaton, line 1: Do any of the fields on these interfaces need to be optional? One thing I've learned when working with the community api is that, particularly locally, all of these fields can be missing from the response, but that might not be the case here.

server/routes/shared/showReferralPresenter.test.ts
```diff
@@ -142,10 +145,9 @@ describe(ShowReferralPresenter, () => {
   })
 
   describe('probationPractitionerTeamDetails', () => {
-    it('returns a summary list of probation practitioner team details', () => {
-      const sentReferral = sentReferralFactory.build(referralParams)
-      const presenter = new ShowReferralPresenter(
-        sentReferral,
+    function callShowReferralPresenter(staffDetails: DeliusStaffDetails) {
```
Gweaton, line 148: I wonder if there's a better name for this function, to specify that the only difference is the staff details, something like `createShowReferralPresenterWithStaffDetails` might be clearer?

Gweaton (commented):
Looks good - I'm not approving yet as an extra safety net to stop this being merged, but I've left a couple of comments too.

server/models/delius/deliusStaffDetails.ts
andrewburrowsmoj: `username` should never be null given that it's used for the query.
`startDate` should never be null looking at the communityApi code it defaults the value to the current date. I'll confirm this with Rajiv if he's able to query the database.
`teams` is defaulted to an empty list in communityApi if no teams exist and can never be null.

`emailAddress` and `telephone` are nullable - but making them optional shouldn't change anything (I'll do that)

server/routes/shared/showReferralPresenter.test.ts
andrewburrowsmoj: Sure. I've used your suggestion.

server/models/delius/deliusStaffDetails.ts
andrewburrowsmoj: You were right, I've just tested with real communityApi and teams can be nullable (not sure how since the code should default it to an empty list).
I'll make that change

andrewburrowsmoj: I've now added in these changes.

Gweaton (approved):

PR #497, opened 2021-06-25
tell appinsights that only 500 errors returned from the ui should be marked as errors

tomsmyers:
tell appinsights that only 500 errors returned from the ui should be marked as errors

the reason for this is the the UI is free to serve up client error code like bad request (400) or forbidden (403), but these are not errors. we do not want these intentional non-2xx status codes being counted as errors in app insights as it is really confusing.

### DIFF
--- a/server/azureAppInsights.ts
+++ b/server/azureAppInsights.ts
@@ -37,18 +37,18 @@ function addUsernameProcessor(
   return true
 }
 
-// function errorStatusCodeProcessor(
-//   envelope: Contracts.EnvelopeTelemetry,
-//   _contextObjects: { [name: string]: unknown } | undefined
-// ): boolean {
-//   if (envelope.data.baseType === Contracts.TelemetryTypeString.Request && envelope.data.baseData !== undefined) {
-//     // eslint-disable-next-line no-param-reassign
-//     envelope.data.baseData.success = !(
-//       envelope.data.baseData.responseCode in config.applicationInsights.errorStatusCodes
-//     )
-//   }
-//   return true
-// }
+function errorStatusCodeProcessor(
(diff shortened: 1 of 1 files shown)


server/azureAppInsights.ts
```diff
@@ -64,5 +64,6 @@ export default function initialiseAppInsights(): void {
     // custom processors to fine tune behaviour
     defaultClient.addTelemetryProcessor(ignoreExcludedRequestsProcessor)
     defaultClient.addTelemetryProcessor(addUsernameProcessor)
+    defaultClient.addTelemetryProcessor(errorStatusCodeProcessor)
```
sldblog, line 67: What do you think about
```suggestion
    defaultClient.addTelemetryProcessor(markClientErrorsAsSuccessProcessor)
```

server/config.ts
```diff
@@ -52,6 +52,7 @@ export default {
     connectionString: get('APPLICATIONINSIGHTS_CONNECTION_STRING', null, requiredInProduction),
     cloudRoleName: 'interventions-ui',
     excludedRequests: [/^GET \/assets\/.*/, /^GET \/health.*/],
+    errorStatusCodes: [500],
```
sldblog, line 55: is it possible to have this over the entire range of 500..599? might be simpler to hardcode it as `envelope.data.baseData.responseCode >= 500`?

tomsmyers: good idea

server/azureAppInsights.ts
tomsmyers: left it as a generic status code processor for now, but added a comment which explains how the status code affects the success attribute of the event

sldblog (approved):

PR #513, opened 2021-06-28
IC-1987 update report a problem page to contain latest static help content

tomsmyers:
since the content is no longer dependent on the user type, i removed the view
and presenter to reduce complexity and just have all the html for this page
in the template.

Screenshot 2021-06-28 at 17 28 51


server/views/common/reportAProblem.njk
```diff
@@ -1,6 +1,41 @@
...
+
+      <h2 class="govuk-heading-m">Suppliers</h2>
+
+      <p class="govuk-body">To report a problem with this digital service, please contact your helpdesk who will contact the digital service team if necessary. </p>
+      <p class="govuk-body">If your organisation does not have an IT helpdesk, please contact the helpdesk on <a href="tel:+44800 917 5148">0800 917 5148</a></p>
+
+      <h2 class="govuk-heading-m">Learning and resources for NPS staff</h2>
+
+      <ul class="govuk-list govuk-list--bullet">
+        <li><a class="govuk-link" href="https://intranet.noms.gsi.gov.uk/corporate/probation-changes/Commissioned-Rehabilitative-Services/your-learning-probation-practitioners-and-contract-managers">NPS Intranet Learning Section</a></li>
```
Gweaton, line 20: Do we want these links to open in new tabs (ie. should we add `target="_blank" rel="noopener noreferrer"` to them?)

Gweaton (approved):
One question

server/views/common/reportAProblem.njk
tomsmyers: last time i had that discussion it was discouraged.

PR #516, opened 2021-06-29
IC-1116: Add GetServiceProviderReportingData function and contracts

Gweaton:

Adds a function (with contract tests & mocks) to fetch full referral reporting data for a Service Provider organisation.

The data returned from this will be used to generate a CSV for the Service Provider to download data on all of their referrals.

### DIFF
--- a/server/services/interventionsService.test.ts
+++ b/server/services/interventionsService.test.ts
@@ -14,6 +14,7 @@ import endOfServiceReportFactory from '../../testutils/factories/endOfServiceRep
 import sentReferralFactory from '../../testutils/factories/sentReferral'
 import appointmentFactory from '../../testutils/factories/appointment'
 import supplierAssessmentFactory from '../../testutils/factories/supplierAssessment'
+import CalendarDay from '../utils/calendarDay'
 
 jest.mock('../services/hmppsAuthService')
 
@@ -2910,6 +2911,60 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
     })
   })
   */
+
+  describe('getServiceProviderReportingData', () => {
+    const reportingResponse = [
+      {
+        referralLink: 'https://refer-and-monitor.example/referral/c9f9e22f-ddd9-422a-a9af-76fc02b18b38',
+        referralRef: 'SM1973AC',
+        referralId: 'c9f9e22f-ddd9-422a-a9af-76fc02b18b38',
+        contractId: 'some-contract-id',
+        organisationId: 'XYZ5678',
+        referringOfficerEmail: 'bernard.beaks@justice.gov.uk',
+        caseworkerId: 'liane.supplier@example.com',
+        serviceUserCrn: 'X017844',
+        dateReferralReceived: '2021-06-27T14:49:01+01:00',
+        dateSaaBooked: '2021-06-27T14:49:01+01:00',
+        dateSaaAttended: '2021-06-27T14:49:01+01:00',
+        dateFirstActionPlanSubmitted: '2021-06-27T14:49:01+01:00',
+        dateOfFirstActionPlanApproval: '2021-06-27T14:49:01+01:00',
+        dateOfFirstAttendedSession: '2021-06-27T14:49:01+01:00',
+        outcomesToBeAchievedCount: 8,
+        outcomesAchieved: 5.5,
+        countOfSessionsExpected: 10,
+        countOfSessionsAttended: 6,
+        endRequestedByPPAt: '2021-06-27T14:49:01+01:00',
+        endRequestedByPPReason: 'REC',
+        dateEOSRSubmitted: '2021-06-27T14:49:01+01:00',
+        concludedAt: '2021-06-27T14:49:01+01:00',
+      },
+    ]
+
+    it('returns a list of referrals for with data for reporting', async () => {
+      await provider.addInteraction({
+        state: 'nothing',
+        uponReceiving: 'a GET request for referral reporting data',
+        withRequest: {
+          method: 'GET',
+          path: '/performance-report',
+          query: { fromIncludingDate: '2021-06-01', toIncludingDate: '2021-06-10' },
+          headers: { Accept: 'application/json', Authorization: `Bearer ${token}` },
+        },
+        willRespondWith: {
+          status: 200,
+          body: Matchers.like(reportingResponse),
+          headers: { 'Content-Type': 'application/json' },
+        },
+      })
+
+      expect(
+        await interventionsService.getServiceProviderReportingData(token, {
+          fromIncludingDate: CalendarDay.fromComponents(1, 6, 2021)!,
+          toIncludingDate: CalendarDay.fromComponents(10, 6, 2021)!,
+        })
+      ).toMatchObject(reportingResponse)
+    })
+  })
 })
 
 describe('serializeDeliusServiceUser', () => {
--- a/server/services/interventionsService.ts
+++ b/server/services/interventionsService.ts
@@ -550,6 +550,11 @@ export default class InterventionsService {
         toIncludingDate: reportDates.toIncludingDate.iso8601,
       },
       headers: { Accept: 'application/json' },
+      // this will likely be a very large request, so we're overriding the timeout to play it safe
+      timeout: {
+        deadline: 120000,
+        response: 120000,
+      },
     })) as ServiceProviderReportReferral[]
   }
 }
--- a/server/services/interventionsService.ts
+++ b/server/services/interventionsService.ts
@@ -21,6 +21,7 @@ import ReferralDesiredOutcomes from '../models/referralDesiredOutcomes'
 import ReferralComplexityLevel from '../models/referralComplexityLevel'
 import Appointment from '../models/appointment'
 import SupplierAssessment from '../models/supplierAssessment'
+import { ServiceProviderReportReferral } from '../models/serviceProviderReportReferral'
 
 export interface InterventionsServiceValidationError {
   field: string
@@ -60,6 +61,11 @@ export interface CreateEndOfServiceReportOutcome {
 export interface UpdateDraftEndOfServiceReportParams {
   furtherInformation: string | null
   outcome: CreateEndOfServiceReportOutcome | null
+}
+
+export interface CreateReportDateParams {
+  fromIncludingDate: CalendarDay
+  toIncludingDate: CalendarDay
 }
 
 export default class InterventionsService {
@@ -530,4 +536,20 @@ export default class InterventionsService {
     })) as Appointment
   }
  */
+
+  async getServiceProviderReportingData(
+    token: string,
+    reportDates: CreateReportDateParams
+  ): Promise<ServiceProviderReportReferral[]> {
+    const restClient = this.createRestClient(token)
+
+    return (await restClient.get({
+      path: '/performance-report',
+      query: {
+        fromIncludingDate: reportDates.fromIncludingDate.iso8601,
+        toIncludingDate: reportDates.toIncludingDate.iso8601,
+      },
+      headers: { Accept: 'application/json' },
+    })) as ServiceProviderReportReferral[]
+  }
 }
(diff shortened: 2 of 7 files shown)


server/services/interventionsService.ts
```diff
@@ -516,4 +522,16 @@ export default class InterventionsService {
...
  */
+
+  async getServiceProviderReportingData(
+    token: string,
+    reportDates: CreateReportDateParams
+  ): Promise<ServiceProviderReportReferral[]> {
+    const restClient = this.createRestClient(token)
+
+    return (await restClient.get({
+      path: `/performance-report?from=${reportDates.fromDate}&to=${reportDates.toDate}`,
```
tomsmyers, line 533: `get` takes a query argument which is a bit neater

```
query: { from: ..., to: ... }
```

server/services/interventionsService.ts
```diff
@@ -62,6 +63,11 @@ export interface UpdateDraftEndOfServiceReportParams {
   outcome: CreateEndOfServiceReportOutcome | null
 }
 
+export interface CreateReportDateParams {
+  fromDate: string | null
```
tomsmyers, line 67: what happens if either of these is null?

server/services/interventionsService.ts
Gweaton: Ah yeah, I initially had this as a `post` which didn't but forgot to update this after switching to GET

Gweaton: Have updated this.

server/services/interventionsService.ts
Gweaton: The `null` behaviour is taken from my other branch that uses this function, but I think I might refactor it so it's not necessary. Have updated this to remove the `null`s.

server/services/interventionsService.test.ts
```diff
@@ -2846,6 +2846,60 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
     })
   })
   */
+
+  describe('getServiceProviderReportingData', () => {
+    const reportingResponse = [
+      {
+        referralLink: 'https://refer-and-monitor.com/referral/c9f9e22f-ddd9-422a-a9af-76fc02b18b38',
```
sldblog, line 2853: We have the opportunity to simplify by not expecting this from the API

_nit_ it's recommended to use `domain.example` and similar hostnames in case it ever gets called by accident (https://en.wikipedia.org/wiki/.example)

server/services/interventionsService.test.ts
```diff
@@ -2846,6 +2846,60 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
...
+          status: 200,
+          body: Matchers.like(reportingResponse),
+          headers: { 'Content-Type': 'application/json' },
+        },
+      })
+
+      expect(
+        await interventionsService.getServiceProviderReportingData(token, {
+          fromDate: '2021-06-01',
+          toDate: '2021-06-10',
```
sldblog, line 2898: `from`/`to` can be ambiguous to interpret -- `fromIncludingDate`, `toExcludingDate` might be explicit

server/services/interventionsService.test.ts
```diff
@@ -2846,6 +2846,60 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
...
+        uponReceiving: 'a GET request for referral reporting data',
+        withRequest: {
+          method: 'GET',
+          path: '/performance-report',
+          query: { from: Matchers.like('2021-06-01'), to: Matchers.like('2021-06-10') },
+          headers: { Accept: 'application/json', Authorization: `Bearer ${token}` },
+        },
+        willRespondWith: {
+          status: 200,
+          body: Matchers.like(reportingResponse),
```
sldblog, line 2890: a free-form date range query might put quite a load on the server to serve this and quite a payload
can we monitor its performance explicitly/have its own bigger timeout?

server/services/interventionsService.test.ts
Gweaton: Good point, would you expect the same for the query params?

Gweaton: Have updated the query params too for that reason

server/services/interventionsService.test.ts
Gweaton: It's a little more fiddly on the UI side to construct this due to the way we rely on Typescript assertions to essentially deserialize our objects for us. I can have a go, but I feel it's a bit simpler to send this in the response body so we can pipe the whole response straight into a CSV, rather than prepending this field. What do you think?

In the meantime, I've updated the link with `.example`.

Gweaton:
After a chat with @mahalrmoj and @rgforsyth we've decided to go for a streaming approach which should address @sldblog's concerns about payload size. Converting this to draft for now.

server/services/interventionsService.test.ts
Gweaton: I've pushed a new commit for this to increase the timeout for this request to 2 minutes, do you think that's enough?

I wonder if we should restrict the number of months/days that you can generate a report for, but that could come further down the line when we have more months' worth of referrals?

server/services/interventionsService.test.ts
```diff
@@ -2846,6 +2846,59 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
...
+    const reportingResponse = [
+      {
+        referralLink: 'https://refer-and-monitor.example/referral/c9f9e22f-ddd9-422a-a9af-76fc02b18b38',
+        referralRef: 'SM1973AC',
+        referralId: 'c9f9e22f-ddd9-422a-a9af-76fc02b18b38',
+        contractId: 'thing',
+        organisationId: 'XYZ5678',
+        caseworkerId: 'liane.supplier@example.com',
+        serviceUserCrn: 'X017844',
+        dateReferralReceived: '2021-06-27T14:49:01+01:00',
```
lawrence-forooghian, line 2924: I'd expect a property whose name begins with `date` to be a date - i.e. no time components. Would it be more consistent to call these `referralReceivedAt`, etc? Or does this come down to the field names not being changeable in the CSV, and not wanting to do any mappings on the frontend?

server/services/interventionsService.test.ts
```diff
@@ -2846,6 +2846,59 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
...
+    ]
+
+    it('returns a list of referrals for with data for reporting', async () => {
+      await provider.addInteraction({
+        state: 'nothing',
+        uponReceiving: 'a GET request for referral reporting data',
+        withRequest: {
+          method: 'GET',
+          path: '/performance-report',
+          query: { fromIncludingDate: Matchers.like('2021-06-01'), toIncludingDate: Matchers.like('2021-06-10') },
```
lawrence-forooghian, line 2948: Why do we need the `Matchers` here? Aren't these the exact values that we expect to be sent in the query, given the parameters passed to `getServiceProviderReportingData` below?

server/services/interventionsService.ts
```diff
@@ -62,6 +63,11 @@ export interface UpdateDraftEndOfServiceReportParams {
   outcome: CreateEndOfServiceReportOutcome | null
 }
 
+export interface CreateReportDateParams {
+  fromIncludingDate: string
```
lawrence-forooghian, line 67: Can we use `CalendarDay` for these values instead of `string`, and then call `.iso8601` on those inside `getServiceProviderReportingData`? That way callers don't need to know how they're meant to serialize the date values, and it's clear that the from / to are dates, not datetimes.

lawrence-forooghian (commented):
A few comments from me from a UI point of view, but I'll leave the approval to the backend folks.

server/services/interventionsService.test.ts
Gweaton: Yeah, for this I took all the headers Leigh supplied in the csv schema as I thought we could just use the fields as-is. However, you might have already seen in the next PR that we can actually map them to specific headers (I've just snake_cased everything), so I can update this object if you think it makes more sense? I guess I wanted to map them to the given schema as closely as possible.

server/services/interventionsService.test.ts
Gweaton: I think I thought this was relevant for the Provider tests (ie. so they could put any value in on their side and not be tied to the dates we've put in here) - I think I'm misunderstanding that, right? Is this more useful for us for things like values that might depend on a random value like a uuid that we don't really care about?

server/services/interventionsService.ts
Gweaton: Yep, sounds good

server/services/interventionsService.test.ts
Gweaton: I've updated this

lawrence-forooghian: I think that we usually use the matchers when making assertions about things that we're not fully in control of - e.g. some ID in a response, a timestamp that depends on the moment the request is made, etc. We use them to say "we expect to get some data that looks like this". But here, we are fully in control of how this request is constructed - the `fromIncludingDate` and `toIncludingDate` are just what we pass to the `getServiceProviderReportingData` method in the test.

server/services/interventionsService.test.ts
lawrence-forooghian: Yeah, I'm not too bothered either way. Since we're not doing anything on the frontend other than converting to CSV, we can just pass them through I think.

lawrence-forooghian:
Thanks for the changes, this one looks good from a UI perspective to me now.

lawrence-forooghian:
@rgforsyth please could you take a look at the contracts here and approve if happy?

server/services/interventionsService.test.ts
```diff
@@ -2910,6 +2911,59 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
...
   })
   */
+
+  describe('getServiceProviderReportingData', () => {
+    const reportingResponse = [
+      {
+        referralLink: 'https://refer-and-monitor.example/referral/c9f9e22f-ddd9-422a-a9af-76fc02b18b38',
+        referralRef: 'SM1973AC',
+        referralId: 'c9f9e22f-ddd9-422a-a9af-76fc02b18b38',
+        contractId: 'thing',
```
rgforsyth, line 2921: May just be dummy data but this should be a uuid right?

server/services/interventionsService.test.ts
```diff
@@ -2910,6 +2911,59 @@ pactWith({ consumer: 'Interventions UI', provider: 'Interventions Service' }, pr
...
+        organisationId: 'XYZ5678',
+        caseworkerId: 'liane.supplier@example.com',
```
rgforsyth, line 2923: Also a uuid?

rgforsyth (approved):
LGTM. 
I'm not sure if it matters, but some of the example data looked to be slightly different to what I was expecting.. Random sting rather than uuid?

server/services/interventionsService.test.ts
lawrence-forooghian: Looking at the v1.0.0 schema on the Excel file on [IC-1116](https://dsdmoj.atlassian.net/browse/IC-1116), the example value is `liane.supplier@example.com`.

server/services/interventionsService.test.ts
lawrence-forooghian: I don't think I can answer that one. In the v1.0.0 schema (which I linked to in another comment here) it says "Format of string TBC". There is however a [comment from @sldblog](https://dsdmoj.atlassian.net/browse/IC-1116?focusedCommentId=161187) saying that this perhaps should be called `contract_reference`, since the UUID will have no meaning to the suppliers. I'm not sure if it's too late to make those changes now, though.

lawrence-forooghian:
@rgforsyth following our conversation with Leigh, I've:

- added a `referringOfficerEmail` value here, too (in my example it's an email address but my understanding is it might initially be blank in the API response)
- changed `contractId`'s example to make it clear it's a placeholder value because we don't know what the real thing will look like
- kept `caseworkerId` as an email address

lawrence-forooghian (approved):

PR #518, opened 2021-06-30
IC-1118: Add UI for downloading SP performance report

Gweaton:

Adds a new "reporting" page (linked via a new SP dashboard nav) allowing users to download a Report of referral data between two dates.

This won't work until we have the API changes (contracts for which are in #516).

To allow SPs to download performance data for reporting.

### Banner

image

### Reporting page

image

### DIFF
--- a/integration_tests/integration/serviceProviderReferrals.spec.js
+++ b/integration_tests/integration/serviceProviderReferrals.spec.js
@@ -1331,4 +1331,74 @@ describe('Service provider referrals dashboard', () => {
       cy.get('#supplier-assessment-status').contains(/^\s*scheduled\s*$/)
     })
   })
+
+  describe('User downloads reporting data', () => {
+    it('allows the user to download a CSV based on the requested date range', () => {
+      const intervention = interventionFactory.build()
+
+      const sentReferrals = [
+        sentReferralFactory.build({
+          referral: {
+            interventionId: intervention.id,
+          },
+        }),
+        sentReferralFactory.build({
+          referral: {
+            interventionId: intervention.id,
+          },
+        }),
+      ]
+
+      const deliusUser = deliusUserFactory.build()
+
+      cy.stubGetIntervention(intervention.id, intervention)
+      sentReferrals.forEach(referral => cy.stubGetSentReferral(referral.id, referral))
+      cy.stubGetSentReferralsForUserToken(sentReferrals)
+      cy.stubGetUserByUsername(deliusUser.username, deliusUser)
+      cy.stubGetServiceProviderReportingData([
+        {
+          referralLink: 'https://refer-and-monitor.example/referral/c9f9e22f-ddd9-422a-a9af-76fc02b18b38',
+          referralRef: 'SM1973AC',
+          referralId: 'c9f9e22f-ddd9-422a-a9af-76fc02b18b38',
+          contractId: 'thing',
+          organisationId: 'XYZ5678',
+          referringOfficerEmail: 'joe.probation@example.com',
+          caseworkerId: 'liane.supplier@example.com',
+          serviceUserCrn: 'X017844',
+          dateReferralReceived: '2021-06-27T14:49:01+01:00',
+          dateSaaBooked: '2021-06-27T14:49:01+01:00',
+          dateSaaAttended: '2021-06-27T14:49:01+01:00',
+          dateFirstActionPlanSubmitted: '2021-06-27T14:49:01+01:00',
+          dateOfFirstActionPlanApproval: '2021-06-27T14:49:01+01:00',
+          dateOfFirstAttendedSession: '2021-06-27T14:49:01+01:00',
+          outcomesToBeAchievedCount: 8,
+          outcomesAchieved: 5.5,
+          countOfSessionsExpected: 10,
+          countOfSessionsAttended: 6,
+          endRequestedByPPAt: '2021-06-27T14:49:01+01:00',
+          endRequestedByPPReason: 'REC',
+          dateEOSRSubmitted: '2021-06-27T14:49:01+01:00',
+          concludedAt: '2021-06-27T14:49:01+01:00',
+        },
+      ])
+
+      cy.login()
+
+      cy.contains('Reporting').click()
+      cy.contains('Reporting')
+
+      cy.get('#from-date-day').clear().type('10')
+      cy.get('#from-date-month').clear().type('6')
+      cy.get('#from-date-year').clear().type('2021')
+      cy.get('#to-date-day').clear().type('27')
+      cy.get('#to-date-month').clear().type('6')
+      cy.get('#to-date-year').clear().type('2021')
+
+      // TODO (IC-2040)
+      // Unfortunately there's currently no way to test non 'text/html' files in Cypress.
+      // https://github.com/cypress-io/cypress/issues/1551
+      // When attempting to download the generated CSV, the page hangs.
+      // I'm leaving this test here as-is, in case Cypress is updated and we can test this behaviour.
+    })
+  })
 })
--- a/server/routes/serviceProviderReferrals/reportingForm.test.ts
+++ b/server/routes/serviceProviderReferrals/reportingForm.test.ts
@@ -1,0 +1,176 @@
+import TestUtils from '../../../testutils/testUtils'
+import CalendarDay from '../../utils/calendarDay'
+import ReportingForm from './reportingForm'
+
+describe(ReportingForm, () => {
+  describe('data', () => {
+    describe('with valid data', () => {
+      it('returns an object with two Calendar Day values', async () => {
+        const request = TestUtils.createRequest({
+          'from-date-year': '2021',
+          'from-date-month': '06',
+          'from-date-day': '12',
+          'to-date-year': '2021',
+          'to-date-month': '06',
+          'to-date-day': '13',
+        })
+
+        const result = await new ReportingForm(request).data()
+
+        expect(result.value!.fromIncludingDate).toEqual(CalendarDay.fromComponents(12, 6, 2021))
+        expect(result.value!.toIncludingDate).toEqual(CalendarDay.fromComponents(13, 6, 2021))
+        expect(result.error).toBeNull()
+      })
+    })
+
+    describe('with invalid data', () => {
+      describe('when the "from" date is after the "to" date', () => {
+        it('returns an error object with a message', async () => {
+          const request = TestUtils.createRequest({
+            'from-date-year': '2021',
+            'from-date-month': '06',
+            'from-date-day': '12',
+            'to-date-year': '2021',
+            'to-date-month': '06',
+            'to-date-day': '11',
+          })
+
+          const result = await new ReportingForm(request).data()
+
+          expect(result.value).toBeNull()
+          expect(result.error!.errors).toEqual([
+            {
+              errorSummaryLinkedField: 'from-date-day',
+              formFields: ['from-date-day', 'from-date-month', 'from-date-year'],
+              message: 'The "from" date must be before the "to" date',
+            },
+
+            {
+              errorSummaryLinkedField: 'to-date-day',
+              formFields: ['to-date-day', 'to-date-month', 'to-date-year'],
+              message: 'The "from" date must be before the "to" date',
+            },
+          ])
+        })
+      })
+
+      describe('when the "from" date is before Day One (01/06/2021)', () => {
+        it('returns an error object with a message', async () => {
+          const request = TestUtils.createRequest({
+            'from-date-year': '2021',
+            'from-date-month': '05',
+            'from-date-day': '12',
+            'to-date-year': '2021',
+            'to-date-month': '06',
+            'to-date-day': '11',
+          })
+
+          const result = await new ReportingForm(request).data()
+
+          expect(result.value).toBeNull()
+          expect(result.error!.errors).toEqual([
+            {
+              errorSummaryLinkedField: 'from-date-day',
+              formFields: ['from-date-day', 'from-date-month', 'from-date-year'],
+              message: 'Data before 01 June 2021 is not available',
+            },
+          ])
+        })
+      })
+
+      describe('for dates in the future, for which we have no historical data', () => {
+        const today = new Date()
+        const tomorrow = new Date(today)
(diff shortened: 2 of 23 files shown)


server/utils/serviceProviderCSVGenerator.test.ts
```diff
@@ -0,0 +1,53 @@
...
+        )
+        expect(ServiceProviderCSVGenerator.generate(reportingResponse)).toContain(
+          'https://refer-and-monitor.com/referral/c9f9e22f-ddd9-422a-a9af-76fc02b18b38,SM1973AC,c9f9e22f-ddd9-422a-a9af-76fc02b18b38,thing,XYZ5678,joe.probation@example.com,liane.supplier@example.com,X017844,2021-06-27T14:49:01+01:00,2021-06-27T14:49:01+01:00,2021-06-27T14:49:01+01:00,2021-06-27T14:49:01+01:00,2021-06-27T14:49:01+01:00,2021-06-27T14:49:01+01:00,8,5.5,10,6,2021-06-27T14:49:01+01:00,REC,2021-06-27T14:49:01+01:00,2021-06-27T14:49:01+01:00'
+        )
+      })
+    })
+  })
+
+  describe('.createFilename', () => {
+    it('returns a filename including the date the report is generated', () => {
```
lawrence-forooghian, line 45: We should be explicit about what time zone is being used to calculate the date here. Probably British time is fine.

server/utils/serviceProviderCSVGenerator.test.ts
```diff
@@ -0,0 +1,53 @@
...
+      const mockDate = new Date('2021-06-28')
```
lawrence-forooghian, line 46: Can we be more specific when constructing this date, so it's clear exactly which moment in time this is referring to? e.g. `new Date('2021-06-28T10:00:00Z')`

server/utils/serviceProviderCSVGenerator.ts
```diff
@@ -0,0 +1,54 @@
...
+    reportReferrals.forEach(referral => {
+      stringifier.write(referral)
+    })
+    stringifier.end()
+
+    return data.join('')
+  }
+
+  static createFilename(): string {
+    return `refer-monitor-intervention_report_${new Date().toLocaleDateString('en-CA').replace(/-/g, '')}.csv`
```
lawrence-forooghian, line 52: `en-CA` is a bit confusing here, I think. You should be able to use `CalendarDay` to do this, e.g.

```suggestion
    return `refer-monitor-intervention_report_${CalendarDay.atTimeInBritain(new Date()).iso8601.replace(/-/g, '')}.csv
```

and this'll also handle the time zone comment I left above.

server/utils/serviceProviderCSVGenerator.ts
```diff
@@ -0,0 +1,54 @@
...
+      while (row) {
+        data.push(row)
+        row = stringifier.read()
+      }
+    })
+
+    reportReferrals.forEach(referral => {
+      stringifier.write(referral)
+    })
+    stringifier.end()
```
lawrence-forooghian, line 46: I'm a little confused by how this method is working. As far as I can tell, this streaming API provided by `stringifier` is an _asynchronous_ API - you:

1. give it a bunch of data (`stringifier.write()`)
2. tell it you've given it all the data that you've got to offer (`stringifier.end()`)

and then, over time, it incrementally generates the CSV and calls you back with a handful of lines at a time (`stringifier.on('readable', …)`). It then eventually tells you that it's done via another callback (I think that's `stringifier.on('finish', …)`, seen in [the library's docs](https://csv.js.org/stringify/api/)), but I don't see us using that callback here.

So, if this whole process is asynchronous, then I think that the method which wraps this process also needs to be asynchronous. It feels like the signature for this method should be 

```typescript
static async generate(reportReferrals: ServiceProviderReportReferral[]): Promise<string>
```

and we need to return a promise which we resolve upon receiving the `finish` callback.

On the other hand, this does seem to be working, at least for a single row in the tests, but my feeling is that if we're not using the API as intended then it could stop working at any time or not work for larger data sets.

server/utils/serviceProviderCSVGenerator.test.ts
```diff
@@ -0,0 +1,53 @@
...
+            concludedAt: '2021-06-27T14:49:01+01:00',
+          },
+        ]
+
+        expect(ServiceProviderCSVGenerator.generate(reportingResponse)).toContain(
+          'referral_link,referral_ref,referral_id,contract_id,organisation_id,referring_officer_email,caseworker_id,service_user_crn,date_referral_received,date_saa_booked,date_saa_attended,date_first_action_plan_submitted,date_of_first_action_plan_approval,date_of_first_attended_session,outcomes_to_be_achieved_count,outcomes_achieved,count_of_sessions_expected,count_of_sessions_attended,end_requested_by_pp_at,end_requested_by_pp_reason,date_eosr_submitted,concluded_at'
+        )
+        expect(ServiceProviderCSVGenerator.generate(reportingResponse)).toContain(
+          'https://refer-and-monitor.com/referral/c9f9e22f-ddd9-422a-a9af-76fc02b18b38,SM1973AC,c9f9e22f-ddd9-422a-a9af-76fc02b18b38,thing,XYZ5678,joe.probation@example.com,liane.supplier@example.com,X017844,2021-06-27T14:49:01+01:00,2021-06-27T14:49:01+01:00,2021-06-27T14:49:01+01:00,2021-06-27T14:49:01+01:00,2021-06-27T14:49:01+01:00,2021-06-27T14:49:01+01:00,8,5.5,10,6,2021-06-27T14:49:01+01:00,REC,2021-06-27T14:49:01+01:00,2021-06-27T14:49:01+01:00'
```
lawrence-forooghian, line 38: Could we perhaps test this works with a large number of rows, too? Maybe we could just assert the presence of the IDs or something, for the other rows.

server/utils/forms/inputs/calendarDayInput.test.ts
```diff
@@ -145,4 +145,18 @@ describe(CalendarDayInput, () => {
       })
```
lawrence-forooghian, line 145: I think the commit message here is out of date:

```
in the upcoming `ReportingInput` class
```

— should be `ReportingForm` now?

server/utils/forms/inputs/calendarDayInput.ts
```diff
@@ -37,22 +41,28 @@ export default class CalendarDayInput {
     }
   }
 
+  static createErrors(key: string, message: string): FormValidationError {
```
lawrence-forooghian, line 44:
```suggestion
  static createError(key: string, message: string): FormValidationError {
```

server/routes/serviceProviderReferrals/reportingForm.test.ts
```diff
@@ -0,0 +1,154 @@
...
+          'from-date-month': '06',
+          'from-date-day': '12',
+          'to-date-year': '2021',
+          'to-date-month': '06',
+          'to-date-day': '13',
+        })
+
+        const result = await new ReportingForm(request).data()
+
+        expect(result.reportingDateRange!.fromIncludingDate).toEqual('2021-06-12')
```
lawrence-forooghian, line 20: I reckon (see [my comment on #516](https://github.com/ministryofjustice/hmpps-interventions-ui/pull/516#discussion_r662350752)) that these should be `CalendarDay` instances.

server/routes/serviceProviderReferrals/reportingForm.test.ts
```diff
@@ -0,0 +1,154 @@
...
+            'from-date-day': '12',
+            'to-date-year': '2021',
+            'to-date-month': '06',
+            'to-date-day': '11',
+          })
+
+          const result = await new ReportingForm(request).data()
+
+          expect(result.reportingDateRange).toBeNull()
+          expect(result.error!.errors).toContainEqual({
```
lawrence-forooghian, line 41: Is there a reason this isn't `toEqual` with a single-valued array? Maybe it would be good to check that we aren't getting other spurious errors here, only the one we expect. Ditto throughout this test file.

server/routes/serviceProviderReferrals/reportingForm.test.ts
```diff
@@ -0,0 +1,154 @@
...
+          const request = TestUtils.createRequest({
+            'from-date-year': tomorrowYear,
+            'from-date-month': tomorrowMonth,
+            'from-date-day': tomorrowDay,
+            'to-date-year': '2021',
+            'to-date-month': '06',
+            'to-date-day': '12',
+          })
+
+          const result = await new ReportingForm(request).data()
```
lawrence-forooghian, line 90: It would be easier if you could lock the current time to a known moment here instead of having to do all this formatting logic in the tests. e.g. if you could either pass a `now: Date` instance into the constructor or else stub out `new Date()` (I prefer the former) so that it always returns `new Date('2021-10-21T10:00:00Z')` then you can just check that the error mesage is "Data after 21 October 2021 is not available".

(Although, that said, mightn't it actually be clearer to just say "Data after today is not available"?)

server/routes/serviceProviderReferrals/reportingForm.ts
```diff
@@ -0,0 +1,73 @@
...
+
+interface ReportingDateRangeParams {
+  error: FormValidationError | null
+  reportingDateRange: CreateReportDateParams | null
+}
+
+export default class ReportingForm {
+  constructor(private readonly request: Request) {}
+
+  async data(): Promise<ReportingDateRangeParams> {
```
lawrence-forooghian, line 15: We usually use `FormValidationResult` here, since it expresses that you can have precisely one of `error` or `value` being non-null.

server/utils/errorMessages.ts
```diff
@@ -19,6 +21,28 @@ export default {
...
   },
+  reportingDate: {
+    from: {
+      dayEmpty: 'The "from" date must include a day',
+      monthEmpty: 'The "from" date must include a month',
+      yearEmpty: 'The "from" date must include a year',
+      invalidDate: 'The "from" date must be a real date',
+      mustBeAfterDayOne: 'Data before 01 June 2021 is not available',
+      mustBeBeforeToDate: 'The "from" date must be before the "to" date',
+      mustNotBeInFuture: `Data after ${DateUtils.getDateStringFromDateTimeString(
```
lawrence-forooghian, line 32: I'm not quite sure how this works in JS, but isn't this whole exported object only evaluated once, when the server is started? So this would always return the date that the server was started.

integration_tests/integration/serviceProviderReferrals.spec.js
```diff
@@ -1309,4 +1309,73 @@ describe('Service provider referrals dashboard', () => {
...
+      cy.get('#from-date-month').clear().type('6')
+      cy.get('#from-date-year').clear().type('2021')
+      cy.get('#to-date-day').clear().type('27')
+      cy.get('#to-date-month').clear().type('6')
+      cy.get('#to-date-year').clear().type('2021')
+
+      // Unfortunately there's currently no way to test non 'text/html' files in Cypress.
+      // https://github.com/cypress-io/cypress/issues/1551
+      // When attempting to download the generated CSV, the page hangs.
+      // I'm leaving this test here as-is, in case Cypress is updated and we can test this behaviour.
```
lawrence-forooghian, line 1378: Can we make a tech debt Jira issue for this and mention it here?

server/routes/serviceProviderReferrals/serviceProviderReferralsController.test.ts
```diff
@@ -1794,3 +1794,73 @@ describe('GET /service-provider/referrals/:id/supplier-assessment', () => {
...
+      .type('form')
+      .send({
+        'from-date-day': '20',
+        'from-date-month': '06',
+        'from-date-year': '2021',
+        'to-date-day': '25',
+        'to-date-month': '06',
+        'to-date-year': '2021',
+      })
+      .expect(200)
```
lawrence-forooghian, line 1851: Perhaps worth checking the `Content-Disposition` header, too?

lawrence-forooghian (commented):
Lots of work here and looks good! I think we might need to revisit the details of how we generate the CSV, which I've left comments about. Let me know if you want to look at it together.

server/utils/serviceProviderCSVGenerator.test.ts
```diff
@@ -0,0 +1,53 @@
...
+            referralRef: 'SM1973AC',
+            referralId: 'c9f9e22f-ddd9-422a-a9af-76fc02b18b38',
+            contractId: 'thing',
+            organisationId: 'XYZ5678',
+            referringOfficerEmail: 'joe.probation@example.com',
+            caseworkerId: 'liane.supplier@example.com',
+            serviceUserCRN: 'X017844',
+            dateReferralReceived: '2021-06-27T14:49:01+01:00',
+            dateSAABooked: '2021-06-27T14:49:01+01:00',
+            dateSAAAttended: '2021-06-27T14:49:01+01:00',
```
lawrence-forooghian, line 19: I think we should also test how the CSV looks when some of the nullable columns are `null`.

server/utils/serviceProviderCSVGenerator.test.ts
Gweaton: Have added a factory for this and added a couple more rows.

server/routes/serviceProviderReferrals/reportingForm.test.ts
Gweaton: I think saying "Data after today is not available" is actually a fair bit clearer (otherwise it makes it sound like the given date is special for some reason). I'm gonna change it to this. It also has the side effect of making the code less complex and resolving two of your comments in one 🥇

server/utils/serviceProviderCSVGenerator.ts
lawrence-forooghian: I tried increasing the length of the array in the test to ~200 and it stops working - just outputs an empty string. I'll take a look into this, but presumably related to the above.

lawrence-forooghian (approved):

lawrence-forooghian:
Upon updating the contract tests to use the factory that the PR adds, noticed we have an inconsistency in our capitalisation of `serviceProviderCrn/CRN`. Asked @rgforsyth which is preferred.

PR #520, opened 2021-06-30
Ensure that the referral details page is still rendered even if getStaffDetails from communityAPI returns NOT_FOUND

andrewburrowsmoj:

Ensure that the referral details page is still rendered even if getStaffDetails from communityAPI returns NOT_FOUND

No all users can be considered a staff so we need to simply not display the team details if this is the case.

### DIFF
--- a/server/routes/probationPractitionerReferrals/probationPractitionerReferralsController.test.ts
+++ b/server/routes/probationPractitionerReferrals/probationPractitionerReferralsController.test.ts
@@ -427,6 +427,17 @@ describe('GET /probation-practitioner/referrals/:id/details', () => {
         })
     })
   })
+  describe('when no staff details can be found', () => {
+    it('does not show the team details', async () => {
+      communityApiService.getStaffDetails.mockResolvedValue(null)
+      await request(app)
+        .get(`/probation-practitioner/referrals/${sentReferral.id}/details`)
+        .expect(200)
+        .expect(res => {
+          expect(res.text).not.toContain('Team contact details')
+        })
+    })
+  })
 })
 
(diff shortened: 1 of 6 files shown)


server/routes/probationPractitionerReferrals/probationPractitionerReferralsController.ts
```diff
@@ -129,7 +130,12 @@ export default class ProbationPractitionerReferralsController {
         this.communityApiService.getConvictionById(crn, sentReferral.referral.relevantSentenceId),
         this.assessRisksAndNeedsService.getSupplementaryRiskInformation(sentReferral.supplementaryRiskId, accessToken),
         this.assessRisksAndNeedsService.getRiskSummary(crn, accessToken),
-        this.communityApiService.getStaffDetails(sentReferral.sentBy.username),
+        new Promise<DeliusStaffDetails | null>(resolve => {
```
tomsmyers, line 133: nit: perhaps it would be easier to just have the service method return a null value on 404?

tomsmyers: this way we wouldn't have to duplicate this code in the other controller too.

tomsmyers (approved):
thanks andrew - we should check in with the team about whether or not we want to display _something_ if this information is missing.

server/routes/probationPractitionerReferrals/probationPractitionerReferralsController.ts
andrewburrowsmoj: Yes, I realised that when I pushed .
I'm just writing one more commit for an additional test

PR #522, opened 2021-06-30
IC-2012: Rename {Intervention Type} referral to {Intervention Type} intervention on completion date page for make a referral journey

andrewburrowsmoj:
…

Renames the wording of Referral to Intervention on completion date for make a referral.

A lot of users are complaining that they think this sentence means when the work should start not when it finishes.

Wording is as intervention makes it more clear that it's tied to the actual intervention rather than when the referral is made.

image


tomsmyers (approved):
thanks andrew

PR #524, opened 2021-06-30
Handle delius teams with missing start dates.

tomsmyers:

Handle delius teams with missing start dates.

Assume that these teams are older than any team that does have a start date

allows the referral details page to load when teams come back with no start date.


sldblog (approved):

sldblog:
would be good to add tests, but live service first

PR #531, opened 2021-07-01
IC-1996: Use session-based drafts for multi page flows

lawrence-forooghian:

Proposes a consistent manner for building multi-page journeys where there isn't a permanent backing data store for the data submitted on each page.

To allow us to build things like the check your answers page for the scheduling journey: https://dsdmoj.atlassian.net/browse/IC-1919, and to tidy up existing journeys like the referral assignment and cancellation.

## Next steps after merge

- test on dev
- &gt; 1 week after deploy, remove the legacy routes / actions (IC-2022)

### DIFF
--- a/doc/architecture/decisions/0011-use-session-based-drafts.md
+++ b/doc/architecture/decisions/0011-use-session-based-drafts.md
@@ -1,0 +1,76 @@
+# 11. Use session-based drafts
+
+Date: 2021-06-28
+
+## Status
+
+Accepted
+
+## Context
+
+### Background
+
+Our service frequently needs to take a user through a journey which involves asking them a series of questions over multiple pages, and then playing all the answers back to the user before finally submitting all the answers to the interventions service API.
+
+This means that we need somewhere to store the user’s answers as they progress through the journey. For some journeys, the interventions service provides this storage. For example, in the journey of submitting a referral, the interventions service provides endpoints for creating and updating a draft referral.
+
+However, we have other journeys where the interventions service does not provide this storage. For example, assigning a referral to a caseworker, or cancelling a referral. For these journeys, the UI application needs to provide its own storage.
+
+### Why not have the interventions service store everything?
+
+We _could_ make the interventions service provide storage for the data of every page of every journey in the UI. However, this would be very tied to the user journeys of this UI and less of a general-purpose interventions API. We should reserve interventions service storage for long journeys which the user might want to complete in multiple sittings.
+
+### The solution so far
+
+We pass data from page A to page B to page C of a journey by making sure that the HTTP request for page C includes all of the data that was submitted on pages A and B. We do this either by:
+
+- accepting the data from previous pages as data encoded in the request URL’s query, and then passing it in a GET request to subsequent pages by placing the data from previous pages into the query of the URL that’s used for navigating to the next page
+- accepting the data from previous pages as `application/x-www-form-urlencoded` form data, and then passing it in a POST request to subsequent pages by placing `<input type="hidden">` fields on each page, replaying the data from the previous pages
+
+### The problem with this solution
+
+Using a GET request limits the amount of data a user can submit on a page, since many clients and servers do not support URLs over 2000 bytes long.
+
+Using a POST request means that we cannot redirect to different pages in the journey (for example, check your answers) based on the user’s input, since a redirect response cannot instruct the browser to make a POST request.
+
+Also, the approach of embedding the data in the HTML is laborious. We have to make sure that every possible route through the pages in the journey preserves the data. This becomes particularly easy to get wrong when we have non-linear sequences of pages — for example, a link on the check your answers page that allows the user to edit a previous answer.
+
+### Other requirements for a solution
+
+The solution must also:
+
+- not prevent the user from performing the same journey multiple times concurrently — for example, they should be able to assign two different interventions at the same time in different browser tabs
+- preserve the data that the user entered on previous pages when they use the browser’s Back button
+
+It would also be good if the solution could:
+
+- not introduce new dependencies to the service — for example a database
+- allow us to continue using the same kinds of coding patterns as we do when interacting with the interventions service API — creating and updating a resource
+- allow us to identify and clean up old data
+
+## Decision
+
+We will use the Express `session` object as our storage. It’s backed by a Redis instance and allows us to store essentially unlimited amounts of data. The data is private to the current browsing session and is not accessible by other users.
+
+In the session, we will store “draft” objects. These are containers of arbitrary data, along with:
+
+- a globally unique identifier
+- timestamps of creation and last update
+- an identifier explaining what type of data the draft represents
+
+We’ll:
+
+- provide a “CRUD” API for creating, fetching, updating and deleting these draft objects
+- include the draft’s ID in all URLs in a journey
+- allow these drafts to be created by a GET request, so that we can use `<a>` tags to link to journeys
+- prefer to use POSTs to pass data to a page instead of GET, so we don't have to worry about body size constraints
+
+## Consequences
+
+Draft data will be lost after 2 hours of inactivity, [which is when the session cookie expires](../../../server/app.ts#L107). We should inform the user when this has happened, so that they can start the journey again. If this time limit proves to be a problem then we may need to reconsider the expiry duration.
+
+The URLs for some pages in our service will become longer.
+
+If we forget to delete a draft when the journey is complete, then it will continue to take up space in Redis. The metadata that we’re including on the drafts will allow us to clean this up later if need be, though.
+
+The draft data will only be as secure as our Redis instance.
--- a/server/services/draftsService.test.ts
+++ b/server/services/draftsService.test.ts
@@ -1,0 +1,174 @@
+import { Request } from 'express'
+import { mocked } from 'ts-jest/utils'
+import uuid from 'uuid'
+import DraftsService from './draftsService'
+
+jest.mock('uuid')
+const mockedUuid = mocked(uuid)
+
+describe(DraftsService, () => {
+  function createRequestWithSession(session: unknown): Request {
+    return { session } as unknown as Request
+  }
+
+  const clock = { now: () => new Date('2021-04-01T10:25:00Z') }
+
+  describe('.createDraft', () => {
+    it('generates a random UUID and stores the data in the session under a key derived from that UUID', () => {
+      mockedUuid.v4.mockReturnValue('2dc7ef56-58dd-4339-9924-c33318738068')
+      const req = createRequestWithSession({})
+
+      const draftsService = new DraftsService(clock)
+
+      const draft = draftsService.createDraft('someExampleType', { a: 1, b: 2 }, req)
+
+      expect(draft).toEqual({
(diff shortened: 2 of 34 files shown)


doc/architecture/decisions/0011-use-session-based-drafts.md
```diff
@@ -0,0 +1,76 @@
+# 11. Use session-based drafts
+
+Date: 2021-06-28
+
+## Status
+
+Accepted
```
lawrence-forooghian, line 7: (This is me being hopeful)

server/services/draftsService.test.ts
```diff
@@ -0,0 +1,174 @@
...
+
+      expect(draft).toEqual({
+        id: '2dc7ef56-58dd-4339-9924-c33318738068',
+        type: 'someExampleType',
+        createdAt: new Date('2021-04-01T10:25:00Z'),
+        updatedAt: new Date('2021-04-01T10:25:00Z'),
+        data: { a: 1, b: 2 },
+      })
+
+      expect(req.session['draft-2dc7ef56-58dd-4339-9924-c33318738068']).toEqual({
```
tomsmyers, line 33: nit: i think it would be clearer to use a seperator that's not in the uuid here e.g. `draft:...` or `draft/...`

server/services/draftsService.test.ts
```diff
@@ -0,0 +1,174 @@
...
+            type: 'someExampleType',
+            createdAt: '2021-04-01T10:25:00Z',
+            updatedAt: '2021-04-01T10:25:00Z',
+            data: { a: 1, b: 2 },
+          },
+        })
+
+        const draftsService = new DraftsService(clock)
+
+        expect(() => draftsService.fetchDraft('def456', req)).toThrow()
```
tomsmyers, line 84: how often will calls want to do fetches to check existence? e.g. try { fetchDraft(...) } catch { // no draft, do something else }

if it's a pattern we expect, it's more efficient to return a value here e.g. `null` rather than throw an execption.

server/services/draftsService.test.ts
```diff
@@ -0,0 +1,174 @@
...
+            updatedAt: '2021-03-10T11:50:00Z',
+            data: { a: 1, b: 2 },
+          },
+        })
+
+        const draftsService = new DraftsService(clock)
+
+        draftsService.deleteDraft('2dc7ef56-58dd-4339-9924-c33318738068', req)
+
+        expect(req.session['draft-2dc7ef56-58dd-4339-9924-c33318738068']).toBeUndefined()
```
tomsmyers, line 152: i would probably expect this to behave like fetch and update, i.e. throw on missing draft

tomsmyers, line 152: nit: maybe better to check that `fetchDraft` throws here instead of testing the underlying implementation detail

server/services/draftsService.ts
```diff
@@ -0,0 +1,74 @@
+import { Request } from 'express'
+import uuid from 'uuid'
+
+export interface Draft<Data> {
```
tomsmyers, line 5: nit: i'd find this more readable with a conventional type param e.g. `T`

server/services/draftsService.ts
```diff
@@ -0,0 +1,74 @@
...
+  id: string
+  type: string
+  createdAt: Date
+  updatedAt: Date
+  data: Data
+}
+
+export interface DraftDTO {
```
tomsmyers, line 13: nit: do we need a DTO that is exactly the same as the underlying data type?

server/services/draftsService.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+export interface Clock {
+  now: () => Date
+}
+
+export default class DraftsService {
+  constructor(private readonly clock: Clock) {}
+
+  private sessionKey(id: string): string {
```
tomsmyers, line 28: does the UUID package provider a UUID type? could we use it here?

tomsmyers: and else where in the below method signatures

server/services/draftsService.ts
tomsmyers, line 13: does this need to be exported?

server/services/draftsService.ts
```diff
@@ -32,7 +33,11 @@ export default class DraftsService {
     const dto = req.session[this.sessionKey(id)]
 
     if (dto === undefined) {
-      throw new Error(`Draft with ID ${id} not found in session`)
+      throw createError(404, new Error(`Draft with ID ${id} not found in session`), {
```
tomsmyers, line 36: minor nit: you can just pass the string as the second arg

tomsmyers: do we wanna be alerted to users hitting this error? if so, we should return 500 here, just because 404s are ignored in our monitoring tools

server/services/draftsService.test.ts
lawrence-forooghian: Sure thing, I'll change it.

server/services/draftsService.test.ts
lawrence-forooghian: I think that the "something else" will always be to show some sort of an error page. But, as I've mentioned in one of the commits, the error handling should probably be per-journey anyway, to show useful information about how to restart the journey. So probs makes sense to get rid of the generic error and handle on a case-by-case basis.

server/services/draftsService.test.ts
lawrence-forooghian: It kinda feels like that's then also testing `fetchDraft`, I dunno…

server/services/draftsService.ts
lawrence-forooghian: Nope, good spot 👍

server/services/draftsService.ts
lawrence-forooghian: 👍

server/services/draftsService.ts
lawrence-forooghian: Doesn't look like it, just seems to be `string`.

server/services/draftsService.ts
lawrence-forooghian: Yeah, I think I'd like to see how often it happens for now, to see how much we need to invest in better error messaging. So let's make it a 500 to start

lawrence-forooghian:
@tomsmyers thanks for the review so far. I've addressed all your comments. Would you like me to split this up or are you happy to review the whole thing?

server/services/draftsService.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export default class DraftsService {
+  constructor(private readonly clock: Clock) {}
+
+  private sessionKey(id: string): string {
+    return `draft:${id}`
+  }
+
+  private fetchDraftDTO(id: string, req: Request): DraftDTO | null {
+    const dto = req.session[this.sessionKey(id)]
```
andrewburrowsmoj, line 32: I see that we're using the RedisStore solution for session store. So I think it would be worthwhile testing this in an environment with multiple containers just to be certain its working as expected.

Looking at the config, I believe that the session store time-out is currently 120 minutes. Is this an expectable time-out for saving drafts? I think the users may expect their drafts to last a bit longer. Perhaps some wording to inform the users that their drafts will be saved for x minutes might be useful?

andrewburrowsmoj: Oh Sorry, I see that you made a comment about this in your documentation. 

You mention that it is 16 hours. How did you derive this?

From the code in `app.ts` I see: 

```
      cookie: { secure: config.https, sameSite: 'lax', maxAge: config.session.expiryMinutes * 60 * 1000 },
```

And the expiration defined in `config.ts`:

```
    expiryMinutes: Number(get('WEB_SESSION_TIMEOUT_IN_MINUTES', '120')),
```

I can't see `WEB_SESSION_TIMEOUT_IN_MINUTES` being set within the helm scripts so assuming it falls back to 120 minutes.

andrewburrowsmoj: Oh Sorry... I see that you made a comment about this in your documentation. 

I was looking at the timeout for the cookie rather than timeout for the session in redis store. You mention that it is 16 hours just curious where this is defined?

lawrence-forooghian: You know what, now that you ask, I have no idea how I managed to arrive at 16 hours. The line that I've linked to in `app.ts` in the documentation clearly shows it to be 120 minutes, like you say. I'll update the documentation to reflect this.

As to whether it's an acceptable timeout, I'm not sure. It probably comes down to the length of the journeys that we're using this for.

An alternative would be to, instead of storing these drafts in the session, to store them centrally in Redis, keyed by the username of the logged-in user. That way they wouldn't expire with the session. We could set a long TTL on these, perhaps a few days, after which Redis would clean them up. Do you think that might be a better idea? I can make that change if so.

lawrence-forooghian: > So I think it would be worthwhile testing this in an environment with multiple containers just to be certain its working as expected before deploying to production? Dev environment has 2 containers.

Yep, good idea. I was wondering if you know if any way to check which container a request gets routed to? (So that I can check it really is working with 2 containers and that it's not just that all the requests I test it with are coincidentally being routed to the same container.)

andrewburrowsmoj: > I was wondering if you know of any way to check which container a request gets routed to? 

We can simple do `kubectl log -f` on both containers. 
I suppose we can also test it locally by deleting the intervention-ui container and then starting it back up between requests.

andrewburrowsmoj: > An alternative would be to, instead of storing these drafts in the session, store them directly in Redis, keyed by the username of the logged-in user. That way they wouldn't expire with the session. 

My understanding is that the express session being backed with the Redis store will automatically store them directly in Redis straight away (otherwise we'd have the problem with multiple containers and may need to play around with the config)....

andrewburrowsmoj: I have another question (maybe for David) - when we redeploy, does the Redis container bounce as well? In which case all the session store will be lost given it's in-memory store.

andrewburrowsmoj: > We could set a long TTL on these, perhaps a few days, after which Redis would clean them up. Do you think that might be a better idea? I can make that change if so.

hmmm, I don't think it would be an issues setting it for a few days...

However, in order to reduce problems... we could inform the user that their draft will only be saved for x amount of minutes/hours. I don't know how to present that to the user though...

andrewburrowsmoj: Sorry to be polluting this comments ... another possible downside is that their session will not persist between browsers or different locations. As I think that the session store is based on the cookies stored in the browser.

lawrence-forooghian:
A few notes from what @andrewburrowsmoj and I just talked about:

- if the session were being wiped on deploy that'd mean everyone was being logged out on deploy, which I don't think is the case
- perhaps 2 hours is enough…
- but it might be nice to have the flexibility to set this lifetime independent of that of the session (increasing session lifetime has implications we don't necessarily want, like keeping user logged in longer), and would also mean it worked in multiple browsers (not really an aim, but nice if we get it for free)
- so I'll investigate putting stuff directly into Redis
- I'll confirm that updates to Redis are done synchronously before the response is returned (to avoid multiple container issues)
- speak to David to find out how to get `kubectl` working

lawrence-forooghian:
Superseded by #651.

PR #536, opened 2021-07-01
do not let PPs view action plans that have not been submitted

tomsmyers:

do not let PPs view action plans that have not been submitted

Screenshot 2021-07-01 at 16 08 05

stop PPs viewing action plans that are being edited if they click the link in a email after it has been withdrawn for editing.

### DIFF
--- a/server/views/shared/actionPlan.njk
+++ b/server/views/shared/actionPlan.njk
@@ -6,7 +6,7 @@
 {% from "govuk/components/inset-text/macro.njk" import govukInsetText %}
 {% from "govuk/components/button/macro.njk" import govukButton %}%}
 {% from "govuk/components/checkboxes/macro.njk" import govukCheckboxes %}
-
+{% from "govuk/components/warning-text/macro.njk" import govukWarningText %}
 
 {% block pageTitle %}
     HMPPS Interventions - GOV.UK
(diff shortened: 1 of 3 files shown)


server/views/shared/actionPlan.njk
```diff
@@ -15,43 +15,60 @@
 {% block pageContent %}
     {% block content %}
     <div class="govuk-!-width-two-thirds">
-        {{ govukBackLink(backLinkArgs) }}
-        {% if errorSummaryArgs !== null %}
-            {{ govukErrorSummary(errorSummaryArgs) }}
-        {% endif %}
-        <h1 class="govuk-heading-l">View action plan</h1>
+        {% if presenter.probationPractitionerBlockedFromViewing %}
```
lawrence-forooghian, line 18: What do you think about perhaps a controller test, too? Just something to check that message actually appears on the page. I'm in two minds to be honest, and would like to hear your thoughts.

lawrence-forooghian (approved):

server/views/shared/actionPlan.njk
tomsmyers: yeh - i kinda figured i'd write an integration test to cover it in the future instead (once action plan editing is finished e2e)

PR #537, opened 2021-07-01
IC-1992: Allow users to download list of structured interventions

Gweaton:

Adds a section to Find where users can download an .xlsx or .pdf version of a list of Structured Interventions.

This PR configures `git-lfs` to allow GitHub to more efficiently host our pdf files. It will require developers to install the package to be able to commit locally, as it runs as a pre-commit hook. See commit message in #694aa72 for more information.
 
To give PPs a place to view all structured interventions that are currently unsupported by the service.

## Screenshot

image


package.json
```diff
@@ -77,6 +77,15 @@
     "delay": 2500,
     "ext": "js,json,html,njk"
   },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged",
```
lawrence-forooghian, line 82: In #532 we upgraded to Husky 7, which changes the way hooks are defined. They're now defined as scripts like `pre-commit` in the repo's `.husky` directory. So I reckon that you'll need to take these out of here and put them in` .husky` (perhaps you can move the contents of the `hooks` directory into a directory called e.g. `.husky/git-lfs`, too?)

server/routes/probationPractitionerReferrals/probationPractitionerReferralsController.ts
```diff
@@ -59,7 +60,11 @@ export default class ProbationPractitionerReferralsController {
     const { accessToken } = res.locals.user.token
 
     const existingDraftReferrals = await this.interventionsService.getDraftReferralsForUserToken(accessToken)
-    const presenter = new FindStartPresenter(existingDraftReferrals)
+
+    const downloadFileInfo = await FileUtils.fileInformation(
+      'assets/downloads/Structured interventions list v1_3_3.xlsx'
+    )
+    const presenter = new FindStartPresenter(existingDraftReferrals, downloadFileInfo)
```
lawrence-forooghian, line 67: It feels a bit off for me to have a controller passing a formatted string to a presenter. Perhaps it would be more appropriate for the controller to be responsible for fetching the file's size as a `number` of bytes, and then the presenter would do the formatting.

server/routes/probationPractitionerReferrals/probationPractitionerReferralsController.ts
```diff
@@ -59,7 +60,11 @@ export default class ProbationPractitionerReferralsController {
     const { accessToken } = res.locals.user.token
 
     const existingDraftReferrals = await this.interventionsService.getDraftReferralsForUserToken(accessToken)
-    const presenter = new FindStartPresenter(existingDraftReferrals)
+
+    const downloadFileInfo = await FileUtils.fileInformation(
```
lawrence-forooghian, line 64: Can we find a way to avoid the duplication of this filename between here and `FindStartPresenter`?

server/utils/fileUtils.test.ts
```diff
@@ -0,0 +1,9 @@
+import FileUtils from './fileUtils'
+
+describe(FileUtils, () => {
+  describe('.fileInformation', () => {
+    it('returns the file type and size in KB, rounded to one decimal place', async () => {
+      expect(await FileUtils.fileInformation('testutils/test_logo.png')).toEqual('PNG, 10.9KB')
```
lawrence-forooghian, line 6: Perhaps this could go in a separate commit?

.gitattributes
```diff
@@ -0,0 +1 @@
+*.pdf filter=lfs diff=lfs merge=lfs -text
```
sldblog, line 1: 😍

server/utils/fileUtils.test.ts
```diff
@@ -0,0 +1,9 @@
+import FileUtils from './fileUtils'
+
+describe(FileUtils, () => {
+  describe('.fileInformation', () => {
+    it('returns the file type and size in KB, rounded to one decimal place', async () => {
```
lawrence-forooghian, line 5: I reckon we can return this in bytes; converting to KB and rounding can probs be considered a formatting concern.

assets/sass/components/_structured-interventions-downloads.scss
```diff
@@ -0,0 +1,17 @@
+.download-section {
```
lawrence-forooghian, line 1: We probably need some namespacing here.

server/routes/probationPractitionerReferrals/probationPractitionerReferralsController.ts
```diff
@@ -59,7 +60,19 @@ export default class ProbationPractitionerReferralsController {
...
 
     const existingDraftReferrals = await this.interventionsService.getDraftReferralsForUserToken(accessToken)
-    const presenter = new FindStartPresenter(existingDraftReferrals)
+
+    const structuredInterventionsFileDownloadPaths = {
+      xlsx: 'assets/downloads/Structured interventions list v1_3_3.xlsx',
+      pdf: 'assets/downloads/Structured interventions list v1_3_3.pdf',
+    }
+
+    const downloadFileSize = await FileUtils.fileSizeInKB(structuredInterventionsFileDownloadPaths.xlsx)
```
lawrence-forooghian, line 69: I wonder what these relative paths will be resolved against here. Probably the process's current working directory. I think it would be good to be explicit here, I'll take a look at how these paths get configured for the assets.

lawrence-forooghian: Oh, looks like they resolve relative to `process.cwd()` there, too. OK, will leave it.

PR #541, opened 2021-07-02
Update PostCSS to 7.0.36

lawrence-forooghian:

This fixes a security check error resulting from CVE-2021-23368.

To fix the nightly builds.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17167,9 +17167,9 @@
       }
     },
     "node_modules/postcss": {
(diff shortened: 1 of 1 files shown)


tomsmyers (commented):
looks like this is missing a change in package.json?

lawrence-forooghian:
I don't think so - it's only a transient dependency that comes from Stylelint

tomsmyers (approved):

PR #552, opened 2021-07-05
Merge `feature/sp-reporting-data` into `main`

lawrence-forooghian:

Merges the service provider reporting spreadsheet feature into `main`.

**The contents of this PR have already been reviewed.**

To add this new functionality.


server/routes/serviceProviderReferrals/dashboardNavPresenter.ts
```diff
@@ -0,0 +1,19 @@
...
+      text: 'All cases',
+      href: '/service-provider/dashboard',
+      active: this.active === 'All cases',
+    },
+    // TODO: do we need to restrict this to only SP managers?
+    {
+      text: 'Reporting',
+      href: '/service-provider/performance-report',
+      active: this.active === 'Reporting',
+    },
```
sldblog, line 17: Can we put this behind a feature flag (env var) and merge the branch?

lawrence-forooghian: Last time we talked about this, the worry was that we'd then end up with red contract tests on UI's `main`. We don't really want to be in a situation where red contract tests are acceptable, because it'll become easy to miss genuine violations of the contract. I think if we want to allow feature flags for UI functionality that depends on new API functionality, then we need a way of synchronising the feature flags with the tags used to decide which Pact tests should be used to determine the contract test check result.

lawrence-forooghian:
I've replaced this PR with #572.

PR #559, opened 2021-07-06
Clear inputs before typing when editing text in integration test

lawrence-forooghian:

Clears the text of `<input>` elements before typing new text, when editing text in the integration tests.

To stop a weird-looking appending from happening when we run the tests.


andrewburrowsmoj (approved):
Thanks for this. Was annoying me as well.

PR #590, opened 2021-07-12
IC-2060: Provide links to initial assessment feedback on progress page

andrewburrowsmoj:

Provides links from intervention progress page to access the feedback for supplier assessment for both service provider and probation practitioner.

Enable service provider to add feedback and view feedback afterwards.
Enable probation practitioner to view feedback after service provider has provided it.

Note: I realised a bit late that we need a new status called `AWAITING FEEDBACK` which changes from `SCHEDULED` when the appointment.appointmentTime is after the current date/time.
I will be doing this change in another PR (do not that this is merging into a feature branch). The ticket is here: https://dsdmoj.atlassian.net/browse/IC-2080

For now the `To do` has `View appointment details` and `Add feedback` for `SCHEDULED` appointment status. This is identical to how action plan appointments handles it.

#### NOT SCHEDULED

image

#### SCHEDULED
image

#### COMPLETED/NOT ATTENDED
image

#### NOT SCHEDULED
image

#### SCHEDULED
image

#### COMPLETED/NOT ATTENDED

image


server/routes/shared/appointment/feedback/submittedFeedbackPresenter.test.ts
```diff
@@ -0,0 +1,94 @@
...
+            lines: ['01 Feb 2021'],
+          },
+          {
+            key: 'Time',
+            lines: ['13:00'],
+          },
+        ])
+      }
+      it('extracts the date and time from the appointmentTime and puts it in a SummaryList format alongside the caseworker name', () => {
+        assertSessionDetailsWithCaseWorker(
```
Gweaton, line 50: I'd personally find repetition, rather than the re-usable method, clearer to understand here and the test below in terms of tests serving as documentation (particularly with the `it` description) as I have to scroll up to see what the expected output is.

server/routes/serviceProviderReferrals/serviceProviderReferralsController.test.ts
```diff
@@ -2155,6 +2155,61 @@ describe('GET /service-provider/referrals/:id/supplier-assessment/post-assessmen
...
+      .expect(200)
+      .expect(res => {
+        expect(res.text).toContain('View feedback')
+        expect(res.text).toContain('Did Alex attend the initial assessment appointment?')
+        expect(res.text).toContain('Yes, they were on time')
+        expect(res.text).toContain('Describe Alex&#39;s behaviour in this session')
+        expect(res.text).toContain('Acceptable')
+      })
+  })
+  it('renders an error if there is no current appointment for the supplier assessment', async () => {
```
Gweaton, line 2194: This is probably the worst comment I've ever left on a PR, and I wouldn't if I hadn't seen it a few times in this PR and in the codebase, but it's quite nice to leave a blank line between tests (particularly between `it`s or between `it`s and test setup).

...sorry. 

Don't worry about changing it in this PR, but just for future reference. I can also learn to live with it if you prefer it this way!

server/routes/probationPractitionerReferrals/probationPractitionerReferralsController.ts
```diff
@@ -197,6 +197,33 @@ export default class ProbationPractitionerReferralsController {
...
+      this.interventionsService.getSupplierAssessment(accessToken, referralId),
+    ])
+
+    if (!referral.assignedTo) {
+      throw new Error('Referral has not yet been assigned to a caseworker')
+    }
+
+    const { currentAppointment } = new SupplierAssessmentDecorator(supplierAssessment)
+    if (currentAppointment === null) {
+      throw new Error('Attempting to view supplier assessment feedback without a current appointment')
```
Gweaton, line 216: How likely is this? If fairly likely, do we want to surface a more friendly `userMessage` here, or do you think this is enough details for users to go on? e.g. do we want to use the term "initial assessment" rather than "supplier assessment" here?

Gweaton (approved):
Lots of good stuff here, just a few nitpicky comments from me.

server/routes/serviceProviderReferrals/serviceProviderReferralsController.test.ts
tomsmyers: 😂

server/routes/probationPractitionerReferrals/probationPractitionerReferralsController.ts
andrewburrowsmoj: It should not ever happen because the link would not be visible to the user until it satisfies that criteria. 
However, you're right about changing it away from supplier assessment; I'll make that change.

server/routes/serviceProviderReferrals/serviceProviderReferralsController.test.ts
andrewburrowsmoj: Sure, I'll aim to do that in future!

PR #595, opened 2021-07-13
IC-2072 add back links to action plan form

tomsmyers:

adds back links to all pages of the action plan form (except for the first page) to return to the previous page

allow SPs to navigate back in the form without using browser back button or editing the URL

### DIFF
--- a/server/routes/service-provider/action-plan/add-activities/addActionPlanActivitiesView.ts
+++ b/server/routes/service-provider/action-plan/add-activities/addActionPlanActivitiesView.ts
@@ -1,8 +1,22 @@
 import ViewUtils from '../../../../utils/viewUtils'
 import AddActionPlanActivitiesPresenter from './addActionPlanActivitiesPresenter'
+import { BackLinkArgs } from '../../../../utils/govukFrontendTypes'
 
 export default class AddActionPlanActivitiesView {
   constructor(private readonly presenter: AddActionPlanActivitiesPresenter) {}
+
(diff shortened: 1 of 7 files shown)


server/routes/service-provider/action-plan/add-activities/addActionPlanActivitiesView.ts
```diff
@@ -1,16 +1,31 @@
 import ViewUtils from '../../../../utils/viewUtils'
 import AddActionPlanActivitiesPresenter from './addActionPlanActivitiesPresenter'
+import { BackLinkArgs } from '../../../../utils/govukFrontendTypes'
 
 export default class AddActionPlanActivitiesView {
   constructor(private readonly presenter: AddActionPlanActivitiesPresenter) {}
 
+  private get backLinkArgs(): BackLinkArgs | null {
+    if (this.presenter.activityNumber === 1) {
+      return null
```
andrewburrowsmoj, line 10: We can perhaps this returning back to the progress page rather than null?

andrewburrowsmoj (approved):

server/routes/service-provider/action-plan/add-activities/addActionPlanActivitiesView.ts
tomsmyers: yeh i did think about that. i think we don't want to encourage users to exit this journey once started though. we can ask the UX people tomorrow.

PR #597, opened 2021-07-13
IC-2034: Display Responsible Officer(s) on referral details pages.

Gweaton:

Makes a request to the Community API for a service user's Responsible Officer.

Due to uncertainty about what might we get back from the API, there could be:
- Fields missing in the response
- Multiple `OffenderManager`s with `isResponsibleOfficer` set to true

Because of the above, we're planning to add a new endpoint to the Community API for just getting a Responsible Officer in future, but this should (as gracefully as possible) handle any issues until then by displaying a list of Responsible Officers where needed.

For SPs to know who to contact.

## Screenshot

image

### DIFF
--- a/server/models/delius/deliusOffenderManager.ts
+++ b/server/models/delius/deliusOffenderManager.ts
@@ -1,0 +1,9 @@
+export interface DeliusOffenderManager {
+  isResponsibleOfficer: boolean
+  staff?: {
+    forenames?: string | null
+    surname?: string | null
+    email?: string | null
+    phoneNumber?: string | null
+  }
+}
--- a/server/routes/shared/showReferralPresenter.ts
+++ b/server/routes/shared/showReferralPresenter.ts
@@ -21,6 +21,7 @@ import RiskSummary from '../../models/assessRisksAndNeeds/riskSummary'
 import RiskPresenter from './riskPresenter'
 import { DeliusStaffDetails, DeliusTeam } from '../../models/delius/deliusStaffDetails'
 import CalendarDay from '../../utils/calendarDay'
+import { DeliusOffenderManager } from '../../models/delius/deliusOffenderManager'
 
 export default class ShowReferralPresenter {
   referralOverviewPagePresenter: ReferralOverviewPagePresenter
@@ -39,7 +40,8 @@ export default class ShowReferralPresenter {
     readonly canAssignReferral: boolean,
     private readonly deliusServiceUser: ExpandedDeliusServiceUser,
     private readonly riskSummary: RiskSummary | null,
-    private readonly staffDetails: DeliusStaffDetails | null
+    private readonly staffDetails: DeliusStaffDetails | null,
+    private readonly responsibleOfficers: DeliusOffenderManager[]
   ) {
     this.referralOverviewPagePresenter = new ReferralOverviewPagePresenter(
       ReferralOverviewPageSection.Details,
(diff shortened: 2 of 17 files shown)


server/models/delius/deliusOffenderManager.ts
```diff
@@ -0,0 +1,9 @@
+export interface DeliusOffenderManager {
+  isResponsibleOfficer: boolean
+  staff: {
```
tomsmyers: could `staff` itself be missing in the response?

server/services/communityApiService.ts
```diff
@@ -78,4 +79,28 @@ export default class CommunityApiService {
...
+
+      if (!responsibleOfficer) {
+        throw createError(500, `No offender manager found in Delius for CRN ${crn}`, {
+          userMessage: 'No responsible officer is assigned to this service user in nDelius.',
+        })
+      }
+
+      return responsibleOfficer
+    } catch (err) {
+      throw createError(err.status, err, { userMessage: 'Could retrieve offender managers from nDelius.' })
```
tomsmyers, line 103: is 'offender manager' language SPs are familiar with?

server/services/communityApiService.ts
```diff
@@ -78,4 +79,28 @@ export default class CommunityApiService {
...
+    const token = await this.hmppsAuthService.getApiClientToken()
+
+    logger.info({ crn }, 'getting offender managers for service user')
+    try {
+      const deliusOffenderManagers = (await this.restClient.get({
+        path: `/secure/offenders/crn/${crn}/allOffenderManagers`,
+        token,
+      })) as DeliusOffenderManager[]
+
+      const responsibleOfficer = deliusOffenderManagers.filter(offenderManager => offenderManager.isResponsibleOfficer)
```
tomsmyers, line 93: nit: this is plural right?

server/services/communityApiService.ts
```diff
@@ -78,4 +79,28 @@ export default class CommunityApiService {
...
+
+      if (!responsibleOfficer) {
```
tomsmyers, line 95: is it the concern of this method to throw an error here?

should we not just return the empty list and let the caller handle it?

server/routes/shared/showReferralPresenter.ts
```diff
@@ -75,6 +77,48 @@ export default class ShowReferralPresenter {
         ]
   }
 
+  get responsibleOfficersDetails(): SummaryListItem[][] {
+    return this.responsibleOfficers.map(responsibleOfficer => {
+      const { staff } = responsibleOfficer
+
+      if (!staff) {
+        // This is a temporary step - I want to not break the page if fields are missing, most likely with dev data.
```
tomsmyers, line 85: this looks great - do you think it's ok though that the page will not load if there are no ROs returned for the SU in delius?

server/routes/shared/showReferralPresenter.ts
```diff
@@ -75,6 +77,48 @@ export default class ShowReferralPresenter {
...
+            lines: ['Not found'],
+          },
+          {
+            key: 'Email address',
+            lines: ['Not found'],
+          },
+        ]
+      }
+
+      const name = `${staff.forenames || ''} ${staff.surname || ''}`.trim() || 'Not found'
```
tomsmyers, line 103: nit: can we condense this to one block using `?.` on the staff object?

tomsmyers (approved):
looks great - thank you! couple of small comments.

server/routes/shared/showReferralPresenter.ts
Gweaton: I did try that originally, it turns out `${undefined}` === 'undefined', so if someone was missing a first or last name it would be displayed as "Peter undefined" or the other way round 🤦. 

This implementation does feel a bit over the top though, so let me know if you can think of any improvements

server/services/communityApiService.ts
Gweaton: Yeah, I think that makes sense.

server/routes/shared/showReferralPresenter.ts
Gweaton: Yeah I've stopped raising an error now and will just render a "Not found" section here as we have too many places where the journey can break so far.

server/models/delius/deliusOffenderManager.ts
Gweaton: I don't think it will ever be `null`, but it could be missing / `undefined` - this was added in the wrong commit but have tidied that up.

server/services/communityApiService.ts
Gweaton: I've updated this to say 'Could not retrieve Responsible Officer from nDelius' instead.

server/routes/shared/showReferralPresenter.ts
tomsmyers: sounds good to me - thanks

server/routes/shared/showReferralPresenter.ts
tomsmyers: lol! doesn't `${undefined} || ''` return empty string though?

PR #603, opened 2021-07-14
Quick fix: Stop title casing intervention titles

Gweaton:

Stops converting Intervention titles to Title Case in dashboards.

Looks like I introduced this behaviour a while ago, but it doesn't look right on the dashboards e.g:
- ETE -> Ete
- East of England -> East Of England.
- Accommodation/Good Tenant -> Accommodation/good Tenant

I think we can rely on the case on the Intervention itself, and it's what we're currently doing in the "Find" search results anyway.

### Before
Screenshot 2021-07-14 at 14-41-53 HMPPS Interventions - GOV UK

### After
Screenshot 2021-07-14 at 14-42-47 HMPPS Interventions - GOV UK


andrewburrowsmoj (approved):
lgtm

PR #604, opened 2021-07-14
Change SP reporting data download so that it triggers asynchronous generation on the backend

lawrence-forooghian:

Changes the service provider reporting data download functionality so that instead of directly triggering a download of a CSV file, it requests that the interventions service begins the generation of a CSV file, and then displays a confirmation page to the user telling them that we'll email them once the file is ready to download.

This introduces a new contract that @ministryofjustice/interventions-service will need to review.

To avoid issues (performance / UX) that may result from querying for performance data inside a request / response cycle. See commit messages for more details.

## Screenshots

Service provider referrals dashboard -- User requests reporting data -- allows the user to request a report based on the requested date range
Service provider referrals dashboard -- User requests reporting data -- allows the user to request a report based on the requested date range (1)


Gweaton (approved):
Nice. Sorry to see all that complex CSV behaviour go 😿

PR #605, opened 2021-07-15
IC-2084: Replaced the /report-a-problem link on the footer

andrewburrowsmoj:

Replaced the /report-a-problem link on the footer of all pages to point directly to a google docs link rather than our report-a-problem page.

Improves the quality of reported problems by forcing the user to go through a questionnaire form.

image

### DIFF
--- a/integration_tests/integration/login.spec.js
+++ b/integration_tests/integration/login.spec.js
@@ -29,9 +29,11 @@ context('Login', () => {
     })
 
     it('the user can report a problem', () => {
-      cy.contains('Report a problem').click()
-      cy.location('pathname').should('equal', `/report-a-problem`)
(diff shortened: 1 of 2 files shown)


andrewburrowsmoj:
I've left a question to Leigh to ask if we should instead have this link on the reportAProblem page itself rather than replacing it completely.

tomsmyers:
question to pass on to BAs / others in the team - this means all the support material is not longer accessible - is that ok ??

tomsmyers (approved):
change is obviously all good - however see my comment about support material which is included on the report-a-problem page.

andrewburrowsmoj:
> change is obviously all good - however see my comment about support material which is included on the report-a-problem page.

Thanks for the review.
The story has been updated now

PR #606, opened 2021-07-15
limit access to SP facing pages based on ROLE_CRS_PROVIDER

tomsmyers:

limit access to SP facing pages based on ROLE_CRS_PROVIDER

stop PPs accessing SP pages

### DIFF
--- a/server/routes/index.ts
+++ b/server/routes/index.ts
@@ -4,13 +4,12 @@ import asyncMiddleware from '../middleware/asyncMiddleware'
 import CommunityApiService from '../services/communityApiService'
 import InterventionsService from '../services/interventionsService'
 import HmppsAuthService from '../services/hmppsAuthService'
-import ServiceProviderReferralsController from './serviceProviderReferrals/serviceProviderReferralsController'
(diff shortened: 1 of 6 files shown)


server/routes/index.ts
```diff
@@ -19,32 +18,20 @@ export interface Services {
...
-    services.interventionsService,
-    services.communityApiService,
-    services.hmppsAuthService,
-    services.assessRisksAndNeedsService
-  )
-  const findInterventionsController = new FindInterventionsController(services.interventionsService)
   const commonController = new CommonController()
 
-  get('/', (req, res, next) => {
+  // fixme: we can't put these behind their own router yet because they do not share a common prefix
```
tomsmyers, line 31: note we can't fix this without altering some existing URLs

andrewburrowsmoj (approved):
Sorry took a long time - wanted to confirm all the urls .

Looks good - although not sure if '/find-interventions' should be PP only? Regardless it's not an issue since only the SP urls are guarded in this PR.

Are we going to eventually guard the PP urls for PP only?

PR #607, opened 2021-07-15
IC-2006 limit access to PP pages based on ROLE_PROBATION

tomsmyers:

limits access to the existing pages with prefix '/probation-practitioner', requiring ROLE_PROBATION. 

ensures when the user does _not_ have access, the generic auth error page is displayed.

not all PP facing pages are included in this change. that's because some of them do not share a common prefix, which means we can't use the ready made router middle ware to apply the auth checks. we can easily change them to use the common prefix, but we need to be careful about it because it could break bookmarks or existing sessions.

make the service more secure by limiting what SPs and PPs can see in the UI

### DIFF
--- a/integration_tests/integration/login.spec.js
+++ b/integration_tests/integration/login.spec.js
@@ -79,5 +79,14 @@ context('Login', () => {
       cy.get('[data-qa=logout]').click()
       AuthLoginPage.verifyOnPage()
     })
+
+    it('the user cannot access probation practitioner pages', () => {
+      cy.request({
+        url: '/probation-practitioner/dashboard',
+        failOnStatusCode: false,
+      })
+        .its('status')
+        .should('equal', 403)
+    })
   })
 })
--- a/integration_tests/integration/login.spec.js
+++ b/integration_tests/integration/login.spec.js
@@ -40,12 +40,10 @@ context('Login', () => {
     })
 
     it('the user cannot access service provider pages', () => {
-      cy.request({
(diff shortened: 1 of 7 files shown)


integration_tests/integration/login.spec.js
```diff
@@ -40,12 +40,10 @@ context('Login', () => {
     })
 
     it('the user cannot access service provider pages', () => {
-      cy.request({
-        url: '/service-provider/dashboard',
-        failOnStatusCode: false,
-      })
-        .its('status')
-        .should('equal', 403)
+      cy.request({ url: '/service-provider/dashboard', failOnStatusCode: false }).its('status').should('equal', 403)
```
tomsmyers, line 43: i couldn't find a way to check status _and_ content at once

server/errorHandler.ts
```diff
@@ -11,7 +11,7 @@ export default function createErrorHandler(production: boolean) {
 
     if (createError.isHttpError(err)) {
       // authorization errors from interventions service cause a special error page to be displayed
-      if (err.status === 403 && err.response?.body?.accessErrors) {
+      if (err.status === 403 && (!err.external || err.response?.body?.accessErrors)) {
```
tomsmyers, line 14: reading this in english:

if the status code is 403 and either the error is internal, or the error has a response with 'accessErrors' in the body, show the auth error page.

andrewburrowsmoj (approved):
Nice changes 👍

PR #614, opened 2021-07-20
IC-2094: Completely remove retries to services

sldblog:

Complete remove retries to our and 3rd party services

Currently, we're facing some load-related problems. If there's a problem, a retry creates _more_ load, which we don't need right now.

There's benefit to retries, but only if we can determine that it's not making things worse... I think that requires guarantees that the service wouldn't accept connections it cannot handle.


mahalrmoj (approved):

andrewburrowsmoj (approved):

sldblog:
🎩 🙇‍♂️ Thank you

PR #619, opened 2021-07-21
Fix undefined exception for delius addresses

andrewburrowsmoj:
When all addresses have a to field populated with a date, then this causes a undefined exception to be thrown as it assumes that there is at least one active address at all times

### DIFF
--- a/server/decorators/expandedDeliusServiceUserDecorator.test.ts
+++ b/server/decorators/expandedDeliusServiceUserDecorator.test.ts
@@ -223,79 +223,17 @@ describe(ExpandedDeliusServiceUserDecorator, () => {
         })
       })
 
-      describe('null  values checks', () => {
-        describe('when there is no address returned from nDelius', () => {
-          it('returns null', () => {
-            const serviceUser = new ExpandedDeliusServiceUserDecorator(
-              expandedDeliusServiceUserFactory.build({
-                contactDetails: {
-                  addresses: [],
-                },
-              })
-            )
+      describe('when there is no address returned from nDelius', () => {
+        it('returns null', () => {
+          const serviceUser = new ExpandedDeliusServiceUserDecorator(
+            expandedDeliusServiceUserFactory.build({
+              contactDetails: {
+                addresses: [],
+              },
+            })
+          )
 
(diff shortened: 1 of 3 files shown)


server/decorators/expandedDeliusServiceUserDecorator.test.ts
```diff
@@ -191,81 +191,57 @@ describe(ExpandedDeliusServiceUserDecorator, () => {
...
-                  addresses: [],
-                },
-              })
-            )
 
-            expect(serviceUser.address).toBeNull()
-          })
-        })
-        describe('when the addresses contains a single undefined address', () => {
+        describe('when all address have a "to" field', () => {
```
sldblog, line 195: The (obfuscated) data in question triggering the current `TypeError: Cannot read property 'addressNumber' of undefined` exception is
```
{
  "contactDetails": {
    "addresses": [
      {
        "from": "2021-04-09",
        "to": "2021-04-30",
        "noFixedAbode": true,
        "postcode": "xxx",
        "status": {
          "code": "P",
          "description": "Previous"
        },
        "typeVerified": false,
        "createdDatetime": "2021-04-09T17:07:30",
        "lastUpdatedDatetime": "2021-06-30T10:53:30"
      },
      {
        "from": "2020-06-29",
        "to": "2021-04-09",
        "noFixedAbode": false,
        "addressNumber": "xxx",
        "streetName": "xxx",
        "district": "xxx",
        "postcode": "xxx",
        "status": {
          "code": "P",
          "description": "Previous"
        },
        "typeVerified": false,
        "createdDatetime": "2020-10-09T11:08:00",
        "lastUpdatedDatetime": "2021-04-09T17:04:25"
      }
    ]
  }
}
```
I'd find it useful to describe the real-life scenario as a test case, perhaps "there is no current address added for the service user"?

andrewburrowsmoj: Sure. I've added that into the test. 
The test doesn't have all the fields as you show above because we don't declare all the fields in typescript.

sldblog (approved):

PR #637, opened 2021-07-27
Refactor: Restructure appointment directories

Gweaton:

Restructures the appointment directories for session feedback (Initial assessment & action plan sessions). It doesn't touch anything to do with the current directory structure for scheduling an appointment.

The proposed structure is as follows:
```
- appointments/
  -- feedback/
    --- actionPlanSessions/
      ---- attendance/
      ---- checkYourAnswers/
      ---- confirmation/
    --- initialAssessment/
      ---- attendance/
      ---- checkYourAnswers/
      ---- confirmation/
    --- shared/
      ---- attendance/
      ---- viewFeedback/
      ---- checkYourAnswers/
      ---- confirmation/
```

There's some behaviour shared between PPs and SPs, and some between
initial assessment and action plan sessions, so it makes sense to group
them. It also makes things easier to visualise from the directory
structure.

The motivation behind this came about when wanting to add in custom "behaviour"
presenters for initial assessment and action plan sessions (they
currently use a shared presenter), as we want to add custom content, and
possibly back links, to the respective pages.

The idea behind the original implementation was to structure directories like the URLs in
`index.ts` but this has become unwieldy due to shared components.


andrewburrowsmoj (approved):
Nice. I really like this structure!

PR #638, opened 2021-07-28
Create generic appointment summary component

andrewburrowsmoj:

Create a dedicated class for providing Appointment Summary details.

Have a single place for defining the appointment summary.

This component will be used in the following classes:

AttendanceFeedbackPresenter
CheckFeedbackAnswersPresenter
ScheduleAppointmentView (when rescheduled - not yet implemented)
SubmittedFeedbackPresenter
SupplierAssessmentAppointmentPresenter

I wanted to do this refactoring as part of my work on rescheduling an initial assessment appointment. As part of that work I need to provide an appointment summary. I noticed that this is duplicated in multiple places and has some inconsistencies.

After higher priority work has finished I will work on changing the above classes to use this component.

The appointment summary looks like this:
image

### DIFF
--- a/server/routes/appointments/appointmentSummaryComponent.ts
+++ b/server/routes/appointments/appointmentSummaryComponent.ts
@@ -1,0 +1,72 @@
+import { SummaryListItem } from '../../utils/summaryList'
+import { ActionPlanAppointment } from '../../models/actionPlan'
+import Appointment from '../../models/appointment'
+import AppointmentDecorator from '../../decorators/appointmentDecorator'
+import PresenterUtils from '../../utils/presenterUtils'
+import AuthUserDetails from '../../models/hmppsAuth/authUserDetails'
+
+export default class AppointmentSummaryComponent {
+  constructor(
+    private readonly appointment: Appointment | ActionPlanAppointment,
+    private readonly assignedCaseworker: AuthUserDetails | null = null
+  ) {}
+
+  private readonly appointmentDecorator = new AppointmentDecorator(this.appointment)
+
+  get appointmentSummary(): SummaryListItem[] {
+    return [
+      this.assignedCaseworker !== null
+        ? { key: 'Caseworker', lines: [`${this.assignedCaseworker.firstName} ${this.assignedCaseworker.lastName}`] }
+        : null,
+      {
+        key: 'Date',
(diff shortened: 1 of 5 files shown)


server/routes/appointments/appointmentSummaryComponent.ts
```diff
@@ -0,0 +1,72 @@
+import { SummaryListItem } from '../../utils/summaryList'
+import { ActionPlanAppointment } from '../../models/actionPlan'
+import Appointment from '../../models/appointment'
+import AppointmentDecorator from '../../decorators/appointmentDecorator'
+import PresenterUtils from '../../utils/presenterUtils'
+import AuthUserDetails from '../../models/hmppsAuth/authUserDetails'
+
+export default class AppointmentSummaryComponent {
```
Gweaton, line 8: I think we can drop the `Component` from the name here as we don't tend to use `component` much in our codebase - I initially thought `AppointmentSummaryPresenter` might be better, but equally `AppointmentSummary` could work on its own? What do you think?

server/routes/appointments/appointmentSummaryComponent.ts
```diff
@@ -0,0 +1,72 @@
...
+  constructor(
+    private readonly appointment: Appointment | ActionPlanAppointment,
+    private readonly assignedCaseworker: AuthUserDetails | null = null
+  ) {}
+
+  private readonly appointmentDecorator = new AppointmentDecorator(this.appointment)
+
+  get appointmentSummary(): SummaryListItem[] {
```
Gweaton, line 16: maybe we can call this `.summaryList` if we rename the class as suggested above.

Gweaton (commented):
Looks good - just one thing re: naming.

server/routes/appointments/appointmentSummaryComponent.ts
andrewburrowsmoj: I've renamed to appointmentSummaryList

server/routes/appointments/appointmentSummaryComponent.ts
andrewburrowsmoj: (we do have AddressFromComponent -  I don't think it matters much, but having a consistent way to differentiate between something that is 1-2-1 linked to a page and just a general component for use would be nice)

I've renamed to AppointmentSummary

Gweaton (approved):
Thanks for the changes!

PR #639, opened 2021-07-28
IC-2108 Add specific behaviour feedback presenters

Gweaton:

Refactors `BehaviourFeedbackPresenter`'s shared behaviour into `BehaviourFeedbackInputsPresenter` and creates two new presenters, one for Initial Assessment and one for Action Plan Sessions to provide custom text for each page, but allowing the presenters to share common behaviour.

This is a slight departure from the previous pattern of using abstract classes to share the common behaviour, as I feel composition here is slightly more flexible than using inheritance, and allows us to extend our classes with a bit more custom functionality if the need arises in the future. See commit messge in 3dda1c9 for more information.

To allow us to display different content on the Initial Assessment screen to the usual action plan session page.

This will also make it easier to pass in back links in future.

### DIFF
--- a/server/routes/appointments/feedback/shared/behaviour/behaviourFeedbackView.ts
+++ b/server/routes/appointments/feedback/shared/behaviour/behaviourFeedbackView.ts
@@ -1,11 +1,18 @@
-import { TextareaArgs } from '../../../../../utils/govukFrontendTypes'
+import { ErrorSummaryArgs, TextareaArgs } from '../../../../../utils/govukFrontendTypes'
 import ViewUtils from '../../../../../utils/viewUtils'
-import BehaviourFeedbackPresenter from './behaviourFeedbackPresenter'
+import BehaviourFeedbackInputsPresenter from './behaviourFeedbackInputsPresenter'
+import { BehaviourFeedbackPresenter } from './behaviourFeedbackPresenter'
 
 export default class BehaviourFeedbackView {
-  constructor(private readonly presenter: BehaviourFeedbackPresenter) {}
+  inputsPresenter: BehaviourFeedbackInputsPresenter
 
-  private readonly errorSummaryArgs = ViewUtils.govukErrorSummaryArgs(this.presenter.errorSummary)
+  constructor(private readonly presenter: BehaviourFeedbackPresenter) {
+    this.inputsPresenter = this.presenter.inputsPresenter
+  }
+
+  private get errorSummaryArgs(): ErrorSummaryArgs | null {
+    return ViewUtils.govukErrorSummaryArgs(this.inputsPresenter.errorSummary)
+  }
 
   private get textAreaArgs(): TextareaArgs {
     return {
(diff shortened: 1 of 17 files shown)


server/routes/appointments/feedback/shared/behaviour/behaviourFeedbackView.ts
```diff
@@ -1,11 +1,18 @@
...
 export default class BehaviourFeedbackView {
-  constructor(private readonly presenter: BehaviourFeedbackPresenter) {}
+  inputsPresenter: BehaviourFeedbackInputsPresenter
 
-  private readonly errorSummaryArgs = ViewUtils.govukErrorSummaryArgs(this.presenter.errorSummary)
+  constructor(private readonly presenter: BehaviourFeedbackPresenter) {
+    this.inputsPresenter = this.presenter.inputsPresenter
+  }
+
+  private get errorSummaryArgs(): ErrorSummaryArgs | null {
```
lawrence-forooghian, line 13: I'm curious why this change to `errorSummaryArgs`? Perhaps I missed something.

lawrence-forooghian (approved):
Looks good to me. One thing I was curious about.

server/routes/appointments/feedback/shared/behaviour/behaviourFeedbackView.ts
Gweaton: Yeah, this is an interesting one that had me scratching my head for ages a couple of months ago. This is what I initially wanted to do:
```
export default class BehaviourFeedbackView {
  inputsPresenter: BehaviourFeedbackInputsPresenter // this is the change that means we have to make errorSummaryArgs a get function.

  constructor(private readonly presenter: BehaviourFeedbackPresenter) {
    this.inputsPresenter = this.presenter.inputsPresenter
  }

  private readonly errorSummaryArgs = ViewUtils.govukErrorSummaryArgs(this.inputsPresenter.errorSummary)
```

However, the compiler complains that `this.inputsPresenter.errorSummary` (last line) is used before its initialization due to the way we assign the `inputsPresenter` value inside the body of the constructor.

The reason for this, is that in classes when we create a new instance of the class, behind the scenes it's really creating new object with a prototype full of methods & fields, a bit like:
```
const BehaviourFeedbackView = {}
behaviourFeedbackView.errorSummaryArgs = inputsPresenter.blah // this will be undefined because we haven't called the constructor yet, so inputsPresenter hasn't been assigned.
behaviourFeedbackView.aProperty = "string value" // basically what a readonly variable is doing in TS.
behaviourFeedbackView.aFunction = function () {}  // e.g. a get()
// then call the constructor and populate those values.
```

When we make this a `get`, the contents of the function aren't executed until _after_ the constructor is run, so the `inputsPresenter` value will actually assigned at that point.

To be perfectly honest, I'm not sure why it doesn't behave the same as the shorthand `private readonly presenter` argument passed into the constructor, I feel like that probably does something else behind the scenes, will do some digging.

Let me know if that doesn't make sense!

PR #641, opened 2021-07-29
IC-2081: Add back links to Attendance and Behaviour, CYA and viewing submitted feedback pages

Gweaton:

Adds back links to the attendance & behaviour feedback pages, as well as the check your answers and viewing submitted feedback pages in the Initial Assessment and Action Plan Sessions journeys.

To provide a link back to the previous pages.

## Screenshots

image

image

image

image

### DIFF
--- a/server/routes/appointments/feedback/actionPlanSessions/checkYourAnswers/actionPlanPostSessionFeedbackCheckAnswersPresenter.ts
+++ b/server/routes/appointments/feedback/actionPlanSessions/checkYourAnswers/actionPlanPostSessionFeedbackCheckAnswersPresenter.ts
@@ -28,4 +28,9 @@ export default class ActionPlanPostSessionFeedbackCheckAnswersPresenter extends 
   }
 
   readonly submitHref = `/service-provider/action-plan/${this.actionPlanId}/appointment/${this.actionPlanAppointment.sessionNumber}/post-session-feedback/submit`
+
+  readonly backLinkHref =
+    this.actionPlanAppointment.sessionFeedback.attendance.attended === 'no'
+      ? `/service-provider/action-plan/${this.actionPlanId}/appointment/${this.actionPlanAppointment.sessionNumber}/post-session-feedback/attendance`
+      : `/service-provider/action-plan/${this.actionPlanId}/appointment/${this.actionPlanAppointment.sessionNumber}/post-session-feedback/behaviour`
 }
--- a/server/routes/appointments/feedback/actionPlanSessions/checkYourAnswers/actionPlanPostSessionFeedbackCheckAnswersPresenter.ts
+++ b/server/routes/appointments/feedback/actionPlanSessions/checkYourAnswers/actionPlanPostSessionFeedbackCheckAnswersPresenter.ts
@@ -22,7 +22,8 @@ export default class ActionPlanPostSessionFeedbackCheckAnswersPresenter extends 
     )
     this.behaviourPresenter = new ActionPlanSessionBehaviourFeedbackPresenter(
       this.actionPlanAppointment,
-      this.serviceUser
+      this.serviceUser,
+      this.actionPlanId
     )
   }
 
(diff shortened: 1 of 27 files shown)


server/routes/appointments/feedback/actionPlanSessions/checkYourAnswers/actionPlanPostSessionFeedbackCheckAnswersPresenter.ts
```diff
@@ -22,9 +22,12 @@ export default class ActionPlanPostSessionFeedbackCheckAnswersPresenter extends
...
       this.actionPlanAppointment,
-      this.serviceUser
+      this.serviceUser,
+      this.actionPlanId
     )
   }
 
   readonly submitHref = `/service-provider/action-plan/${this.actionPlanId}/appointment/${this.actionPlanAppointment.sessionNumber}/post-session-feedback/submit`
+
+  readonly backLinkHref = `/service-provider/action-plan/${this.actionPlanId}/appointment/${this.actionPlanAppointment.sessionNumber}/post-session-feedback/behaviour`
```
andrewburrowsmoj, line 32: I don't think this is quite correct. There should be a conditional if the attendance == NO then goto attendance page.

server/routes/appointments/feedback/initialAssessment/checkYourAnswers/initialAssessmentFeedbackCheckAnswersPresenter.ts
```diff
@@ -21,4 +21,6 @@ export default class InitialAssessmentFeedbackCheckAnswersPresenter extends Chec
   }
 
   readonly submitHref = `/service-provider/referrals/${this.referralId}/supplier-assessment/post-assessment-feedback/submit`
+
+  readonly backLinkHref = `/service-provider/referrals/${this.referralId}/supplier-assessment/post-assessment-feedback/behaviour`
```
andrewburrowsmoj, line 25: Same here with the actionPlan appointment, need a conditional on if attendance == no

server/routes/shared/appointment/feedback/submittedFeedbackPresenter.ts
```diff
@@ -20,21 +20,31 @@ export default class SubmittedFeedbackPresenter extends FeedbackAnswersPresenter
...
   ) {
     super(appointmentDetails)
+
     if (this.isActionPlanAppointment(appointmentDetails)) {
       this.attendancePresenter = new ActionPlanPostSessionAttendanceFeedbackPresenter(
         appointmentDetails,
         this.serviceUser
       )
-      this.behaviourPresenter = new ActionPlanSessionBehaviourFeedbackPresenter(appointmentDetails, this.serviceUser)
+      this.behaviourPresenter = new ActionPlanSessionBehaviourFeedbackPresenter(
```
andrewburrowsmoj, line 35: I still think it would be nicer to have a component that encapsulates the `behaviour feedback` question and answers separate from other concerns (such as backlink and form presentation). I feel that passing all these null parameters becomes unwieldly.

We have a class called `feedbackAnswersPresenter.ts` which contains `AttendanceFeedbackPresenter` and `BehaviourFeedbackPresenter`.

I believe that this class doesn't need to be concerned with form details because we only ever use it after submission to the back-end has occurred. So we can remove the dependency to AttendanceFeedbackPresenter and BehaviourFeedbackPresenter and instead just use the appointment directly.
We can possibly also define the questions here as well (effectively inverting the relationship). In this way it would be more useable on other pages.

What do you think?

I realise this would be quite an involved change to do, but if you're in agreement with this - perhaps one of us can do it after this PR has been merged.

server/routes/appointments/feedback/actionPlanSessions/checkYourAnswers/actionPlanPostSessionFeedbackCheckAnswersPresenter.ts
Gweaton: Oops, you're right - forgot about that.

Gweaton: Have pushed these changes now

andrewburrowsmoj (approved):
lgtm; what do you think of my comment regarding the component?

server/routes/shared/appointment/feedback/submittedFeedbackPresenter.ts
Gweaton: I think I see what you mean - it does feel like we're tying a lot of things together and passing in more than we need. I feel like you have a clearer image of how this could look in your head, would you be happy to take it on after this PR is merged?

andrewburrowsmoj: Sure. It might not work once I get down to the details!

PR #642, opened 2021-07-29
Ic 1525/reschedule initial appointment when did not attend

andrewburrowsmoj:

Provides ability to create a new appointment if the service user did not attend the last appointment. Currently the SP user can only view the feedback.

Allows SP user to reschedule a new appointment if the user did not attend.

## Note:

In order to reduce the length of this PR I will add the attendance notes showing in previous appointment details in a future PR.

### When viewing referral progress the following link is now available:
image

### When the SP user clicks on reschedule
image

### When the SP user saves the new appointment
image

### When the user goes back to referral progress page
image

### DIFF
--- a/server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
+++ b/server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
@@ -540,7 +540,7 @@ export default class ServiceProviderReferralsController {
 
     await this.scheduleAppointment(req, res, {
       getReferral: async () => referral,
-      getCurrentAppointment: async () => new SupplierAssessmentDecorator(supplierAssessment).currentAppointment,
+      getCurrentAppointment: async () => currentAppointment,
       scheduleAppointment: paramsForUpdate =>
         this.interventionsService
           .scheduleSupplierAssessmentAppointment(
--- a/server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
+++ b/server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
@@ -536,7 +536,8 @@ export default class ServiceProviderReferralsController {
       referralId
     )
 
-    const hasExistingAppointment = supplierAssessment.currentAppointmentId !== null
+    const { currentAppointment } = new SupplierAssessmentDecorator(supplierAssessment)
+    const hasExistingScheduledAppointment = currentAppointment !== null && !currentAppointment.sessionFeedback.submitted
 
     await this.scheduleAppointment(req, res, {
       getReferral: async () => referral,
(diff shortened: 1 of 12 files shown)


server/utils/utils.ts
```diff
@@ -20,7 +23,12 @@ function convertToTitleCase(sentence: string | null): string {
   return isBlank(sentence) ? '' : sentence!.split(' ').map(properCaseName).join(' ')
 }
 
+function isInitialAssessmentAppointment(appointmentDetails: Appointment | ActionPlanAppointment): boolean {
```
Gweaton, line 26: can we take this out of `sessionStatus.ts` now and import it from here instead?

server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
```diff
@@ -540,7 +540,7 @@ export default class ServiceProviderReferralsController {
 
     await this.scheduleAppointment(req, res, {
       getReferral: async () => referral,
-      getCurrentAppointment: async () => new SupplierAssessmentDecorator(supplierAssessment).currentAppointment,
+      getCurrentAppointment: async () => currentAppointment,
```
Gweaton, line 544: where is `currentAppointment` defined? am I missing something?

Gweaton (approved):
This looks good to me, a couple of questions

server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
andrewburrowsmoj: It's defined on line 539

server/utils/utils.ts
andrewburrowsmoj: Yes. I forgot it was there! I think I do this a few times.

Seems that we need a general AppointmentUtils class?

Gweaton: I think `AppointmentDecorator` would be a sensible place for it, I don't think we'd need to change the name

andrewburrowsmoj: Thanks I've done this

server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
Gweaton: Ah, I was looking at the PR commit by commit, so it - it looks like this change was added in 7a89ff384 but the assignment on like 539 is added in the next commit in 093761ef40, hence the confusion. Ideally we'd have those two changes in at the same time, but don't worry too much about it now!

andrewburrowsmoj: ah I'm sorry about that.

I think the reason why this happened was because I did quite a bit of rebasing and I must have added it into the wrong commit.

PR #646, opened 2021-08-02
Ic 2130/add casework details to missed last appointment details

andrewburrowsmoj:

Adds caseworker details to show previous missed appointment when applicable.

If the user has rescheduled a (non-attended) missed appointment they will see the previous appointment details (time, date and delivery type). 

This PR adds to this section to also include the caseworker who submitted the feedback for that appointment.

Allows SP user to see which caseworker worked on the previous appointment

image


Gweaton (approved):
Looks good - do we need to wait for the backend changes (due to change in contract) before merging this?

PR #506, opened 2021-06-28
Use ephemeral session-based draft objects for multi page journeys that aren't backed by API calls

lawrence-forooghian:
## What does this pull request do?

Proposes a consistent manner for building multi-page journeys where there isn't a permanent backing data store for the data submitted on each page, and refactors the referral assignment and cancellation journeys to use it.

## What is the intent behind these changes?

To allow us to build things like the check your answers page for the scheduling journey: https://dsdmoj.atlassian.net/browse/IC-1919, and to tidy up existing journeys like the referral assignment and cancellation.

## Next steps after merge

- test on dev
- &gt; 1 week after deploy, remove the legacy routes / actions (IC-2022)

PR #651, opened 2021-08-03
Use temporary Redis-based draft objects for multi page journeys that aren't backed by API calls

lawrence-forooghian:

Proposes a consistent manner for building multi-page journeys where there isn't a permanent backing data store for the data submitted on each page, and refactors the referral assignment and cancellation journeys to use it.

**Note: this is a second pass at this functionality, first attempted in #506.**

To allow us to build things like the check your answers page for the scheduling journey: https://dsdmoj.atlassian.net/browse/IC-1919, and to tidy up existing journeys like the referral assignment and cancellation.

## Next steps after merge

- &gt; 1 week after deploy, remove the legacy routes / actions (IC-2022)

### DIFF
--- a/doc/architecture/decisions/0011-use-redis-backed-drafts.md
+++ b/doc/architecture/decisions/0011-use-redis-backed-drafts.md
@@ -1,0 +1,86 @@
+# 11. Use Redis-backed drafts
+
+Date: 2021-06-28
+
+## Status
+
+Accepted
+
+## Context
+
+### Background
+
+Our service frequently needs to take a user through a journey which involves asking them a series of questions over multiple pages, and then playing all the answers back to the user before finally submitting all the answers to the interventions service API.
+
+This means that we need somewhere to store the user’s answers as they progress through the journey. For some journeys, the interventions service provides this storage. For example, in the journey of submitting a referral, the interventions service provides endpoints for creating and updating a draft referral.
+
+However, we have other journeys where the interventions service does not provide this storage. For example, assigning a referral to a caseworker, or cancelling a referral. For these journeys, the UI application needs to provide its own storage.
+
+### Why not have the interventions service store everything?
+
+We _could_ make the interventions service provide storage for the data of every page of every journey in the UI. However, this would be very tied to the user journeys of this UI and less of a general-purpose interventions API. We should reserve interventions service storage for long journeys which the user might want to complete in multiple sittings.
+
+### The solution so far
+
+We pass data from page A to page B to page C of a journey by making sure that the HTTP request for page C includes all of the data that was submitted on pages A and B. We do this either by:
+
+- accepting the data from previous pages as data encoded in the request URL’s query, and then passing it in a GET request to subsequent pages by placing the data from previous pages into the query of the URL that’s used for navigating to the next page
+- accepting the data from previous pages as `application/x-www-form-urlencoded` form data, and then passing it in a POST request to subsequent pages by placing `<input type="hidden">` fields on each page, replaying the data from the previous pages
+
+### The problem with this solution
+
+Using a GET request limits the amount of data a user can submit on a page, since many clients and servers do not support URLs over 2000 bytes long.
+
+Using a POST request means that we cannot redirect to different pages in the journey (for example, check your answers) based on the user’s input, since a redirect response cannot instruct the browser to make a POST request.
+
+Also, the approach of embedding the data in the HTML is laborious. We have to make sure that every possible route through the pages in the journey preserves the data. This becomes particularly easy to get wrong when we have non-linear sequences of pages — for example, a link on the check your answers page that allows the user to edit a previous answer.
+
+### Other requirements for a solution
+
+The solution must also:
+
+- make sure that a user is not able to access data entered by a different user
+- not prevent the user from performing the same journey multiple times concurrently — for example, they should be able to assign two different interventions at the same time in different browser tabs
+- preserve the data that the user entered on previous pages when they use the browser’s Back button
+- give us maximum flexibility in deciding how to meet [WCAG 2.1’s Success Criterion 2.2.1 Timing Adjustable](https://www.w3.org/TR/WCAG21/#timing-adjustable) — for example, by making sure that the data is kept for at least 20 hours before it expires
+
+It would also be good if the solution could:
+
+- not introduce new dependencies to the service — for example a database
+- allow us to continue using the same kinds of coding patterns as we do when interacting with the interventions service API — creating and updating a resource
+- allow us to identify and clean up old data
+
+## Decision
+
+We’ll use the UI application’s existing Redis server as our storage. It allows us to store essentially unlimited amounts of data. The data is private to the current browsing session and is not accessible by other users.
+
+In Redis, we’ll store “draft” objects. These are containers of arbitrary JSON data, along with:
+
+- a globally unique identifier
+- the unique identifier of the user who created the draft
+- timestamps of creation and last update
+- an identifier explaining what type of data the draft represents
+
+We’ll:
+
+- provide a “CRUD” API for creating, fetching, updating and deleting these draft objects
+- include the draft’s ID in all URLs in a journey
+- allow these drafts to be created by a GET request, so that we can use `<a>` tags to link to journeys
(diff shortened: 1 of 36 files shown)


server/services/draftsService.ts
```diff
@@ -0,0 +1,116 @@
...
+      createdBy: { id: userId },
+      updatedAt: now,
+    }
+
+    await this.writeDraft(draft)
+
+    return draft
+  }
+
+  async fetchDraft<Data>(id: string, { userId }: { userId: string }): Promise<Draft<Data> | null> {
```
lawrence-forooghian, line 79: Question: would we expect fetching a draft to also bump its expiry date? If we consider refreshing the page to constitute "activity" then perhaps yes. But on the other hand, perhaps having 24 hours from initiating the journey is plenty. (I didn't build the thing I'm asking about here yet just because it'd be a bit more fiddly and I wanted to check first.)

doc/architecture/decisions/0011-use-redis-backed-drafts.md
```diff
@@ -0,0 +1,86 @@
...
+- prefer to use POSTs to pass data to a page instead of GET, so we don't have to worry about body size constraints
+
+The aforementioned API will:
+
+- enforce access control — making sure that a user is only allowed to access drafts that they created
+- make sure that drafts are automatically removed after they are no longer accessed for a certain amount of time (for example, 1 day), using Redis’s expiry functionality
+
+### Alternatives
+
+We might consider using the (Redis-backed) Express `session` object. However, this object expires after 2 hours of inactivity, which is insufficient for our needs. We don’t want to increase this timeout since there are security implications.
```
Gweaton, line 78: Sorry, I might have missed this, but what is the timeout going to be in our approach? If longer than 2 hours, are there not security implications with our approach too?

server/services/draftsService.ts
Gweaton: Good question - I wonder if this is something we'd want to ask George G or Jonathan, as they might know more information about how users are completing these journeys.

I'd say 24 hours is probably fine if bumping the expiry date is going to be fiddly, unless we hear otherwise from the UR team.

server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
```diff
@@ -211,11 +222,10 @@ export default class ServiceProviderReferralsController {
       )
     }
 
-    let assignee: AuthUserDetails
     const token = await this.hmppsAuthService.getApiClientToken()
 
     try {
-      assignee = await this.hmppsAuthService.getSPUserByEmailAddress(token, email)
+      await this.hmppsAuthService.getSPUserByEmailAddress(token, email)
```
Gweaton, line 228: Possibly a silly question, but what's this line doing, now that we're not using the return value?

server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
```diff
@@ -200,9 +205,15 @@ export default class ServiceProviderReferralsController {
     ControllerUtils.renderWithLayout(res, view, serviceUser)
   }
 
-  async checkAssignment(req: Request, res: Response): Promise<void> {
-    const email = req.query.email as string
+  async backwardsCompatibilityStartAssignment(req: Request, res: Response): Promise<void> {
```
Gweaton, line 208: Good thinking

Gweaton (approved):
This is great, nice work, and well done for making it backwards compatible.

I just have a couple of questions

doc/architecture/decisions/0011-use-redis-backed-drafts.md
lawrence-forooghian: I think that the new timeout was implicit by the stuff about the WCAG expiry time requirement, and later on I say "make sure that drafts are automatically removed after they are no longer accessed for a certain amount of time (for example, 1 day)".

I'll add a bit explaining why a long-lived session is a security risk, but long-lived drafts aren't.

server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
lawrence-forooghian: It's to check whether the user exists and to show an error page if not (see the adjacent `catch`).

doc/architecture/decisions/0011-use-redis-backed-drafts.md
Gweaton: New write-up makes sense, thanks for updating.

Gweaton (approved):

server/routes/probationPractitionerReferrals/probationPractitionerReferralsController.ts
```diff
@@ -246,82 +249,191 @@ export default class ProbationPractitionerReferralsController {
...
 
-  async submitFormAndShowCancellationCheckAnswersPage(req: Request, res: Response): Promise<void> {
+  private async fetchDraftCancellationOrThrowSpecificError(req: Request, res: Response) {
+    const id = req.params.draftCancellationId
+    const draftCancellation = await this.draftsService.fetchDraft<DraftCancellationData>(id, {
+      userId: res.locals.user.userId,
+    })
+
+    if (draftCancellation === null) {
+      throw createError(500, `Draft cancellation with ID ${id} not found by drafts service`, {
```
andrewburrowsmoj, line 332: nit: I think this should throw a 404 (not found) rather than 500?

server/routes/probationPractitionerReferrals/probationPractitionerReferralsController.ts
```diff
@@ -246,82 +249,191 @@ export default class ProbationPractitionerReferralsController {
...
+        userMessage:
+          'Too much time has passed since you started cancelling this referral. Your answers have not been saved, and you will need to start again.',
```
andrewburrowsmoj, line 334: nit: I wonder if it would be nicer to indicate to the user what "too much time" means. Maybe something like:

"Unable to retrieve draft as it has been over 24 hours since it has been accessed."

(maybe asking UX design on best wording)

andrewburrowsmoj (approved):
Really neat solution; I just left a minor comment on error status code and wording.

Perhaps the wording thing can be done as a change after if you're in agreement with it. (as it might take time to get approval from UX guys and also this scenario is unlikely to occur for the user)

server/routes/probationPractitionerReferrals/probationPractitionerReferralsController.ts
lawrence-forooghian: Yeah, I think it's unlikely they'll see it, as you say. I'd prefer not to expose the concept of a "draft" to the user since it's kinda just an implementation detail that we don't mention elsewhere in the journey. But perhaps I should be explicit about the time period (problem is that it's configurable so might be a slight faff to get the text right, but will take a look).

lawrence-forooghian: (Also, if we see that this error is actually happening, then we can iterate and make this page friendlier. Or just bump the timeout.)

server/routes/probationPractitionerReferrals/probationPractitionerReferralsController.ts
lawrence-forooghian: I think semantically that's probably right, but if I remember correctly I wanted to make sure this error shows up in monitoring to see if it's happening in real life, and I think that our alerting doesn't trigger for 404s. Does that sound sensible?

server/routes/probationPractitionerReferrals/probationPractitionerReferralsController.ts
lawrence-forooghian: Yeah, looking at it I think it's going to be a faff to do this with the configurable timeout (we have to do it in two places too). Reckon we can leave it until we find out whether it actually happens in real life?

server/routes/probationPractitionerReferrals/probationPractitionerReferralsController.ts
andrewburrowsmoj: yeah that makes sense, perhaps we should do a change to allow for any error with a userMessage to display? 
Although this is another PR...

PR #652, opened 2021-08-03
IC-2134: Add Session Type to Appointments

Gweaton:

Adds radio button section for setting the Session Type (1:1 or Group session) for action plan appointments.

Initial assessments are always 1:1, so these are hardcoded to 1:1 - I felt this was more re-usable than creating a specific type of session for Action Plan Appointments, even if we're not going to display this value anywhere at the moment.

I realised we don't have a page for showing the scheduled session from either a PP or SP's point of view, so this value isn't played back on a page anywhere - we have a way of viewing the scheduled initial assessment, but I don't think we need to actually display the session type on that page.

To allow SPs to specify what type of session the Service User is attending.

# Screenshot

image


andrewburrowsmoj (approved):
lgtm

PR #654, opened 2021-08-04
IC-1116 update this branch with latest changes from main and change URLs and params to match latest API

tomsmyers:

update this branch with latest changes from main and change URLs and params to match latest API

i tried rebasing this and the conflicts were heinous. a merge from main was the only sane way to do it, and even that was horrible.

the only new code here, other than the merge, is in the top commit. so that's all that really needs reviewing.

get this branch ready for merging into main. there will be one more PR coming on this which puts it all behind a feature flag.


andrewburrowsmoj (approved):
To me it doesn't appear that it contains the latest from main... (unless you're doing that in the to-be-merged branch)

Just wondering what is the issue with a feature flag instead of feature branch in this specific case?

PR #661, opened 2021-08-05
Fix missing csv file in dev

andrewburrowsmoj:
There is an issue with the csv file used by ReferenceDataService.ts whereby the location of the file is different when deployed on a docker container than when running locally.

We believe that /dist/server/ represents the root directory when running on docker so have adjusted the reference data path to reflec this.


Gweaton (approved):
🤞

PR #664, opened 2021-08-06
NPS office location feature flag

andrewburrowsmoj:
Added a feature flag for nps office location selection; this is defaulted to false.

    It can be overridden by setting the value in helm_deploy.

    The reason why that has been disabled is because there are some issues with delius returning conflicts when rescheduling appointments with different locations but with the same time.


tomsmyers (approved):
LGTM - this also could use integration tests which validate the 'turned off' feature flag. 

can we open a ticket to track this so we don't forget? thanks !

andrewburrowsmoj:
Thanks. I created the following ticket: https://dsdmoj.atlassian.net/browse/IC-2164

PR #665, opened 2021-08-06
IC-2161: Add function & contract tests for fetching Approved Action Plan Summaries from the Interventions Service

Gweaton:

Adds `GetApprovedActionPlanSummaries` method to the Interventions Service client to fetch a new `ApprovedActionPlanSummary` object from the backend, as well as a contract test for this.

This will be used to display the list of previously approved action plans in a list, so to keep this small we're using a "summary"-type model here.

## Screenshot of where this will be used:

image

### DIFF
--- a/integration_tests/plugins/index.js
+++ b/integration_tests/plugins/index.js
@@ -127,6 +127,10 @@ module.exports = on => {
       return interventionsService.stubGetActionPlan(arg.id, arg.responseJson)
     },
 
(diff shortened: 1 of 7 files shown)


Gweaton:
Here's the relevant backend PR https://github.com/ministryofjustice/hmpps-interventions-service/pull/501

andrewburrowsmoj (approved):
lgtm

PR #666, opened 2021-08-06
Refactor: Favour composition over inheritance for feedback pages.

andrewburrowsmoj:
Created a questionnaire class for attendance and behaviour. This holds all the questions for the feedback page.

The reason for this class is because presenting the attendance/behaviour feedback is used on multiple pages:
1. postSessionAttendanceFeedback.njk: this is the page where the user supplies the answers. It is used by both initial assessment and action plan session.
2. postSessionFeedbackResponses.njk: this is a partial page where the user can view/review their answers. It is used by three pages: (review answers before submitting, view feedback and reschedule appointment - this last one is still to be done).

The current attendanceFeedbackPresenter and behaviourFeedbackPresenter classes are coupled with backlinks and page title. For presenting the feedback we just want the actual question and answers; so by making it a seperate component it should reduce coupling.

### DIFF
--- a/server/routes/appointments/feedback/initialAssessment/checkYourAnswers/initialAssessmentFeedbackCheckAnswersPresenter.ts
+++ b/server/routes/appointments/feedback/initialAssessment/checkYourAnswers/initialAssessmentFeedbackCheckAnswersPresenter.ts
@@ -1,14 +1,10 @@
 import DeliusServiceUser from '../../../../../models/delius/deliusServiceUser'
 import CheckFeedbackAnswersPresenter from '../../shared/checkYourAnswers/checkFeedbackAnswersPresenter'
-import AttendanceFeedbackPresenter from '../../shared/attendance/attendanceFeedbackPresenter'
-import InitialAssessmentAttendanceFeedbackPresenter from '../attendance/initialAssessmentAttendanceFeedbackPresenter'
 import { InitialAssessmentAppointment } from '../../../../../models/appointment'
-import InitialAssessmentBehaviourFeedbackPresenter from '../behaviour/initialAssessmentBehaviourFeedbackPresenter'
+import FeedbackAnswersPresenter from '../../shared/viewFeedback/feedbackAnswersPresenter'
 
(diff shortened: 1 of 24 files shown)


server/routes/appointments/feedback/initialAssessment/checkYourAnswers/initialAssessmentFeedbackCheckAnswersPresenter.ts
```diff
@@ -1,23 +1,24 @@
...
     private readonly serviceUser: DeliusServiceUser,
     private readonly referralId: string
   ) {
     super(appointment)
-    this.attendancePresenter = new InitialAssessmentAttendanceFeedbackPresenter(appointment, this.serviceUser)
-    this.behaviourPresenter = new InitialAssessmentBehaviourFeedbackPresenter(appointment, this.serviceUser)
+    this.feedbackAnswersPresenter = new FeedbackAnswersPresenter(
+      appointment,
+      new AttendanceFeedbackQuestionnaire(appointment, serviceUser),
+      new BehaviourFeedbackQuestionnaire(appointment, this.serviceUser)
```
Gweaton, line 20: nit: is there a reason for `this.serviceUser` vs `serviceUser` above?

Gweaton (approved):
This looks good - I like the move towards composition, as it feels easier to use smaller components rather than having to incoporate the whole parent object we initially inherited from that could have more functionality than we needed. Just one tiny thing.

server/routes/appointments/feedback/initialAssessment/checkYourAnswers/initialAssessmentFeedbackCheckAnswersPresenter.ts
andrewburrowsmoj: They're the same thing in this case, but I've changed it to be more consistent.

Gweaton: Yeah - it's weird, my editor sometimes marks it as an usused variable if I do one but not the other, I don't know which. Feels good to be consistent though!

PR #667, opened 2021-08-06
IC-1116 allow user to download SP performance report

tomsmyers:

adds a new controller which creates a signed URL for accessing the performance report file in s3 and renders a page with a link.

Screenshot 2021-08-06 at 15 10 45

allow the SP to download completed performance reports.


Gweaton (approved):
Nice

PR #676, opened 2021-08-10
npm security audit fix

andrewburrowsmoj:
npm audit causes a security failure in circle ci because we are using an insecure version of tar defined as a dependency within pact-node.
The latest version of pact/pact-node does not have a fix to update the tar version, so this needs to be done by manually updating the version in package-lock.json

passes when running `npx audit-ci --config audit-ci.json`

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,6 +89,7 @@
       "prettier --write"
     ]
   },
+  "//": "tar is not a direct dependency of our project; we’ve only added it here to remove a security vulnerability introduced by a transitive dependency. This should be removed; see IC-2177",
   "dependencies": {
     "@ministryofjustice/frontend": "0.2.5",
     "@sentry/node": "^6.10.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -125,6 +125,7 @@
     "passport-oauth2": "^1.6.0",
     "redis": "^3.1.1",
     "superagent": "^6.1.0",
+    "tar": "^4.4.15",
```
lawrence-forooghian, line 128: We don't want to keep `tar` in our project's dependencies long-term because (amongst other reasons) it'll eventually affect our ability to update to the latest version of the Pact libraries once they become available (`pact-js-core` version's `main` has already updated to `tar` 6.1.4). Can we create a tech debt ticket to track this?

I'd also say that it's worth adding a `"//": "tar is not a direct dependency of our project; we’ve only added it here to remove a security vulnerability introduced by a transitive dependency. This should be removed; see IC-xxxx"` comment too.

andrewburrowsmoj: I agree and great suggestion. I've created a JIRA ticket: https://dsdmoj.atlassian.net/browse/IC-2177
and I've added a comment.

lawrence-forooghian (approved):

PR #679, opened 2021-08-10
Added the ability to use wiremock for community-api calls.

andrewburrowsmoj:
Added the ability to use wiremock for community-api calls.
The default url for community-api is now wiremock service.

This was added to have control of errors being thrown in local dev run when a user does not have any active convictions when making a referral.

Wiremock has been configured to proxy any requests that are unresolved and that start with /secure/ to community-api service, the default proxy for all calls still goes to intervention-service but it is at a lower priority in wiremock resolutions.

Ideally I would have proxied based on the urlPattern of '/community-api.\*' and '/interventions-service/.\*'; however it's not possible to strip the url path when proxying (at least in the standalone version - it's possible to write a custom transformer if you programmatically set up wiremock service). To fix this we could extend wiremock in a new project, write the transformer and have it available as a docker container.

Alternatively I could have added a custom header only for dev run whereby we set up a pattern based on the header being present, but I didn't want to play around too much with the rest client/ service classes.

### DIFF
--- a/server/config.ts
+++ b/server/config.ts
@@ -80,7 +80,7 @@ export default {
   },
   apis: {
     communityApi: {
-      url: get('COMMUNITY_API_URL', 'http://localhost:8091', requiredInProduction),
+      url: get('COMMUNITY_API_URL', 'http://localhost:9092', requiredInProduction),
       timeout: {
         response: Number(get('COMMUNITY_API_TIMEOUT_RESPONSE', 10000)),
         deadline: Number(get('COMMUNITY_API_TIMEOUT_DEADLINE', 10000)),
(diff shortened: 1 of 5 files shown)


server/config.ts
```diff
@@ -80,7 +80,7 @@ export default {
   },
   apis: {
     communityApi: {
-      url: get('COMMUNITY_API_URL', 'http://localhost:8091', requiredInProduction),
+      url: get('COMMUNITY_API_URL', 'http://localhost:9092', requiredInProduction),
```
tomsmyers, line 83: is this intentional? this breaks local dev with docker-compose

if you want to alter this just for the tests you can alter test.env

tomsmyers (commented):
does this affect logging in as a PP in local envs?

andrewburrowsmoj:
No I am able to log in as PP and SP.

All community-api calls that do not have mocks defined in `mock.ts` will be proxied to the real community-api.

tomsmyers (approved):

PR #680, opened 2021-08-10
IC-2161: Display list of previously approved action plans to PP

Gweaton:

Adds table with dates of previously approved action plans to the Action Plan view for PPs. It does not display for SPs (by design).

We're missing version number and a link to view the action plan here, which will come in future PRs. This PR is currently behind a feature flag so it won't show up to real users.

To give PPs an overview of previous edits to the action plan.

## Screenshot

image


server/routes/shared/action-plan/actionPlanPresenter.ts
```diff
@@ -14,7 +16,8 @@ export default class ActionPlanPresenter {
     private readonly actionPlan: ActionPlan,
     private readonly serviceCategories: ServiceCategory[],
     readonly userType: 'service-provider' | 'probation-practitioner',
-    private readonly validationError: FormValidationError | null = null
+    private readonly validationError: FormValidationError | null = null,
+    private readonly approvedActionPlanSummaries: ApprovedActionPlanSummary[] | null = null
```
andrewburrowsmoj, line 20: Rather than null it would be easier to defaulting to empty list so you don't need to handle null value. (that's what you pass from probationPractitionerReferralsController.ts if feature is off)

server/routes/shared/action-plan/actionPlanView.ts
```diff
@@ -39,6 +39,17 @@ export default class ActionPlanView {
     }
   }
 
+  get approvedActionPlansTableArgs(): TableArgs | null {
+    if (!this.presenter.showActionPlanVersions) {
```
andrewburrowsmoj, line 43: I personally don't think it is necessary to have this additional hide logic since it's the `njk` responsibility on weather to display(/call) the table or not

andrewburrowsmoj (commented):
Two small comments

server/routes/shared/action-plan/actionPlanPresenter.ts
Gweaton: I went back and forth on this, started out with an empty list and changed it to `null` - I agree, I think it is easier and we have fewer `null` checks. I changed it because I thought it might be useful to be able to distinguish between having no  previously-approved action plan summaries or it being `null` because it's an SP viewing the page and shouldn't see this section. I think I'm going to change it back though as I don't think the distinction is that useful, particularly as we can use the `userType`.

andrewburrowsmoj: ah I see your point -  I was thinking the null was there because of the feature flag; so in the way you describe a null does make sense. If you want to keep it as a null then I'm happy with that too.

andrewburrowsmoj (approved):

server/routes/shared/action-plan/actionPlanPresenter.ts
Gweaton: Have changed this back to an empty array

server/routes/shared/action-plan/actionPlanView.ts
Gweaton: Have removed this

PR #686, opened 2021-08-12
Refactor: Use appointment Summary class in view feedback pages

andrewburrowsmoj:
We have the concept of an appointment summary list which presents the: **Date** and **Time** of the appointment. 

See below `Before_img` for an example.

This is encountered on the following pages:
1. View appointment feedback page.
2. Check your answers for feedback page.
3. Attendance feedback page.

In all of these pages the logic for constructing the summary list is duplicated for each `View`.

---
We also have the following pages which present : **Date**, **Time**, **Meeting method**, **address** (if non-nps address) and **caseworker** (if applicable):

1. View initial assessment appointment details page
2. Rescheduling page for initial assessment appointment (shows last appointment details if present). 

See below `After_img` for an example.

---
This PR is to unify all of those pages to use a single AppointmentSummary component so that there is a single component responsible for how appointment summary is displayed.
I have already created this component which is in use in the last two pages described above.

The effect of having this component will mean that:

1. The **Time** will be a duration rather than a single moment (start time).
i.e. `1:01` vs `1:01am to 1:02am`
I think it's preferable to show more information than less.

2. The **Date** uses the full month name rather than short hand
i.e. `Jan` vs `January`

2. The **meeting method** and **address** will be visible for the appointment summary on the first three pages described above. Again I also think it's preferable to show more information than less

Confirmed with Rob if this is acceptable. The guidance for dates are:

image
https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#dates

And for times are:

image
https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#times

Before_img:
image

After_img:
image


server/routes/appointments/appointmentSummary.ts
```diff
@@ -3,43 +3,68 @@ import AppointmentDecorator from '../../decorators/appointmentDecorator'
...
 import { InitialAssessmentAppointment, ActionPlanAppointment } from '../../models/appointment'
+import User from '../../models/hmppsAuth/user'
+import { AppointmentDeliveryType } from '../../models/appointmentDeliveryType'
+import Address from '../../models/address'
 
 export default class AppointmentSummary {
   constructor(
     private readonly appointment: InitialAssessmentAppointment | ActionPlanAppointment,
-    private readonly assignedCaseworker: AuthUserDetails | null = null
+    private readonly assignedCaseworker: AuthUserDetails | User | null = null
```
Gweaton, line 13: I'm trying to figure out if there's a way we could only pass one of `User` or `AuthUserDetails` in here as I feel like checking the type of something e.g. based on it having a `firstName` field feels like a bit of a code smell - it might take a bit of refactoring the rest of the codebase (or not be possible, in which case, apologies), but what do you think?

We could potentially, as a first step, convert an `AuthUserDetails` into a `User` (or vice versa) and pass it in?

andrewburrowsmoj: I don't think it should be converted into `User` as that can be confusing for the caller.
What do you think about passing in the name into the constructor rather than an `AuthUserDetails` or `User` ?

andrewburrowsmoj: Yes, you're right....

I've created a common interface:

```
interface Caseworker {
  username?: string
  firstName?: string
  lastName?: string
}
```

Gweaton (approved):
Thanks for the change, looks good.

PR #692, opened 2021-08-13
Refactor dependency inversion for appointment summary

andrewburrowsmoj:
Refactored code so that AppointmentSummary is constructed outside of the Presenter classes and the responsibility of the Controller classes.

The reason for this change:

1. AppointmentSummary acts as a Presenter component - Controller is currently responsible for constructing Presenters. Constructing inside Presenter classes violated Dependency Inversion principle.
2. AppointmentSummary will eventually need to be modified to accept addition parameters (nps office address). This will make the constructor parameters in each of the components longer and longer.


Gweaton (approved):
This makes sense to me, good refactor.

PR #697, opened 2021-08-17
IC-1919: Add a "check your answers" page to the initial assessment scheduling journey

lawrence-forooghian:

Adds a page to the service provider initial assessment scheduling journey, which replays the answers they have submitted so that they can check them.

To allow the service provider to check their answers before scheduling the appointment.

## Screenshot

Screenshot 2021-08-17 at 11-55-08 HMPPS Interventions - GOV UK


Gweaton (approved):
Thanks for taking me through this - it looks good to me.

PR #698, opened 2021-08-17
Added assess risks and needs proxy

andrewburrowsmoj:
Added the ability to stub request with wiremock for assess-risks-and-needs calls. This is only for when you are running the application locally with "npm run:dev".

This will fix errors that are thrown when accessing the referral details page due to no data being available on assess-risks-and-needs service.
    
All stub requests can now be made in "mocks.ts".

Do note that all requests starting with "/risks/" will still be forwarded (by wiremork) to the assess-risks-and-needs container if you wish to use the actual container response instead of mocking.

### DIFF
--- a/mocks.ts
+++ b/mocks.ts
@@ -11,10 +11,14 @@ import draftReferralFactory from './testutils/factories/draftReferral'
 import serviceCategoryFactory from './testutils/factories/serviceCategory'
 import CommunityApiMocks from './mockApis/communityApi'
 import deliusConvictionFactory from './testutils/factories/deliusConviction'
+import AssessRisksAndNeedsServiceMocks from './mockApis/assessRisksAndNeedsService'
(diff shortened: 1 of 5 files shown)


mocks.ts
```diff
@@ -11,10 +11,14 @@ import draftReferralFactory from './testutils/factories/draftReferral'
 import serviceCategoryFactory from './testutils/factories/serviceCategory'
 import CommunityApiMocks from './mockApis/communityApi'
 import deliusConvictionFactory from './testutils/factories/deliusConviction'
+import AssessRisksAndNeedsServiceMocks from './mockApis/assessRisksAndNeedsService'
+import riskSummary from './testutils/factories/riskSummary'
```
lawrence-forooghian, line 15: Can we suffix these two variable names with `Factory` as elsewhere?

lawrence-forooghian (approved):

PR #699, opened 2021-08-17
Display risk to self on make a referral risk summary page

andrewburrowsmoj:
This PR is still in draft.

It is currently on hold because I need confirmation of the style that should be used.


andrewburrowsmoj:
These changes are no longer relevant - the way we display Risk has changed.

PR #700, opened 2021-08-17
Remove session type from supplier assessment scheduling form

lawrence-forooghian:

Makes sure that we never display the session type input (1:1 / group) on a supplier assessment scheduling form. I noticed this was happening when changing an upcoming appointment or rescheduling after a missed appointment. A supplier assessment is always 1:1.

To prevent the user from being confused and possibly entering invalid data.

### DIFF
--- a/server/routes/serviceProviderReferrals/scheduleAppointmentPresenter.ts
+++ b/server/routes/serviceProviderReferrals/scheduleAppointmentPresenter.ts
@@ -10,6 +10,7 @@ import config from '../../config'
 
 export default class ScheduleAppointmentPresenter {
   constructor(
+    private readonly formType: 'supplierAssessment' | 'actionPlan',
     private readonly referral: SentReferral,
     private readonly currentAppointment: InitialAssessmentAppointment | ActionPlanAppointment | null,
     private readonly currentAppointmentSummary: AppointmentSummary | null,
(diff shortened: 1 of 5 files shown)


server/routes/serviceProviderReferrals/scheduleAppointmentPresenter.ts
```diff
@@ -68,11 +69,11 @@ export default class ScheduleAppointmentPresenter {
...
   }
 
-  get isInitialAssessmentAppointment(): boolean {
-    return this.currentAppointment === null
+  get allowSessionTypeSelection(): boolean {
+    return this.formType !== 'supplierAssessment'
   }
 
-  readonly initialAssessmentSessionType = 'ONE_TO_ONE'
+  readonly sessionTypeWhenSelectionNotAllowed = 'ONE_TO_ONE'
```
Gweaton, line 76: I think the idea I was trying to convey behind the previous variable name was that all initial assessment appointments are 'ONE_TO_ONE', which is the reason we don't show a session type - I feel like it was possibly more explicit than `whenSelectionNotAllowed`, what do you think?

Gweaton (approved):
Thanks for spotting this, just one comment re: variable name

server/routes/serviceProviderReferrals/scheduleAppointmentPresenter.ts
lawrence-forooghian: So, I was thinking about the interface that the presenter exposes to the view. I chose the naming so that the view doesn't need to know about business logic (i.e. that an initial assessment appointment is distinguished in some way), and all it needs to know is whether to show the session type controls and which value to use in that case. What do you think?

Gweaton: Yeah, I suppose that's a fair point - the view doesn't need to know that logic, I wonder if there's some other place we can document that logic though, is there anywhere we do that at the moment?

lawrence-forooghian: I don't think we do – I've added a comment above the variable here.

PR #702, opened 2021-08-18
Ic 1481/display correctly formatted date and time across app

andrewburrowsmoj:
This change is to make sure that we are using the correct GOV UK styling for dates and times.

Documentation is located here:
 Dates: https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#dates
 Times: https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#times

Essentially the changes ask for:

1. Use long names for months unless space is an issue
2. Day of month should not have a leading 0
3. Time should be 12 hour clock (am/pm)
4. Use midday and middnight for 12:00pm and 12:00am
5. Range of time should follow the format: 1:00pm to 2:00pm
6. DateTime should follow the format: 1:00pm on 1 April 2021

This changes also involves a bit of refactoring to consolidate all Date presentation logic into DateUtils. There was logic in PresenterUtils that also dealt with dates.

The intent is to just have a single class responsible that returns the correct format; providing the correct formats for:
- a date
- a time
- a date and time
- a duration between times

It is still the responsibility of the caller of the methods to handle null values and ensure that they're passing in a correct string input for the date.

I thought this was preferrable to either returning an empty value or 'Not found' as the caller expects a valid formatted date to be produced.

### DIFF
--- a/server/routes/serviceProviderReferrals/scheduleAppointmentPresenter.test.ts
+++ b/server/routes/serviceProviderReferrals/scheduleAppointmentPresenter.test.ts
@@ -76,7 +76,7 @@ describe(ScheduleAppointmentPresenter, () => {
           expect(presenter.appointmentSummary).toEqual([
             { key: 'Caseworker', lines: ['firstName lastName'] },
             { key: 'Date', lines: ['2 January 2021'] },
-            { key: 'Time', lines: ['midday to 1:00pm'] },
+            { key: 'Time', lines: ['Midday to 1:00pm'] },
             { key: 'Method', lines: ['Video call'] },
           ])
         })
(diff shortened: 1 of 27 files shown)


server/routes/serviceProviderReferrals/scheduleAppointmentPresenter.test.ts
```diff
@@ -76,7 +76,7 @@ describe(ScheduleAppointmentPresenter, () => {
           expect(presenter.appointmentSummary).toEqual([
             { key: 'Caseworker', lines: ['firstName lastName'] },
             { key: 'Date', lines: ['2 January 2021'] },
-            { key: 'Time', lines: ['12:00pm to 1:00pm'] },
+            { key: 'Time', lines: ['midday to 1:00pm'] },
```
lawrence-forooghian, line 79: This should probably be capitalised? Ditto elsewhere when it's at the start of a sentence.

lawrence-forooghian (approved):

server/routes/serviceProviderReferrals/scheduleAppointmentPresenter.test.ts
andrewburrowsmoj: Thank you. I've added a new commit for this.

PR #706, opened 2021-08-19
enable SP performance report in prod

andrewburrowsmoj:

Turns on SP performance reporting in prod env

Allows users to access performance reports page and request for csv reports


Gweaton (approved):
Should we remove the flag itself, or is there a reason we're keeping it in?

andrewburrowsmoj:
Let#s leave the flag for now incase there is any serious problems with this feature it will be easier to turn it off

PR #709, opened 2021-08-20
IC-2188: Fix missing drafts error

lawrence-forooghian:

Displays a helpful message when a user attempts to view a page that uses a draft object after it has been deleted.

To stop the user from seeing a confusing error page when pressing the browser's Back button after completing a journey (e.g. scheduling a supplier assessment appointment).

## Screenshot

Screenshot 2021-08-19 at 16-15-14 HMPPS Interventions - GOV UK

### DIFF
--- a/server/utils/controllerUtils.ts
+++ b/server/utils/controllerUtils.ts
@@ -4,6 +4,7 @@ import LayoutPresenter from '../routes/shared/layoutPresenter'
 import LayoutView, { PageContentView } from '../routes/shared/layoutView'
 import DeliusServiceUser from '../models/delius/deliusServiceUser'
 import DraftsService, { Draft } from '../services/draftsService'
+import DraftSoftDeletedView from '../routes/shared/draftSoftDeletedView'
 
 interface DraftFetchSuccessResult<T> {
   rendered: false
@@ -24,7 +25,7 @@ export default class ControllerUtils {
     res.render(...view.renderArgs)
   }
 
-  static async fetchDraft<T>(
+  static async fetchDraftOrRenderMessage<T>(
     req: Request,
     res: Response,
     draftsService: DraftsService,
@@ -43,6 +44,13 @@ export default class ControllerUtils {
       throw createError(500, `Draft ${typeName} with ID ${id} not found by drafts service`, {
         userMessage: notFoundUserMessage,
       })
+    }
+
+    if (draft.softDeleted) {
+      res.status(410 /* Gone */)
+      const view = new DraftSoftDeletedView()
+      this.renderWithLayout(res, view, null)
+      return { rendered: true }
     }
 
     return { rendered: false, draft }
--- a/server/utils/controllerUtils.ts
+++ b/server/utils/controllerUtils.ts
@@ -5,6 +5,17 @@ import LayoutView, { PageContentView } from '../routes/shared/layoutView'
 import DeliusServiceUser from '../models/delius/deliusServiceUser'
 import DraftsService, { Draft } from '../services/draftsService'
 
+interface DraftFetchSuccessResult<T> {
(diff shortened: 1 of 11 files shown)


server/utils/controllerUtils.ts
```diff
@@ -10,4 +24,35 @@ export default class ControllerUtils {
...
+      userId: res.locals.user.userId,
+    })
+
+    if (draft === null) {
+      throw createError(500, `Draft ${typeName} with ID ${id} not found by drafts service`, {
+        userMessage: notFoundUserMessage,
+      })
+    }
+
+    if (draft.softDeleted) {
```
andrewburrowsmoj, line 49: I think that rendering a View should not be the responsibility of this method and instead should be the caller's responsibility. 
This is because:
1. The caller shouldn't expect the response to be terminated based on fetching a draft.
2. The caller will have to explicitly put scaffolding after calling to ensure that softDeleted drafts don't break the reset of the request. If you're doing that may as well do the rendering as well.

What do you think about throwing a custom error which is caught by the errorHandler to render the specific page?

lawrence-forooghian: I think they're valid concerns, but I wonder if it could perhaps be solved by improved naming of the method, like `fetchDraftOrRenderMessage`? I think it's possibly fair game for a method in `ControllerUtils `to render something, if we think it'd be reasonable for a helper instance method in a controller to do the same.

It's true that the caller could instead render the message, given that it has to check whether a render has happened anyway. My worry is that every controller method that wants to fetch a draft would then also have to make sure they check whether the draft is soft deleted before allowing it to be used. The way that the method is written at the moment, either the `fetchDraft` method returns a usable draft, or it doesn't return a draft at all. But I'm happy to move the rendering to the caller if you think the current approach is particularly egregious?

I'm also not a big fan of using exceptions to modify behaviour and control flow when there are more explicit ways to do it. I think that things like middlewares can be useful but they can also obscure how a controller method behaves by making it the result of multiple behaviours that are distant from each other in the codebase. I think that if we went down the route of using the `errorHandler` to handle a growing number of error-_ish_ cases then that class would become larger and larger as we started to handle more special cases. Unless we built a mechanism where we could attach an arbitrary view to the error for the error handler to render. But I think I'm a bit ambivalent about going down that route.

andrewburrowsmoj: > My worry is that every controller method that wants to fetch a draft would then also have to make sure they check whether the draft is soft deleted before allowing it to be used.

I personally think that is preferable than having the method rendering a page without the caller knowing about it. It is not obvious when looking at the code. Perhaps renaming the method might be best solution... 

I see your point about the caller having to know what to do with a softDeleted draft; however it can still be considered the callers responsibility to deal with it how they feel best - perhaps they might want to render what has been soft deleted, or perhaps the time in which it was soft deleted, etc.

One suggestion is to change the logic to this:
```
   const draft = await this.fetchDraftBookingOrThrowSpecificError(req, res)
    if (draft.softDeleted) {
         controllerUtils.renderDraftSoftDeletedView(req, res)
    }
```

Functionally it's no different than what you have currently, but this makes it more obvious on the controller level how pages are rendered based on logic.

lawrence-forooghian: I think your suggestion is a good one too. And if it _does_ turn out that we'd like to have custom handling for soft deleted drafts (or custom logic for which drafts to disallow), then can switch to that pattern in the future. In the interests of getting this released and not being sure whether we need that kind of customisability yet, I think that I'd like to stick with the working solution that I've already done. I've renamed the methods to hopefully make it clearer what they do. Please could you take another look?

andrewburrowsmoj (approved):

PR #720, opened 2021-08-20
Enable SP performance report in prod.

andrewburrowsmoj:
Turns on SP performance reporting in production. This means that the tab will be shown when a user logs in that allows them to request for a report.

It was turned off because we had a bug in production whereby the request fails for certain users that don't have their Auth user details persisted in the database.

Allow users to generated reports for SP performance.


sldblog (approved):

sldblog:
merge condition: testing of https://github.com/ministryofjustice/hmpps-interventions-service/pull/529

PR #726, opened 2021-08-23
Ic 1927/add specific page title for each page

andrewburrowsmoj:

Consolidated all the pageTitle logic into a single place located in layout.njk. The default pageTitle will be "HMPPS Interventions - GOV.UK".

Any page that extends this layout partial may modify the pageTitle by setting the variable pageTitle; in doing so the pageTitle will be displayed as "HMPPS Interventions - $PAGE_TITLE - GOV.UK".

There is also the option of a page subtitle which appends onto the pageTitle as : "HMPPS Interventions - $PAGE_TITLE - $PAGE_SUB_TITLE - GOV.UK"

In addition most pages now have a specific page title and page sub title associated with them.

Provide users with additional information on the page that they're on.
Improve flow of pages in Google Analytics.

Example:
image

### DIFF
--- a/server/views/referrals/checkAnswers.njk
+++ b/server/views/referrals/checkAnswers.njk
@@ -3,6 +3,9 @@
 
 {% extends "../partials/layout.njk" %}
 
+{% set pageTitle = "Make a referral" %}
+{% set pageSubTitle = "Check your answers" %}
+
 {% block pageContent %}
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
--- a/server/views/referrals/checkAnswers.njk
+++ b/server/views/referrals/checkAnswers.njk
@@ -3,12 +3,6 @@
 
 {% extends "../partials/layout.njk" %}
 
-{% set pageTitle = "HMPPS Interventions" %}
-{% block pageTitle %}
-  {{ pageTitle }}
-  - GOV.UK
-{% endblock %}
-
 {% block pageContent %}
   <div class="govuk-grid-row">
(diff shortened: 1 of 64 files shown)


server/views/referrals/checkAnswers.njk
```diff
@@ -3,10 +3,9 @@
 
 {% extends "../partials/layout.njk" %}
 
-{% set pageTitle = "HMPPS Interventions" %}
+{% set pageTitle = "HMPPS Interventions - Make a referral - Check your answers" %}
 {% block pageTitle %}
-  {{ pageTitle }}
-  - GOV.UK
+  HMPPS Interventions - {{ pageTitle }} - GOV.UK
```
lawrence-forooghian, line 8: I think that the prefix and suffix will be duplicated here?

lawrence-forooghian (commented):
Perhaps it would be a good time to de-duplicate the "HMPPS Interventions - " and " - GOV.UK" pre/suffix and put them in the shared template? And then each page template just provides the bit that's unique to the page.

server/views/referrals/checkAnswers.njk
andrewburrowsmoj: good spot, fixed

andrewburrowsmoj:
Thanks for your suggestion Lawrence. I agree with you. I've put all the logic for displaying page title in layout.njk since all pages extend from there.

lawrence-forooghian (approved):
Looks good, thanks for the changes 👍

PR #727, opened 2021-08-23
IC-1910: Order the sessions on referral details

andrewburrowsmoj:

There is a bug whereby the order of the sessions is not always in the correct order. This change ensures that the order is guaranteed to always be in order of the session number.

Ensures that users do not get out of order sessions.

### DIFF
--- a/server/routes/serviceProviderReferrals/interventionProgressPresenter.test.ts
+++ b/server/routes/serviceProviderReferrals/interventionProgressPresenter.test.ts
@@ -87,6 +87,26 @@ describe(InterventionProgressPresenter, () => {
       expect(presenter.sessionTableRows).toEqual([])
     })
 
+    describe('when there are multiple sessions', () => {
+      it('they must be ordered by session number', () => {
+        const referral = sentReferralFactory.build()
+        const actionPlan = actionPlanFactory.submitted().build({ id: '77923562-755c-48d9-a74c-0c8565aac9a2' })
+        const intervention = interventionFactory.build()
+        const presenter = new InterventionProgressPresenter(
+          referral,
+          intervention,
+          actionPlan,
+          [
+            actionPlanAppointmentFactory.build({ sessionNumber: 2 }),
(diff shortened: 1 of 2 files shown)


server/routes/serviceProviderReferrals/interventionProgressPresenter.test.ts
```diff
@@ -87,6 +87,60 @@ describe(InterventionProgressPresenter, () => {
       expect(presenter.sessionTableRows).toEqual([])
     })
 
+    describe('when there are multiple sessions', () => {
+      it('they must be ordered correctly', () => {
```
sldblog, line 91: _nit_ when a regression occurs, I find it useful if the test defines what "correct" is, so the test failure also explains what the correct should be

server/routes/serviceProviderReferrals/interventionProgressPresenter.test.ts
```diff
@@ -87,6 +87,60 @@ describe(InterventionProgressPresenter, () => {
...
+          [
+            actionPlanAppointmentFactory.build({ sessionNumber: 2 }),
+            actionPlanAppointmentFactory.build({ sessionNumber: 3 }),
+            actionPlanAppointmentFactory.build({ sessionNumber: 1 }),
+          ],
+          supplierAssessmentFactory.build()
+        )
+        expect(presenter.sessionTableRows).toEqual([
+          {
+            sessionNumber: 1,
```
sldblog, line 108: _suggestion_ what do you think about mapping to the sessionNumber, so the assertion does not fail on other, unrelated content changes in the future?

sldblog (approved):
LGTM, suggestions on test names/focussed assertion

server/routes/serviceProviderReferrals/interventionProgressPresenter.test.ts
andrewburrowsmoj: excellent suggestion

PR #752, opened 2021-08-31
IC-2022: Remove backwards compatibility routes for draft journeys

lawrence-forooghian:

Removes some routes that are no longer needed. These are routes that we preserved for backwards compatibility with the pre-drafts cancellation / assignment / supplier assessment scheduling journeys, but which we don't need any more now that sufficient time has passed since deploying the new routes.

To remove now-dead code.


sldblog:
> To remove now-dead code.

🙌 🥳

sldblog (approved):
Double checked ingress logs in Kibana, there were indeed no requests for these routes since 19 August at least (where ingress logs are still accessible)

PR #753, opened 2021-08-31
Mute log output when running Jest tests

lawrence-forooghian:

Disables log output when running Jest tests.

The test output (e.g. when we’re testing error cases in controllers) has
a tendency to swamp all of the Jest output, making it hard to find the
part of the output that actually describes the failing test assertion.

To be able to more easily locate information relating to the tests that are being run.


tomsmyers (approved):
i've never had a problem with this - but happy to see how it looks.

PR #764, opened 2021-09-02
Remove incorrect statement re privacy from drafts ADR

lawrence-forooghian:

Removes an incorrect statement from an ADR.

To not have misleading ADRs.

### DIFF
--- a/doc/architecture/decisions/0011-use-redis-backed-drafts.md
+++ b/doc/architecture/decisions/0011-use-redis-backed-drafts.md
@@ -52,7 +52,7 @@ It would also be good if the solution could:
 
 ## Decision
 
-We’ll use the UI application’s existing Redis server as our storage. It allows us to store essentially unlimited amounts of data. The data is private to the current browsing session and is not accessible by other users.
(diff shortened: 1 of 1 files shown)


doc/architecture/decisions/0011-use-redis-backed-drafts.md
```diff
@@ -52,7 +52,7 @@ It would also be good if the solution could:
 
 ## Decision
 
-We’ll use the UI application’s existing Redis server as our storage. It allows us to store essentially unlimited amounts of data. The data is private to the current browsing session and is not accessible by other users.
+We’ll use the UI application’s existing Redis server as our storage. It allows us to store essentially unlimited amounts of data.
```
tomsmyers, line 55: what is the impact of this practically speaking? does it expose to exploits where other users draft data can be read?

lawrence-forooghian: No practical implication – the access control is handled by the application instead (see "make sure that a user is not able to access data entered by a different user" further down).

Gweaton (approved):

PR #765, opened 2021-09-02
IC-2151 preserve line breaks and improve formatting across the whole application 

tomsmyers:

1405b1c09b7789dde5303d747678af524710dc7d removes html tags from intervention summary descriptions in the find page (we do not support html markup in intervention descriptions)

before: 
Screenshot 2021-09-02 at 10 47 36
after: 
Screenshot 2021-09-02 at 10 48 05

5d1392d315f69d6e82e3a5cf9f6a624fecc26d4a is a config change which stops a warning being printed on every commit

```
⚠ Detected incorrect braces with only single value: `*.{json}`. Reformatted as: `*.json`
```

b14f0bb67324ca48c35dfaaaa0b8d7ec63c55598 makes the risk information section of the referral check your answers page consistent with the rest of the page

before:
Screenshot 2021-09-02 at 10 54 12

after:
Screenshot 2021-09-02 at 10 54 56

73ecf2dee795f85e637c0104f70b68917a7f4077 tweaks the way conditional fields are displayed in the referral check your answers to better show line breaks

before:
Screenshot 2021-09-02 at 11 58 50

after:
Screenshot 2021-09-02 at 11 59 45

8a3410f3a3c347e7577f7fd738d5be63fc45a083 renders line breaks in summary lists - this affects pages like check your answers, referral details, etc etc 

before:
Screenshot 2021-09-02 at 12 04 43

after:
Screenshot 2021-09-02 at 12 05 35

143fa99c9d7b8895d19e68558f43b84e78718d38 renders line breaks in the post session feedback pages. this affects check your answers as well as viewing the post session feedback later.

before:
Screenshot 2021-09-02 at 12 04 59

after:
Screenshot 2021-09-02 at 12 05 21

e1f7c634c1bcc539d79e7887afd1b89ee74e3626 preserves line breaks in action plan activities across the board

before:
Screenshot 2021-09-02 at 12 25 37

after:
Screenshot 2021-09-02 at 12 29 15

08d37bf38a66e9c0ee333ca125c9725c24985f34 preserves line breaks in EOSR answers 

before:
Screenshot 2021-09-02 at 12 42 07

after:
Screenshot 2021-09-02 at 12 44 05

improve the overall UX for both SPs and PPs


server/utils/viewUtils.ts
```diff
@@ -13,6 +13,14 @@ export default class ViewUtils {
     return escape(val).val
   }
 
+  static nl2br(val: string): string {
+    // handle things like nunjucks macros (which are 'SafeStrings') that do not need formatting
```
lawrence-forooghian, line 17: I'm not sure I understand this comment. Do you mean "Nunjucks macros" or "the return value of a Nunjucks macro"?

server/utils/viewUtils.ts
```diff
@@ -13,6 +13,14 @@ export default class ViewUtils {
     return escape(val).val
   }
 
+  static nl2br(val: string): string {
```
lawrence-forooghian, line 16: The type of `val` is a bit confusing here given that on the next line we're checking whether it's a `string`. Would `unknown` be more appropriate?

server/utils/viewUtils.ts
```diff
@@ -13,6 +13,14 @@ export default class ViewUtils {
...
+    // handle things like nunjucks macros (which are 'SafeStrings') that do not need formatting
+    if (typeof val !== 'string') return val
```
lawrence-forooghian, line 18: Is it possible to write a test that expresses what this check is trying to achieve?

server/utils/viewUtils.ts
```diff
@@ -13,6 +13,14 @@ export default class ViewUtils {
...
+
+    const nl2br = new nunjucks.Environment().getFilter('nl2br')
+    return nl2br(ViewUtils.escape(val))
```
lawrence-forooghian, line 21: Again, would be good to have a test that describes the behaviour (i.e. that it escapes).

lawrence-forooghian: Thinking about it some more, I don't think that this method should be doing any escaping. I think we want to be able to look at any of our view code and be able to tell, at a glance, that it's correctly escaping values. We can currently do that since all of the escaping is done directly in the view code. With this change, developers will have to start carrying around the additional piece of knowledge that if they see a `ViewUtils.nl2br` then that's also correctly escaping things too.

server/utils/viewUtils.ts
tomsmyers: will clarify this comment. it's mainly the govuk tag macro. the thing where we pass the macro from the template back to anonymous js functions means this method ends up getting called with non-string arguments. it surprised me too when i saw it.

server/utils/viewUtils.ts
tomsmyers: see above comments. we want this method to only be called with strings. but due to some weirdness it ends up getting called with other stuff too.

server/utils/viewUtils.ts
tomsmyers: no drama - will remove.

tomsmyers:
> Why are we stripping tags? Would be useful to have some context in the commit message. If it's that SPs are including HTML in the descriptions then shouldn't we be educating them not to do that, and correcting the problem at source?
> 
> Also the commit message doesn't mention what seems to be the main thing here which is that we're preserving line breaks.

server/utils/viewUtils.ts
tomsmyers: sure - done

andrewburrowsmoj (approved):
awesome changes!

PR #767, opened 2021-09-03
Framework for pagination.

andrewburrowsmoj:

Uses the MOJ pagination component; it is quite a simple component just to display page numbers and previous/next buttons. It doesn't involve itself in the actual content for each page.
It doesn't calculate which page numbers to show, this is instead done by the new class /server/utils/pagination.ts.

No pagination is shown if all elements exist on one page.

The next button is hidden for the 1st page.

image

The previous button is hidden for the last page.

image

The maximum number of pages to show is 5
image

If it is the first 3 pages then `...` is shown before the last element.
image

If it is the last 3 pages then `...` is shown after the first element.

image

If it is somewhere in the middle then `...` is shown after the first element and before the last element.

image

The Page interface mirrors the Pageable interface within Spring.

Provides a framework for pagination

### DIFF
--- a/server/services/interventionsService.ts
+++ b/server/services/interventionsService.ts
@@ -64,6 +64,16 @@ export interface CreateReportDateParams {
   toIncludingDate: CalendarDay
 }
 
+// Pagination parameters for resources that use spring pageable
+export interface PaginationParams {
+  // Page number to retrieve -- starts from 1
+  page?: number
+  // Number of elements in a page
+  size?: number
+  // Sort by property, defaults to ascending order. If descending is required then add ',DESC' at the end of the property you want sorted i.e. ['$PROPERTY_NAME,DESC']
+  sort?: string[]
+}
+
 export default class InterventionsService {
   constructor(private readonly config: ApiConfig) {}
 
--- a/server/utils/pagination/pagination.ts
+++ b/server/utils/pagination/pagination.ts
@@ -1,0 +1,94 @@
+import { Page } from '../../models/pagination'
+
+interface MojPaginationArgs {
+  classes?: string
+  items: (
+    | {
+        type: 'dots'
+      }
+    | {
+        type: 'pageNumber'
+        selected: boolean
+        text: number
+        href: string
+      }
+  )[]
+  previous?: {
+    text: string
+    href: string
+  }
+  next?: {
+    text: string
+    href: string
+  }
+  results?: {
+    from: number
+    to: number
+    count: number
+  }
+}
+
+export default class Pagination {
+  constructor(private readonly page: Page<unknown>) {}
+
+  private constructPageItems(pageNumberRange: number[], chosenPageNumber: number): MojPaginationArgs['items'] {
+    return pageNumberRange.map(pageNumber => {
+      return {
+        type: 'pageNumber',
+        selected: chosenPageNumber === pageNumber,
+        text: pageNumber,
+        href: `?page=${pageNumber}`,
+      }
+    })
+  }
+
+  get mojPaginationArgs(): MojPaginationArgs {
+    const { totalPages } = this.page
+    if (totalPages <= 1) {
+      return {
+        items: [],
+      }
+    }
+    const count = this.page.totalElements
+    // page.number is zero indexed
+    const zeroIndexPageNumber = this.page.number
+    const chosenPageNumber = zeroIndexPageNumber + 1
+    const pageSize = this.page.size
+    const from = zeroIndexPageNumber * pageSize + 1
+    let to = from + pageSize - 1
+    if (to > count) {
+      to = count
+    }
+    const items: MojPaginationArgs['items'] = []
+    if (totalPages <= 5) {
+      items.push(
+        ...this.constructPageItems(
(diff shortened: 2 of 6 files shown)


server/services/interventionsService.ts
```diff
@@ -64,6 +64,16 @@ export interface CreateReportDateParams {
   toIncludingDate: CalendarDay
 }
 
+// Pagination parameters for resources that use spring pageable
```
tomsmyers, line 67: should this interface also live in the shared 'pagination' module?

server/services/interventionsService.ts
```diff
@@ -64,6 +64,16 @@ export interface CreateReportDateParams {
...
+export interface PaginationParams {
+  // Page number to retrieve -- starts from 1
+  page?: number
+  // Number of elements in a page
+  size?: number
+  // Sort by property, defaults to ascending order. If descending is required then use ['$PROPERTY_NAME,DESC'] i.e. ['sentAt,DESC']
```
tomsmyers, line 73: small nit: the quotes aren't quite right in the example arrays `['sentAt', 'DESC']`

tomsmyers: or are they correct as they are? if so it might be worth clarifying why this param is a list

server/utils/pagination/pagination.ts
```diff
@@ -0,0 +1,95 @@
...
+  results?: {
+    from: number
+    to: number
+    count: number
+  }
+}
+
+export default class Pagination {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  constructor(private readonly page: Page<any>) {}
```
tomsmyers, line 33: can you get rid of the es-lint-disable by using `unknown`?

server/services/interventionsService.ts
andrewburrowsmoj: DESC and ASC have to be coupled together with the sort property i.e. : ['sortByProperty1, DESC', 'sortByProperty2,ASC']

The documentation doesn't really explain that but it was what I observed from my testing.

server/utils/pagination/pagination.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  private constructPageItems(pageNumberRange: number[], chosenPageNumber: number): MojPaginationArgs['items'] {
```
tomsmyers, line 35: what does this type mean? `MojPaginationArgs['items']`

server/utils/pagination/pagination.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  get mojPaginationArgs(): MojPaginationArgs {
+    const { totalPages } = this.page
+    if (totalPages <= 1) {
+      return {
+        items: [],
+      }
+    }
+    const count = this.page.totalElements
+    // page.number is zero indexed
```
tomsmyers, line 54: is this still the case?

server/utils/pagination/pagination.ts
andrewburrowsmoj: MojPaginationArgs is an interface from Moj design for pagination.

We don't have @types defined for MOJ design - I see we do the same for other MOJ components in ViewUtils.

server/utils/pagination/pagination.ts
andrewburrowsmoj: Unfortunately yes.

You can only make the request non-zero indexed but the response is always zero-indexed.

I assume the reason is to not break the contract of the response.

I think it's still useful to have the request being non-zero indexed because response handling is encapsulated in this single class and all users of pagination won't have to worry about decrementing before sending the request.

server/views/partials/pagination.njk
```diff
@@ -0,0 +1,2 @@
+{%- from "moj/components/pagination/macro.njk" import mojPagination -%}
```
tomsmyers, line 1: is this nessecary ? can't the templates that want to use this just import it?

tomsmyers (approved):
looks good! couple of small questions

server/services/interventionsService.ts
andrewburrowsmoj: These are parameters just for the pagination request - all other params exist in this class too.

i.e.

```
export interface UpdateDraftEndOfServiceReportParams {
  furtherInformation: string | null
  outcome: CreateEndOfServiceReportOutcome | null
}
```

server/utils/pagination/pagination.ts
andrewburrowsmoj: yes, Thank you! I've changed to unknown.

server/services/interventionsService.ts
andrewburrowsmoj: I've modified the comment to be more descriptive

server/views/partials/pagination.njk
andrewburrowsmoj: I liked the idea of a partial page that handled the pagination as it's compartmentalised into a single place; however I didn't expect it to be a single line...

I still like the idea of leaving it as is because if there is a change that we need to do to all pagination elements then it would be easy to do by modifying this partial page...

PR #769, opened 2021-09-06
Update typescript to 4.4.2

lawrence-forooghian:

This replaces the Dependabot pull request #743. It updates TypeScript to 4.4.2. This involves changes to the code, to add type information to caught values. See commit message for more information.

To keep our dependencies up to date.

### DIFF
--- a/server/authentication/passport.ts
+++ b/server/authentication/passport.ts
@@ -66,6 +66,12 @@ export default function passportSetup(app: Application, hmppsAuthService: HmppsA
 
           verified(null, loggedInUser)
         } catch (error) {
+          // It almost certainly is an Error, but we need to convince the compiler
+          // so we can pass it to the callback
+          if (!(error instanceof Error)) {
+            throw new Error(`${error} is not an Error`)
+          }
+
           logger.error({ err: error, username: res.user_name }, 'failed to decode token or get user details')
           verified(error)
         }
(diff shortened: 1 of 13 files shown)


lawrence-forooghian:
Putting this back to draft, realised it's a bit rough around the edges.

Gweaton (approved):
Nice approach, looks good (just need to sort the merge conflicts)

server/authentication/passport.ts
```diff
@@ -66,6 +66,12 @@ export default function passportSetup(app: Application, hmppsAuthService: HmppsA
 
           verified(null, loggedInUser)
         } catch (error) {
+          // It almost certainly is an Error, but we need to convince the compiler
```
tomsmyers, line 69: what actually happens in the framework if we don't call the verified callback with the error?

could it be better to just cast the error to an `Error` and pass it anyway?

tomsmyers (approved):
lgtm - just one question about the passport verified callback

server/authentication/passport.ts
lawrence-forooghian: Or we could create our own `Error` explaining what's gone wrong and the non-`Error` value we caught?

lawrence-forooghian: Hang on, wait, that's exactly what I've done. I reckon this approach is alright - it's unlikely to occur and we'll still describe the value we caught.

PR #770, opened 2021-09-06
View all case notes

andrewburrowsmoj:
Please review this PR first: https://github.com/ministryofjustice/hmpps-interventions-ui/pull/770

This PR branches from it.

Show all case notes for SP and PP user.
    
This shows a paginated view of all case notes; defaulted to 5 notes per page, and ordered by sentAt in descending order.
    
When displaying the case note content the full case note is visible, there will be a future change to restrict this to 250 characters. This requires having a page to display the full case note - since this has not been implemented yet, it's best to show the full case note until that has been done.

Allow both SP and PP users to view case notes for a referral.

## Example

image

## Note 

The page won't be navigable for the user (unless they type it in the URL). This is because it doesn't make sense to show case notes without the ability to add a case note.

Eventually the page will be navigable by the navigation tabs:

image

This will be added in the `add case note` PR.

## Latest

latest changes to date presentation now look like:

image

### DIFF
--- a/server/routes/caseNotes/caseNotesController.ts
+++ b/server/routes/caseNotes/caseNotesController.ts
@@ -3,11 +3,13 @@ import InterventionsService from '../../services/interventionsService'
 import CaseNotesPresenter from './caseNotesPresenter'
 import CaseNotesView from './caseNotesView'
 import ControllerUtils from '../../utils/controllerUtils'
+import CommunityApiService from '../../services/communityApiService'
 import HmppsAuthService from '../../services/hmppsAuthService'
 
 export default class CaseNotesController {
   constructor(
     private readonly interventionsService: InterventionsService,
+    private readonly communityApiService: CommunityApiService,
     private readonly hmppsAuthService: HmppsAuthService
   ) {}
 
@@ -17,7 +19,12 @@ export default class CaseNotesController {
       page: Number(req.query.page),
       size: 5,
     }
-    const caseNotesPage = await this.interventionsService.getCaseNotes(accessToken, req.params.id, paginationQuery)
+    const referralId = req.params.id
+    const [referral, caseNotesPage] = await Promise.all([
+      this.interventionsService.getSentReferral(accessToken, referralId),
+      this.interventionsService.getCaseNotes(accessToken, referralId, paginationQuery),
+    ])
+    const serviceUser = await this.communityApiService.getServiceUserByCRN(referral.referral.serviceUser.crn)
     const uniqueUsers = new Set(caseNotesPage.content.map(caseNote => caseNote.sentBy.username))
     const userDetails: Map<string, undefined | string> = new Map(
       (
@@ -35,7 +42,7 @@ export default class CaseNotesController {
         )
       ).map(user => [user.username, user.fullName])
     )
-    const presenter = new CaseNotesPresenter(caseNotesPage, userDetails)
+    const presenter = new CaseNotesPresenter(caseNotesPage, userDetails, serviceUser)
     const view = new CaseNotesView(presenter)
     ControllerUtils.renderWithLayout(res, view, null)
   }
--- a/server/routes/caseNotes/caseNotesController.ts
+++ b/server/routes/caseNotes/caseNotesController.ts
@@ -3,9 +3,13 @@ import InterventionsService from '../../services/interventionsService'
 import CaseNotesPresenter from './caseNotesPresenter'
 import CaseNotesView from './caseNotesView'
 import ControllerUtils from '../../utils/controllerUtils'
+import HmppsAuthService from '../../services/hmppsAuthService'
 
 export default class CaseNotesController {
-  constructor(private readonly interventionsService: InterventionsService) {}
+  constructor(
+    private readonly interventionsService: InterventionsService,
+    private readonly hmppsAuthService: HmppsAuthService
+  ) {}
 
   async showCaseNotes(req: Request, res: Response): Promise<void> {
     const { accessToken } = res.locals.user.token
@@ -14,7 +18,24 @@ export default class CaseNotesController {
       size: 5,
     }
     const caseNotesPage = await this.interventionsService.getCaseNotes(accessToken, req.params.id, paginationQuery)
-    const presenter = new CaseNotesPresenter(caseNotesPage)
+    const uniqueUsers = new Set(caseNotesPage.content.map(caseNote => caseNote.sentBy.username))
+    const userDetails: Map<string, undefined | string> = new Map(
+      (
+        await Promise.all(
+          [...uniqueUsers].map(username => {
+            return this.hmppsAuthService
+              .getUserDetailsByUsername(accessToken, username)
+              .then(userDetail => {
+                return { username, fullName: userDetail.name }
+              })
+              .catch(error => {
+                return { username, fullName: undefined }
+              })
+          })
+        )
+      ).map(user => [user.username, user.fullName])
+    )
+    const presenter = new CaseNotesPresenter(caseNotesPage, userDetails)
     const view = new CaseNotesView(presenter)
     ControllerUtils.renderWithLayout(res, view, null)
   }
--- a/server/routes/caseNotes/caseNotesController.ts
+++ b/server/routes/caseNotes/caseNotesController.ts
@@ -1,0 +1,21 @@
+import { Request, Response } from 'express'
+import InterventionsService from '../../services/interventionsService'
+import CaseNotesPresenter from './caseNotesPresenter'
+import CaseNotesView from './caseNotesView'
+import ControllerUtils from '../../utils/controllerUtils'
+
+export default class CaseNotesController {
+  constructor(private readonly interventionsService: InterventionsService) {}
+
+  async showCaseNotes(req: Request, res: Response): Promise<void> {
+    const { accessToken } = res.locals.user.token
+    const paginationQuery = {
+      page: Number(req.query.page),
+      size: 5,
+    }
+    const caseNotesPage = await this.interventionsService.getCaseNotes(accessToken, req.params.id, paginationQuery)
+    const presenter = new CaseNotesPresenter(caseNotesPage)
+    const view = new CaseNotesView(presenter)
+    ControllerUtils.renderWithLayout(res, view, null)
+  }
+}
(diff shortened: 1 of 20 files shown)


server/routes/caseNotes/caseNotesView.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  private get tableArgs(): TableArgs {
+    return {
+      head: [
+        { text: 'Details', classes: 'govuk-!-width-one-quarter' },
+        { text: 'Case notes', classes: 'govuk-!-width-three-quarters' },
+      ],
+      rows: this.presenter.tableRows.map(row => {
+        return [
+          { html: `${row.sentAtDateTime}<p>${row.sentBy}` },
```
tomsmyers, line 20: paragraph tags should contain `class=govuk-body`

also where are the closing tags?

andrewburrowsmoj: closing <p> tag is optional according to html: https://html.com/tags/p/

I should make this a <br> rather than <p>

server/routes/caseNotes/caseNotesPresenter.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  constructor(private caseNotes: Page<CaseNote>) {
+    this.pagination = new Pagination(caseNotes)
+  }
+
+  readonly tableHeadings: string[] = ['Details', 'Case notes']
+
+  readonly tableRows: CaseNotesTableRow[] = this.caseNotes.content.map(caseNote => {
+    return {
+      sentAtDateTime: DateUtils.formattedDateTime(caseNote.sentAt),
```
tomsmyers, line 44: this seems like a sensible method to use. however i really don't like the way the formatted date times look for case notes. I think they probably work really well for appointments, but i would much prefer to see the date first, then the time. it makes visually grepping the dates much easier i think. what does the original design have?

server/routes/caseNotes/caseNotesView.ts
andrewburrowsmoj: I've changed the \<p\>s to \<br\>s

server/routes/caseNotes/caseNotesPresenter.ts
andrewburrowsmoj: The original has it as `Wednesday 16 April 2021 3:55pm`

However I'm following the guidance from 

https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#dates

and

https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#times

server/routes/caseNotes/caseNotesView.ts
```diff
@@ -0,0 +1,32 @@
...
+  private get tableArgs(): TableArgs {
+    return {
+      head: [
+        { text: 'Details', classes: 'govuk-!-width-one-quarter' },
+        { text: 'Case notes', classes: 'govuk-!-width-three-quarters' },
+      ],
+      rows: this.presenter.tableRows.map(row => {
+        return [
+          { html: `${row.sentAtDateTime}<br>${row.sentBy}` },
+          { html: `<b>${ViewUtils.escape(row.subject)}</b><br><br>${ViewUtils.escape(row.body)}` },
```
tomsmyers, line 21: please can we use govuk styles here?

server/routes/caseNotes/caseNotesPresenter.ts
andrewburrowsmoj: Ok, good point - I'll reform the date

server/routes/caseNotes/caseNotesView.ts
tomsmyers: it could also be worth seeing how the first cell looks with two paragraphs instead of a line break.

server/routes/caseNotes/caseNotesPresenter.ts
andrewburrowsmoj: Sorry, I deleted my previous comment but you must have seen it. 

I agree regarding formatting of the date. I'll make that change.

server/views/caseNotes/caseNotes.njk
```diff
@@ -0,0 +1,9 @@
+{% from "govuk/components/table/macro.njk" import govukTable %}
+{% extends "../partials/layout.njk" %}
+{% set pageTitle = "Referral" %}
+{% set pageSubTitle = "Case notes" %}
+
+{% block pageContent %}
+    {{ govukTable(tableArgs) }}
+    {% include "../partials/pagination.njk" %}
```
tomsmyers, line 9: now i see this being used i agree it's very neat.

server/routes/caseNotes/caseNotesController.ts
```diff
@@ -14,7 +22,26 @@ export default class CaseNotesController {
       size: 5,
     }
     const caseNotesPage = await this.interventionsService.getCaseNotes(accessToken, req.params.id, paginationQuery)
-    const presenter = new CaseNotesPresenter(caseNotesPage)
+    const uniqueUsers = caseNotesPage.content
```
tomsmyers, line 30: a Set is a more natural collection for this - it removes the requirement to filter dups too

server/routes/caseNotes/caseNotesController.ts
```diff
@@ -14,7 +22,26 @@ export default class CaseNotesController {
...
+      .map(caseNote => caseNote.sentBy)
+      // filter out duplicate username
+      .filter((user, index, self) => index === self.findIndex(otherUser => otherUser.username === user.username))
+    const userDetails: { username: string; officer: undefined | DeliusUser | AuthUserDetails }[] = await Promise.all(
+      uniqueUsers.map(async user => {
+        let officer: undefined | DeliusUser | AuthUserDetails
+        try {
+          if (user.authSource === 'delius') {
```
tomsmyers, line 38: you can do this using only hmpps auth - which is much preferred over CAPI. 

for an example see HMPPSAuthService::getUserDetail in the interventions service repo.

server/routes/caseNotes/caseNotesController.ts
```diff
@@ -14,7 +22,26 @@ export default class CaseNotesController {
       size: 5,
     }
     const caseNotesPage = await this.interventionsService.getCaseNotes(accessToken, req.params.id, paginationQuery)
-    const presenter = new CaseNotesPresenter(caseNotesPage)
+    const uniqueUsers = caseNotesPage.content
+      .map(caseNote => caseNote.sentBy)
+      // filter out duplicate username
+      .filter((user, index, self) => index === self.findIndex(otherUser => otherUser.username === user.username))
+    const userDetails: { username: string; officer: undefined | DeliusUser | AuthUserDetails }[] = await Promise.all(
```
tomsmyers, line 34: why don't we just make this thing return a set of single strings which are either the senders name or the username otherwise?

then the consumer of this code doesn't have to know about undefined `officer`s

server/routes/caseNotes/caseNotesController.ts
tomsmyers: oh is it more like a map of username -> display name? 

in that case why not make it a map?

`const displayNames: Map<string, string> = await Promise.all( .... )`

server/routes/caseNotes/caseNotesController.ts
tomsmyers: oh is it more like a map of username -> display name? 

in that case why not make it a map?

`const displayNames: Map<string, string> = await Promise.all( .... )`

server/routes/caseNotes/caseNotesController.ts
andrewburrowsmoj: Changed to a set (and there is a test to make sure only one call is made)

server/routes/caseNotes/caseNotesController.ts
andrewburrowsmoj: I've added a new end point for hmpps auth '/api/user/$USERNAME'. This is a new commit.

server/routes/caseNotes/caseNotesPresenter.ts
andrewburrowsmoj: I've added new commit for getting day of week into DateUtils.

server/routes/caseNotes/caseNotesController.ts
andrewburrowsmoj: I thought {username: string, userDetails: unknown} was more appropriate because username is defined with a key.

A map is easier to work with so I've changed to that.

server/routes/caseNotes/caseNotesView.ts
```diff
@@ -0,0 +1,38 @@
...
+        { text: 'Details', classes: 'govuk-!-width-one-quarter' },
+        { text: 'Case notes', classes: 'govuk-!-width-three-quarters' },
+      ],
+      rows: this.presenter.tableRows.map(row => {
+        return [
+          {
+            html: `<p class="govuk-body">${row.sentAtDay}<br>${row.sentAtDate}<br>${row.sentAtTime}<br>${row.sentBy}</p>`,
+          },
+          {
+            html: `<p class="govuk-body"><b>${ViewUtils.escape(row.subject)}</b><br><br>${ViewUtils.escape(
```
tomsmyers, line 24: sorry i missed this before - please can we also preserve line breaks here?

`ViewUtils.nl2br(ViewUtils.escape(...))`

tomsmyers: and this is a small nit, but previously when i mentioned govukstyles i was thinking more about using those styles to format the bold text using one of the heading classes. it's inconsequential, but would be preferred if it doesn't totally screw the formatting.

andrewburrowsmoj: Thanks, I missed this.

server/views/caseNotes/caseNotes.njk
```diff
@@ -4,6 +4,7 @@
 {% set pageSubTitle = "Case notes" %}
 
 {% block pageContent %}
+    <p>See all case notes about {{ serviceUserName }}</p>
```
tomsmyers, line 7: this is missing a govuk-body class

tomsmyers (approved):
awesome work - thanks! 

couple of lingering html styling notes

PR #772, opened 2021-09-06
IC-1751: Add version number to table of approved Action Plans

Gweaton:

Adds a version number to the table of previously-approved action plans.

To inform the PP about which version of an action plan they are viewing.

## Screenshot

image

### DIFF
--- a/integration_tests/integration/monitor.spec.js
+++ b/integration_tests/integration/monitor.spec.js
@@ -565,9 +565,11 @@ describe('Probation Practitioner monitor journey', () => {
         .getTable()
         .should('deep.equal', [
           {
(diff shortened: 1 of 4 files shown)


tomsmyers:
is the 'this version' thing in the designs george? it seems a bit strange to me because it's not an action!

andrewburrowsmoj (approved):
Looks good to me.

I can see the `This version` in the ticket.

Although, I personally would include this in the same PR that also links to `view action plan`.

Gweaton:
@tomsmyers @andrewburrowsmoj would you rather I took off the "this version" commit and added it in with the links to view the previous action plan instead?

PR #776, opened 2021-09-07
IC-2178: Add links to view previously approved action plans

Gweaton:

Adds the UI for viewing previously-approved action plans, with links to view them in the table at the bottom of the latest action plan page.

To allow PPs to view previous revisions of an action plan

### Seeing a list of previously approved action plans from the latest version

image

### Viewing an older version of the action plan

image

### DIFF
--- a/server/routes/shared/action-plan/actionPlanPresenter.ts
+++ b/server/routes/shared/action-plan/actionPlanPresenter.ts
@@ -34,6 +34,8 @@ export default class ActionPlanPresenter {
 
   readonly actionPlanEditConfirmationUrl = `/service-provider/referrals/${this.referral.id}/action-plan/edit`
 
+  readonly viewProbationPractitionerLatestActionPlanURL = `/probation-practitioner/referrals/${this.referral.id}/action-plan`
+
   readonly errorSummary = PresenterUtils.errorSummary(this.validationError)
 
   readonly fieldErrors = {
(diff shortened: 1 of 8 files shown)


server/routes/shared/action-plan/actionPlanPresenter.ts
```diff
@@ -68,19 +70,39 @@ export default class ActionPlanPresenter {
...
     return this.approvedActionPlanSummaries
       .sort((summaryA, summaryB) => new Date(summaryB.approvedAt).getTime() - new Date(summaryA.approvedAt).getTime())
       .map((summary, index) => ({
         versionNumber: this.approvedActionPlanSummaries.length - index,
         approvalDate: dateUtils.formattedDate(summary.approvedAt, { month: 'short' }),
+        href: this.determineActionPlanHref(summary.id),
       }))
   }
 
+  private determineActionPlanHref(summaryId: string): string | null {
```
andrewburrowsmoj, line 83: I personally would prefer that PP and SP logic is less confusing by having the switch logic directly on the element (so SP would return null or empty).
Though I realise that it's not directly part of your PR since the `showActionPlanVersions` already exists.

server/views/shared/actionPlan.njk
```diff
@@ -38,6 +39,10 @@
         {% endif %}
         <h1 class="govuk-heading-l">View action plan</h1>
 
+        {% if presenter.isViewingPreviousVersionOfActionPlan %}
+          {{ govukNotificationBanner(viewingPreviousActionPlanNotificationBannerArgs) }}
```
andrewburrowsmoj, line 43: If viewingPreviousActionPlanNotificationBannerArgs was null, would it simply not render like with govukTable?

andrewburrowsmoj (approved):
Overall some great changes. 

I think that the logic for what to display on View/Presenter/NJK can have reduced complexity to make it easier to understand.

server/views/shared/actionPlan.njk
Gweaton: As discussed earlier, we probably don't want to rely on this as it feels like it's relying on nunjucks ignoring a possible error (trying to render null)

server/routes/shared/action-plan/actionPlanPresenter.ts
Gweaton: I've spent a bit of time trying to figure out an alternative, but have ended up keeping this the same - I've moved up the function `showActionPlanVersions` to keep it closer to the main function so it's a bit clearer when this will be called, but I think a newcomer to the codebase will probably check the `.njk` template first to see what this logic is, then work backwards to the presenter, rather than the other way round.

I feel like this PP-only restriction might not be a permanent thing, and it might be visible to SPs soon, so I'm going to leave it for now, unless you have any strong objections. We can always refactor it in future.

PR #777, opened 2021-09-07
IC-1833: Surface API error when reducing number of sessions

Gweaton:

Resurfaces a 400 error from the Interventions Service when trying to reduce the number of sessions for a previously-approved action plan.

If there's already a pre-approved action plan with a set number of sessions, we can't remove any as it's unclear which ones we'd want to remove, and there's no behaviour for this written in the API.

### DIFF
--- a/server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
+++ b/server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
@@ -87,6 +87,7 @@ import AppointmentSummary from '../appointments/appointmentSummary'
 import ReferenceDataService from '../../services/referenceDataService'
 import InitialAssessmentCheckAnswersPresenter from './initialAssessmentCheckAnswersPresenter'
 import InitialAssessmentCheckAnswersView from './initialAssessmentCheckAnswersView'
+import createFormValidationErrorOrRethrow from '../../utils/interventionsFormError'
 
 export interface DraftAssignmentData {
   email: string | null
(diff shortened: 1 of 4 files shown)


server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
```diff
@@ -523,8 +523,25 @@ export default class ServiceProviderReferralsController {
       formError = form.error
 
       if (form.isValid) {
-        await this.interventionsService.updateDraftActionPlan(token, actionPlanId, form.paramsForUpdate)
-        return res.redirect(`/service-provider/action-plan/${actionPlanId}/review`)
+        try {
+          await this.interventionsService.updateDraftActionPlan(token, actionPlanId, form.paramsForUpdate)
+          return res.redirect(`/service-provider/action-plan/${actionPlanId}/review`)
+        } catch (e) {
+          if (e.status === 400) {
```
andrewburrowsmoj, line 530: How do we know that it will always be an issue with number of sessions?

Seems that we should have a check on the body returned containing `numberOfSessions` with error `CANNOT_BE_REDUCED`?

Gweaton: Yeah, I was thinking the same - I've realised we already have a mechanism for handling server errors like this, so have re-used that functionality.

andrewburrowsmoj (approved):

PR #782, opened 2021-09-08
40: Display notification about nDelius downtime

Gweaton:

Displays the broadcast message to users about nDelius being down from Friday 10th September at 5pm - Sunday 12th September at 12pm.

To inform users of the service that it will be unavailable during this time.

### DIFF
--- a/server/broadcast-message-config.json
+++ b/server/broadcast-message-config.json
@@ -1,5 +1,5 @@
 {
-  "start": "2021-08-19T00:00:00+01:00",
-  "end": "2021-08-22T23:00:00+01:00",
-  "message": "From 9:00am on Saturday 21 August until Sunday 22 August, it will not be possible to make referrals or record activity details in Refer and monitor an intervention. This is due to nDelius maintenance work."
+  "start": "2021-09-10T17:00:00+01:00",
+  "end": "2021-09-12T12:00:00+01:00",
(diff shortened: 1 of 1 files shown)


server/broadcast-message-config.json
```diff
@@ -1,5 +1,5 @@
 {
-  "start": "2021-08-19T00:00:00+01:00",
-  "end": "2021-08-22T23:00:00+01:00",
-  "message": "From 9:00am on Saturday 21 August until Sunday 22 August, it will not be possible to make referrals or record activity details in Refer and monitor an intervention. This is due to nDelius maintenance work."
+  "start": "2021-09-10T17:00:00+01:00",
+  "end": "2021-09-12T12:00:00+01:00",
+  "message": "From 18:30pm on Saturday 11 September until 12:00pm Sunday 12 September, it will not be possible to make referrals or record activity details in Refer and monitor an intervention. This is due to nDelius maintenance work."
```
lawrence-forooghian, line 4: I think this should either be "18:30" or "6:30pm"?

lawrence-forooghian: And perhaps "midday" instead of "12:00pm"?

Gweaton: It's a fair point, I was blindly following orders from the instructions on the ticket but your suggestions make more sense!

lawrence-forooghian (approved):

PR #786, opened 2021-09-08
Add new case note check your answers

andrewburrowsmoj:

Adds a check your answers page.
The case note is not created until `Confirm case note` button is clicked; this is because the draft service is being used.

If the user clicks on `< back` they will be taken back to the `add case note` page and with the subject and case note details filled in from the draft.

Provides a review screen for the user to double check their answers before submitting to add the new case note.

## Screenshot

image

### DIFF
--- a/server/routes/caseNotes/add/checkAnswers/addCaseNoteCheckAnswersPresenter.test.ts
+++ b/server/routes/caseNotes/add/checkAnswers/addCaseNoteCheckAnswersPresenter.test.ts
@@ -1,0 +1,57 @@
+import caseNoteFactory from '../../../../../testutils/factories/caseNote'
+import AddCaseNoteCheckAnswersPresenter from './addCaseNoteCheckAnswersPresenter'
+import userDetailsFactory from '../../../../../testutils/factories/userDetails'
+
+describe('AddCaseNoteCheckAnswersPresenter', () => {
+  const referralId = '8b588c52-91fd-48fa-89fe-6438748bceda'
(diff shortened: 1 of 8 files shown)


server/routes/caseNotes/checkAnswers/checkAddCaseNoteAnswersPresenter.ts
```diff
@@ -0,0 +1,33 @@
...
+
+interface CaseNotesSummary {
+  subject: string
+  date: string
+  time: string
+  from: string
+  body: string
+}
+
+export default class CheckAddCaseNoteAnswersPresenter {
```
Gweaton, line 13: Do you think this class can go in the `/caseNotes/add/` directory (as it's part of that journey) rather than a new `/check-answers` one?

Another nitpicky one, but I also wonder if `AddCaseNoteCheckAnswersPresenter` as a name might be more consistent with other classes in the codebase - I think we have examples of both, but we tend to lean towards having the `CheckAnswersPresenter` after the name of the user journey.

Gweaton (approved):
Looks good - I'm approving but just have a couple of pernickety comments, feel free to merge after making the changes if you agree with them

server/routes/caseNotes/checkAnswers/checkAddCaseNoteAnswersPresenter.ts
andrewburrowsmoj: Ah thanks for your suggestions. I agree.

I'll create a new directory as:

`/caseNotes/add/check-answers`

PR #788, opened 2021-09-08
Add new case note functionality

andrewburrowsmoj:

Provides ability to create a new case note for a referral.
When starting the new case note journey a new empty case note draft is created.

There is only validation that the subject or body should be provided.

Note that this page is not exposed to the user as there is no `Add case note` button on the view all case notes.

Allow users to add new case notes.

## Screenshots

image

image


server/routes/caseNotes/caseNotesController.test.ts
```diff
@@ -98,6 +98,7 @@ describe.each([
         communityApiService.getServiceUserByCRN.mockResolvedValue(deliusServiceUserFactory.build())
         await request(app)
           .get(`/${user.userType}/referrals/${sentReferral.id}/case-notes`)
+          .expect(200)
```
Gweaton, line 112: I think the commit message here (show full SU name on case notes) isn't right for the code change (adding an expectation for a 200 status code)

server/services/interventionsService.ts
```diff
@@ -599,4 +605,18 @@ export default class InterventionsService {
       query: { ...paginationParams },
     })) as Page<CaseNote>
   }
+
+  async addCaseNotes(token: string, createCaseNoteParam: CreateCaseNoteParams): Promise<CaseNote> {
+    const restClient = this.createRestClient(token)
+    const createCaseNoteDTO: Record<string, unknown> = {
```
Gweaton: Why do we need the type assertion here (`Record<string, unknown>`)? Also, is it not a `CreateCaseNoteParams`?

server/routes/caseNotes/caseNotesController.test.ts
```diff
@@ -104,26 +115,109 @@ describe.each([
...
     })
+
+    it('should combine the same users into a single call', async () => {
+      const caseNote1 = caseNoteFactory.build({
+        sentBy: { userId: '1', username: 'username', authSource: 'delius' },
+      })
+      const caseNote2 = caseNoteFactory.build({
+        sentBy: { userId: '1', username: 'username', authSource: 'delius' },
+      })
+      const caseNotePage: Page<CaseNote> = pageFactory.pageContent([caseNote1, caseNote2]).build() as Page<CaseNote>
```
Gweaton, line 126: It looks like we're double-casting here? Do we need the `: Page<CaseNote>` and `as Page<CaseNote>`?

server/routes/caseNotes/caseNotesController.ts
```diff
@@ -46,4 +56,64 @@ export default class CaseNotesController {
...
+      return
+    }
+    const draftCaseNote = fetchResult.draft
+    let error: FormValidationError | null = null
+    let userInputData: Record<string, string> | null = null
+    if (req.method === 'POST') {
+      const data = await new AddNewCaseNoteForm(req).data(referralId)
+      if (!data.error) {
+        try {
+          await this.draftsService.updateDraft(draftCaseNote.id, data.paramsForUpdate, {
```
Gweaton, line 90: I might be misunderstanding here, but do we need to update the draft here if we're finalising the case note on the next line?

Gweaton (commented):
Great work, a couple of small questions

server/services/interventionsService.ts
Gweaton: Ah I think this was removed in a subsequent commit, don't worry.

server/routes/caseNotes/caseNotesController.test.ts
andrewburrowsmoj: Think that was introduced by a rebase. I've dropped the commit.

server/routes/caseNotes/caseNotesController.test.ts
andrewburrowsmoj: I've changed it to just `as Page<CageNote>`

server/routes/caseNotes/caseNotesController.ts
andrewburrowsmoj: Yes, technically a draft referral is not necessary at all with a single page.
It comes into play when there is more than one page on the journey. 

I must have added it into the wrong commit, as it makes more sense in the next commit (on another PR) where there is a check your answers page.

I've removed it - I will make sure the following PR is rebased after this is merged into main.

Gweaton (approved):
Thanks for making the changes

PR #792, opened 2021-09-09
41: Fetch main address for Service Users if set in nDelius

Gweaton:

Fetches the main address specified in nDelius (`status: { code: "M" }`), falling back to the previous logic to fetch the address with the most recent `from` date` with a `to` date that hasn't elapsed if the status isn't present in the response.

To make it easier for users to correct the address in Delius without having to follow complex logic like to/from dates.


andrewburrowsmoj (commented):
I believe that the Status will return both a `code` and a `description`. Do you think it's safer to do the check on code rather than description? It's less likely to change.

Gweaton:
@andrewburrowsmoj Is the `code` just 'M' for main address? I wasn't sure if this was consistent across records. Although 'Main' could not be consistent either...

andrewburrowsmoj:
I really am not sure - I rasied the question in the delius channel:

https://mojdt.slack.com/archives/C01L8P3U9J8/p1631616770043600

andrewburrowsmoj (approved):

PR #798, opened 2021-09-10
Ic-1833 consistent url

rgforsyth:

Updates the pp post-session-feedback url to be consistent with the others.

All other url's are in the form `probation-practitioner/referral/:id/...` . This change is to bring the pp-post-session-feedback url in line as it was the only one not following this form.

Old: `probation-practitioner/:actionPlanId/appointment/:sessionNumber/post-session-feedback`
New: `probation-practitioner/referrals/:referralId/appointment/:sessionNumber/post-session-feedback`

### DIFF
--- a/server/routes/probationPractitionerReferrals/interventionProgressPresenter.test.ts
+++ b/server/routes/probationPractitionerReferrals/interventionProgressPresenter.test.ts
@@ -139,7 +139,7 @@ describe(InterventionProgressPresenter, () => {
                 classes: 'govuk-tag--green',
               },
               link: {
-                href: `/probation-practitioner/action-plan/c59809e0-ab78-4723-bbef-bd34bc6df110/appointment/1/post-session-feedback`,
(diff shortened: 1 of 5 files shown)


Gweaton (commented):
Nice - would you mind squashing the two commits together? They essentially have the same commit message and the first one introduces the method for getting a session that's later removed, so makes sense to take it out.

Gweaton (approved):
Thanks for squashing!

sldblog:
📝 From deploy huddle: we should figure out how to handle the old URLs after this change

Gweaton (changes requested):
Dismissing my last review to stop us merging by mistake as I didn't consider on the backwards compatibility side of this

Gweaton:
Closing this in favour of https://github.com/ministryofjustice/hmpps-interventions-ui/pull/830 which leaves the old routes in.

PR #800, opened 2021-09-10
View single case note

andrewburrowsmoj:

`bc88cc903704a15e1cccf73d82f6cc4dec3a29ad`
Added option to remove borders (horizontal lines) for summary list.

`8aecf9126fff5b7a69762895fe63cec052cc4b2c`
Add a new page to view a single case note detail for SP or PP.
    
The url to access this page matches what is sent to the user by email when the case note was created.

`cb405e43d616af99139a424051d1e781d70b99fc`
Added a link to case notes list to view single case note details as a 'Read More' link.
    
The body of the case note in the paginated table is restricted to 250 characters.
    
The `Read more` link is shown even if the body is less than 250 characters based on the comments in the ticket (IC-728).

Allow users to view case note details.

## Note

The back link for the case-note goes back to the first page for view all case note.

There will be a future commit which returns them to the correct page.

## Screen shot

**View all case notes with character < 250**
image

**View all case notes with character > 250**
image

**View single case note**

image

### DIFF
--- a/server/routes/caseNotes/viewAll/caseNotesView.ts
+++ b/server/routes/caseNotes/viewAll/caseNotesView.ts
@@ -1,6 +1,7 @@
 import { TableArgs } from '../../../utils/govukFrontendTypes'
 import CaseNotesPresenter from './caseNotesPresenter'
 import ViewUtils from '../../../utils/viewUtils'
+import PresenterUtils from '../../../utils/presenterUtils'
 
 export default class CaseNotesView {
   constructor(private presenter: CaseNotesPresenter) {}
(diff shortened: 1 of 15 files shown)


server/routes/caseNotes/viewAll/caseNotesView.ts
```diff
@@ -23,13 +23,23 @@ export default class CaseNotesView {
...
           },
         ]
       }),
     }
   }
 
+  private getCharacters(text: string, characterCount: number): string {
+    const truncatedCharacters = text.slice(0, characterCount)
+    if (truncatedCharacters.length === characterCount) {
+      return `${truncatedCharacters}...`
```
Gweaton, line 38: I feel like it would be good to have a test for this behaviour - we don't tend to test our view classes directly, other than in End to End tests. Could it be worth extracting this to a function in its own file (or like `PresenterUtils` or something similar) so we could write a test for it?

Gweaton (approved):
Nice work, just one thing re: testing.

PR #803, opened 2021-09-13
50: Simplify feature flagging logic and update documentation

Gweaton:

- Removes the now unused feature flag for Selecting an NPS office in the referral journey (it's visible in all environments)
- Makes the logic for feature flags a bit more user friendly by removing the nested `enabled` object, as this was missed by people writing code and reviewing code.
- Adds an ADR for why we use feature flags
- Updates README with instructions for how to add a feature flag

We had an unfinished feature slip through onto Production without us knowing - the changes to the config and instructions in the README should hopefully make it harder for that to happen again.

### DIFF
--- a/server/config.ts
+++ b/server/config.ts
@@ -43,9 +43,6 @@ export default {
   deploymentEnvironment: get('DEPLOYMENT_ENV', 'local', requiredInProduction),
   googleAnalyticsTrackingId: get('GA_ID', '', requiredInProduction),
   features: {
-    npsOfficeLocationSelection: {
-      enabled: get('FEATURE_NPS_OFFICE_LOCATION_SELECTION_ENABLED', 'true') === 'true',
-    },
     serviceProviderReporting: {
       enabled: get('FEATURE_SP_REPORTING_ENABLED', 'false') === 'true',
     },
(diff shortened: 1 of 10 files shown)


sldblog (commented):
😍 Thanks for making these changes

I'll defer approval to people closer to the codebase

server/config.ts
```diff
@@ -43,15 +43,8 @@ export default {
...
-    npsOfficeLocationSelection: {
-      enabled: get('FEATURE_NPS_OFFICE_LOCATION_SELECTION_ENABLED', 'true') === 'true',
-    },
-    serviceProviderReporting: {
-      enabled: get('FEATURE_SP_REPORTING_ENABLED', 'false') === 'true',
-    },
-    previouslyApprovedActionPlans: {
-      enabled: get('FEATURE_PREVIOUSLY_APPROVED_ACTION_PLANS', 'false') === 'true',
-    },
+    serviceProviderReporting: get('FEATURE_SP_REPORTING_ENABLED', 'false') === 'true',
```
andrewburrowsmoj, line 46: wonder if we can remove this as well (not for this PR though)

andrewburrowsmoj (approved):
lgtm - maybe we should double check in pre-prod before deploying?

PR #816, opened 2021-09-15
44: Add Referral date to My Cases table

Gweaton:

Adds the date of a referral to the PP My Cases table.

I initially tried to do a bit of a refactor to remove one of the two dashboard tables, as i'd really like to be able to use one table for both SPs and PPs, but due to the different fields required (and while we're using the `Summary`-type component for the SP view), it's been a bit of a pain building a one-size-fits-all approach, as some of the fields we need aren't on the Summary component yet.

We should revisit this once we've introduced pagination to the table, when we will possibly no longer have a need for the summary component, meaning we'll have access to all the fields on the referral.

In the first two rounds of research on the live service, PP participants have consistently told us that they would find Referral date more useful in My Cases than the Referral number, which is currently displayed in the left-hand column. 

They would regularly sort by this column as a way of monitoring which referrals need to be chased up.

## Screenshot

image

### DIFF
--- a/integration_tests/integration/probationPractitionerReferrals.spec.js
+++ b/integration_tests/integration/probationPractitionerReferrals.spec.js
@@ -88,6 +88,7 @@ describe('Probation practitioner referrals dashboard', () => {
       .getTable()
       .should('deep.equal', [
         {
+          'Date sent': '26 Jan 2021',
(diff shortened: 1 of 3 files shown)


andrewburrowsmoj (approved):
lgtm

PR #827, opened 2021-09-17
IC-2153: Swap order of enforceable days and completion date on make a referral form page.

andrewburrowsmoj:

Swaps order of enforceable days and completion date on make a referral form page.

Our PPs are consistently entering poor quality answers in the completion date input. Having enforceable days before completion date may improve quality.


integration_tests/integration/referral.spec.js
```diff
@@ -441,23 +441,23 @@ describe('Referral form', () => {
           `/referrals/${draftReferral.id}/service-category/428ee70f-3001-4399-95a6-ad25eaaede16/desired-outcomes`
         )
 
-      cy.contains('Accommodation completion date')
+      cy.contains('Enforceable days')
```
sldblog, line 444: Out of curiosity, does swapping this have an effect? I don't know the matcher syntax here but it looks like it doesn't actually make any sequence assertions on top level

andrewburrowsmoj: No, it does not have an effect; however I thought it best to change the order as it's more clear to the reader/dev the order in how it should appear on the page.

testutils/factories/draftReferral.ts
```diff
@@ -104,23 +104,23 @@ class DraftReferralFactory extends Factory<DraftReferral> {
     })
   }
 
-  filledFormUpToCompletionDate(serviceCategories: ServiceCategory[] = [serviceCategoryFactory.build()]) {
+  filledFormUpToEnforceableDays(serviceCategories: ServiceCategory[] = [serviceCategoryFactory.build()]) {
```
Gweaton, line 107: Did you change the order here to have the factory methods in the same order as the order of questions in the referral form, or is there another reason? If it's just the former, I'd say maybe no need next time we have to change the order of stuff in the referral form.

Gweaton (approved):

integration_tests/integration/referral.spec.js
sldblog: Thanks!

testutils/factories/draftReferral.ts
andrewburrowsmoj: yes that is correct. I realise that it's unnecessary, but I think it would help a bit in terms of reading the code as this section of the code can be a bit confusing.

PR #828, opened 2021-09-17
77: View submitted EOSR as an SP

Gweaton:

Adds a link to view an End of Service report as a Service Provider.

To allow the SP to view an end of service report, so they can read back what they've already sent to the PP.

### Link on SP intervention progress page

image

### Viewing EOSR (exactly the same content as the PP)

image


server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
```diff
@@ -1483,6 +1485,32 @@ export default class ServiceProviderReferralsController {
...
+    const serviceCategories = intervention.serviceCategories.filter(serviceCategory =>
+      referral.referral.serviceCategoryIds.some(serviceCategoryId => serviceCategoryId === serviceCategory.id)
+    )
+    if (serviceCategories.length !== referral.referral.serviceCategoryIds.length) {
+      throw new Error('Expected service categories are missing in intervention')
+    }
+
+    if (endOfServiceReport.submittedAt === null) {
+      throw new Error(
+        'You cannot view an end of service report that has not yet been submitted. Please submit the end of service report before trying to view it.'
```
andrewburrowsmoj, line 1502: It's good to add additional protection but I am wondering how would the user ever have access to this page?

The link to this page would only be shown when the EOSR is submitted and the UUID is impossible to guess?

andrewburrowsmoj (approved):
great changes

server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
Gweaton: It's definitely very unlikely - I thought best to err on the side of caution though to make it a bit more future-proof in case e.g. we decide to allow people to edit EOSRs.

PR #830, opened 2021-09-20
IC-1833: Update PP post session feedback url to be more consistent with other routes

Gweaton:

 Make PP Post Session Feedback URL consistent with other PP routes.

We're leaving the old URL and route in there for now, as we want to ensure that links in older emails to view Post Session Feedback still function and don't take users to a now 404ing page. We're going to monitor the old `/probation-practitioner/action-plan/:id/appointment/:id/post-session-feedback` URL and remove this duplicate route once traffic drops off.

All other URLs are in the form `probation-practitioner/referral/:id/...` This was the only one not following this form.

This replaces https://github.com/ministryofjustice/hmpps-interventions-ui/pull/798 as that didn't take into account that we needed to keep old links working.

### DIFF
--- a/server/routes/probationPractitionerRoutes.ts
+++ b/server/routes/probationPractitionerRoutes.ts
@@ -22,7 +22,11 @@ export default function probationPractitionerRoutes(router: Router, services: Se
     probationPractitionerReferralsController.showInterventionProgress(req, res)
   )
   get(router, '/referrals/:id/details', (req, res) => probationPractitionerReferralsController.showReferral(req, res))
+  // Legacy route to keep links in old emails still working. We'll monitor and remove once traffic drops off
   get(router, '/action-plan/:actionPlanId/appointment/:sessionNumber/post-session-feedback', (req, res) =>
+    probationPractitionerReferralsController.viewLegacySubmittedPostSessionFeedback(req, res)
+  )
+  get(router, '/referrals/:referralId/appointment/:sessionNumber/post-session-feedback', (req, res) =>
     probationPractitionerReferralsController.viewSubmittedPostSessionFeedback(req, res)
   )
   get(router, '/action-plan/:actionPlanId', (req, res) =>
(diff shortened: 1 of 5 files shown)


server/routes/probationPractitionerRoutes.ts
```diff
@@ -22,7 +22,11 @@ export default function probationPractitionerRoutes(router: Router, services: Se
     probationPractitionerReferralsController.showInterventionProgress(req, res)
   )
   get(router, '/referrals/:id/details', (req, res) => probationPractitionerReferralsController.showReferral(req, res))
+  // Legacy route to keep links in old emails still working. We'll monitor and remove once traffic drops off
   get(router, '/action-plan/:actionPlanId/appointment/:sessionNumber/post-session-feedback', (req, res) =>
+    probationPractitionerReferralsController.viewLegacySubmittedPostSessionFeedback(req, res)
+  )
+  get(router, '/referrals/:referralId/appointment/:sessionNumber/post-session-feedback', (req, res) =>
```
andrewburrowsmoj, line 29: I'm just wondering if there is plan for an equivalent change for SP as well (in a different ticket)?

I realise it's less important due to emails, but I do like how appointments are connected to a referral than an action plan.

andrewburrowsmoj (approved):
Nice changes.
LGTM

server/routes/probationPractitionerRoutes.ts
Gweaton: I think this was influenced by a conversation between @rgforsyth and @tomsmyers - there was a corresponding back-end change for the SP side too?

rgforsyth: Sorry, just seen this. The backend change for this is in the Step 3 backend pr that we plan to merge soon(pending review) https://github.com/ministryofjustice/hmpps-interventions-service/pull/578

We made this change as it was the only pp url used in the backend that wasn't referenced by referral. We wanted to make it consistent. I think the sp url's are in the same form?

PR #840, opened 2021-09-22
43: Preserve dashboards’ sort orders

lawrence-forooghian:
I'm nearing the end of my time on the project, and I don't think I'll have much time to work on this further. So, I'm opening this as a draft, with a list of the outstanding work.

1. Adds a TypeScript build process for the browser-side code, and converts the existing browser-side code to TypeScript.
2. Preserves the user's chosen ordering for the sortable tables (PP and SP case lists)

## What's left to do?

- [x] Write Cypress tests to check that this behaviour works
- [ ] Test that this works across browsers (I believe that the ECMAScript version I've chosen for transpilation will work all the way down to IE11, but not sure if we'll need a polyfill / to use an older alternative for some web APIs e.g. `MutationObserver`).

Currently, when the user sorts a sortable table, their chosen sort order is lost when they next return to that page (e.g. using the browser back button). Our user research has shown that users find this confusing, and suggested that we try preserving the sort order across page visits. That is what this change does.

## Video

https://user-images.githubusercontent.com/53756884/134332237-9ad71056-3c2d-41ac-a6d5-64b28f9c8cc2.mov

### DIFF
--- a/integration_tests/support/commands.js
+++ b/integration_tests/support/commands.js
@@ -1,3 +1,9 @@
+import deliusUserFactory from '../../testutils/factories/deliusUser'
+import deliusConvictionFactory from '../../testutils/factories/deliusConviction'
+import supplementaryRiskInformationFactory from '../../testutils/factories/supplementaryRiskInformation'
+import expandedDeliusServiceUserFactory from '../../testutils/factories/expandedDeliusServiceUser'
+import deliusOffenderManagerFactory from '../../testutils/factories/deliusOffenderManager'
+
 Cypress.Commands.add('login', () => {
   cy.request(`/`)
   cy.task('getLoginUrl').then(cy.visit)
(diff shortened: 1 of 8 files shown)


integration_tests/support/commands.js
```diff
@@ -15,6 +21,24 @@ Cypress.Commands.add('withinFieldsetThatContains', (text, action) => {
   cy.contains(text).parent('fieldset').within(action)
 })
 
+Cypress.Commands.add('stubViewReferralDetails', referralToView => {
```
andrewburrowsmoj, line 24: This is a great idea....

What do you think of making it into a discrete action of actually viewing the referral details page? 

This simply requires renaming it to viewReferralDetailsPage
and adding the command 
cy.visit(`/service-provider/referrals/${referralToSelect.id}/details`)
at the bottom

PR #841, opened 2021-09-22
Get supplementary risk information for crn

andrewburrowsmoj:

For Oasys risk information page the additional information is now populated from the ARN supplementary info call rather than populated from our own additional information data source.
This was an invalid assumption made regarding the source of the data.

Ensure that PP see's supplementary risk information before editing.

Note: Changes are behind a feature flag so this page would not be rendered until all the risk work is finished.


Gweaton (approved):
Looks good!

PR #842, opened 2021-09-22
Create script to ensure stable dependency boot

sldblog:

Creates a script that boots the dependencies in predictable, stable order.

TL;DR: Booting community-api first, then hmpps-auth, then the rest will ensure a stable boot order without excessive restarts and CPU load.

Without this, all services will boot up, and then randomly restart until the right startup order is achieved. This works, but  the laptop fans find it strenuous  :)

---

Every application is dependent on `hmpps-auth` being responsive;
otherwise, they end up with

```
org.springframework.web.client.ResourceAccessException: ⏎
  I/O error on GET request for ⏎
  "http://hmpps-auth:8090/auth/issuer/.well-known/openid-configuration"
```

In turn, `hmpps-auth` requires `community-api` to be responsive during
boot, resulting in
```
reactor.core.Exceptions$ReactiveException: ⏎
  java.net.UnknownHostException: failed to resolve 'community-api' after 2 queries
```

This forces the boot order described in the script.

## How does it look like?

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ You'll also need to add the following line to your `/etc/hosts` file:
 ### Running the app
 
 ```
-docker-compose up -d
+bin/start-dependencies
 npm run start:dev
 ```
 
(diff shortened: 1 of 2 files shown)


andrewburrowsmoj (approved):
This is awesome!
Thanks for this.

PR #845, opened 2021-09-23
Refactor classes with new reporting control

andrewburrowsmoj:
We discussed ages ago about splitting up the SP and PP referralController classes as they have a lot going on.

caseNotesController.ts is an example of some changes that were made outside of these controllers.

This PR is to move the reporting specific handlers into a reportingController.ts and move Presenter/View classes along with the new controller into its own dedicated directory.


Gweaton (approved):
Nice work, it's definitely nicer to have smaller, more self-contained controllers that are mostly focused on one thing.

PR #847, opened 2021-09-24
Add Delius unavailable broadcast message

Gweaton:

Adds Delius unavailable broadcast message from tonight until Sunday 26th

To inform users that they can't use the service to make referrals etc during this downtime.


Gweaton:
This wasn't deployed in time...

PR #850, opened 2021-09-27
Added new SP dashboards

andrewburrowsmoj:

Adds 4 dashboard types for logged in SPs.

These are:

1. My cases : Cases which are only assigned for the current logged in user
2. All open cases: Cases which are not completed and within the scope of the logged in user.
3. Unassinged cases: Cases which are not completed, within the scope of the logged in user and do not have an assigned user.
4. Completed cases: Cases which are completed and within the scope of the logged in user.

Makes it easier for SPs to locate the referral that they want.

## Notes:
 Count display for each dashboard and 2nd order filtering on `received at` will be in the next PRs.

## Screenshot
image

### DIFF
--- a/integration_tests/integration/dashboards.spec.js
+++ b/integration_tests/integration/dashboards.spec.js
@@ -289,61 +289,148 @@ describe('Dashboards', () => {
       title: "Women's Services - West Midlands",
     })
 
-    const sentReferralSummaries = [
-      serviceProviderSentReferralSummaryFactory.build({
-        sentAt: '2021-01-26T13:00:00.000000Z',
-        referenceNumber: 'ABCABCA1',
-        interventionTitle: accommodationIntervention.title,
-        serviceUserFirstName: 'George',
-        serviceUserLastName: 'Michael',
-      }),
-      serviceProviderSentReferralSummaryFactory.build({
-        sentAt: '2020-12-13T13:00:00.000000Z',
-        referenceNumber: 'ABCABCA2',
-        interventionTitle: womensServicesIntervention.title,
-        serviceUserFirstName: 'Jenny',
-        serviceUserLastName: 'Jones',
-        assignedToUserName: 'A.Caseworker',
-      }),
-    ]
-
     beforeEach(() => {
       cy.task('stubServiceProviderToken')
       cy.task('stubServiceProviderAuthUser')
 
       cy.stubGetIntervention(accommodationIntervention.id, accommodationIntervention)
       cy.stubGetIntervention(womensServicesIntervention.id, womensServicesIntervention)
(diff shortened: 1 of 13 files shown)


Gweaton (approved):
This looks great, nice one. Just a couple of comments but nothing major

server/models/serviceProviderSentReferralSummary.ts
```diff
@@ -6,4 +6,6 @@ export default interface ServiceProviderSentReferralSummary {
   assignedToUserName?: string | null
   serviceUserFirstName: string | null
   serviceUserLastName: string | null
+  endRequestedAt?: string | null
```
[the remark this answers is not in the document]
andrewburrowsmoj: I agree. I didn't realise we didn't have a contract for it.

I'll make sure to add one as part of this PR.

server/routes/serviceProviderRoutes.ts
```diff
@@ -17,7 +17,18 @@ export default function serviceProviderRoutes(router: Router, services: Services
     services.referenceDataService
   )
 
-  get(router, '/dashboard', (req, res) => serviceProviderReferralsController.showDashboard(req, res))
+  get(router, '/dashboard', (req, res) => serviceProviderReferralsController.showMyCasesDashboard(req, res))
```
Gweaton, line 20: I wonder if it would make sense to make this a redirect instead to reduce future bookmarking (maybe?):
```
  get(router, '/dashboard', (req, res) => res.redirect(301, `${serviceProviderUrlPrefix}/dashboard/my-cases`))
```

It would probably allow us to make the test for this smaller, testing that it's redirected, rather than duplicating the two tests.

Up to you though!

integration_tests/integration/dashboards.spec.js
```diff
@@ -410,193 +410,237 @@ describe('Dashboards', () => {
...
           interventionId: accommodationIntervention.id,
           serviceUser: { firstName: 'George', lastName: 'Michael' },
         },
       })
 
-      headersAndTables.forEach(({ header, sortedTable }) => {
-        describe(`sorting by "${header}"`, () => {
-          it(`allows the user to sort by "${header}"`, () => {
-            cy.login()
+      dashBoardTables.forEach(({ dashboardType, referrals, initialTable, sortedTables }) => {
```
Gweaton, line 605: How long do these tests take now, out of interest? I think this file took something like 18 seconds to run before. I guess this one of the drawbacks of E2E testing being slow. I think it's probably fine as we use these in place of unit tests for user-facing Javascript.

Gweaton (approved):
Nice work on this - just have a couple of small optional comments.

PR #853, opened 2021-09-29
Turn off create new action plan feature

Gweaton:

Turns off functionality to create new action plans after approval on pre-prod, as for some reason this was set to be `on`. We want this to mimic the prod environment as much as possible so we can test it.

So we don't put any experimental/incomplete functionality in front of users


jacksonj04 (approved):
LGTM 🚀

PR #857, opened 2021-09-30
Add a RoSH score summary panel to OASys risk page

jacksonj04:

Adds a small summary of the RoSH score to the OASys risk page

So probation practitioners can see a quick summary of a service user's RoSH score and be made aware of any risks.

localhost_3000_referrals_efd7cd24-e34d-49ac-a487-1e6b128c269e_risk-information


assets/sass/components/_rosh-analysis-table.scss
```diff
@@ -0,0 +1,24 @@
...
+  padding: 20px;
+
+
+  .last-updated {
+    color: $govuk-secondary-text-colour;
+    margin-bottom: 5px;
+  }
+}
+
+.rosh-header {
```
Gweaton, line 12: We can probably nest these in the `.refer-and-monitor-analysis-table` actually.

assets/sass/components/_rosh-analysis-table.scss
```diff
@@ -0,0 +1,24 @@
...
+  color: #942514;
+  margin-bottom: 0;
+}
+
+.rosh-header-low {
```
andrewburrowsmoj, line 17: Could we use BEM naming conventions here?
This is what MoJ CSS follows.

server/models/assessRisksAndNeeds/riskSummary.ts
```diff
@@ -24,5 +24,5 @@ export default interface RiskSummary {
     riskMitigationFactors?: string | null
     riskInCommunity: Partial<Record<RiskScore, string[]>>
   }
-  assessedOn: string
+  assessedOn?: string | null
```
andrewburrowsmoj, line 27: My understanding of the ARN contract is that this is never null?
Curious if you found it not to be the case?

server/routes/shared/riskPresenter.ts
```diff
@@ -23,10 +24,43 @@ export default class RiskPresenter {
...
+    const { riskInCommunity } = this.riskSummary.summary
+
+    const roshRankings = ['VERY_HIGH', 'HIGH', 'MEDIUM', 'LOW']
+
+    let i = 0
+
+    // For each roshRanking, check to see if it is defined. If not, step to the
+    // next, rinse and repeat.
+    while (riskInCommunity[roshRankings[i]] === undefined) {
+      i += 1
```
andrewburrowsmoj, line 53: Whilst we have a contract defined by ARN and there is an expectation that they should never violate it, if they do violate it then it shouldn't break our app.

I think here that there would be an infinite recursion on contract violation - for instance if they return something like `VERY_LOW`.

andrewburrowsmoj (commented):
Great stuff, just a few comments

server/models/assessRisksAndNeeds/riskSummary.ts
Gweaton: We were following the example with all this possibly being undefined above - do you think it's safe to rely on the field? and when you say this is never null, do you mean never `null` _or_ `undefined`?

andrewburrowsmoj: Ah I see that now - it's nice to have a clean contract...

... but I suppose based on previous experience with communityApi what they assert is not always the reality.

Gweaton: As discussed on slack, it looks like this could come back empty as it comes from the assessments API (not ARN) and could be undefined:

https://github.com/ministryofjustice/hmpps-assess-risks-and-needs/blob/cd927d4bda94ac24ca45332ee752d3dcc5119823/src/main/kotlin/uk/gov/justice/digital/hmpps/assessrisksandneeds/services/RiskService.kt#L31

server/routes/shared/riskPresenter.ts
```diff
@@ -51,6 +51,8 @@ export default class RiskPresenter {
     // next, rinse and repeat.
     while (riskInCommunity[roshRankings[i]] === undefined) {
       i += 1
+      // If we've run out of rankings, short circuit and return 'UNDEFINED'
+      if (i > roshRankings.length) return 'UNDEFINED'
```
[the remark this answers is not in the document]
Gweaton: Also, could we log an error here if this is the case? I'm hoping we'll know _before_ this happens, but if not, it'd be good for this not to silently display "undefined" to users without telling us. We can do this by adding something like:
```
logger.error({ err: `Unable to get ROSH score from risk in community response: ${ Object.keys(riskInCommunity) } })
```

It's not ideal, but would at least give us a list of the keys to show which one we're not supporting?

PR #859, opened 2021-10-01
8: Only show "Create End of service report" button when we should

Gweaton:

Uses the `endOfServiceReportCreationRequired`, now a field on the `SentReferral` object to determine whether or not we should show the `Create End of Service Report` button.

We were showing the button when we shouldn't have been. Now the backend is taking care of this logic for us, meaning we only need to have the logic for when an EOSR can be created in one place.

## Note

It's worth testing this out a fair bit on dev as I'm not 100% convinced it worked every time, but this may be due to incomplete / inconsistent data on the local environment.


andrewburrowsmoj (approved):
LGTM

PR #869, opened 2021-10-05
Case note badge for user type

andrewburrowsmoj:

Adds a user type badge to inform the reader if the case note creator was a probation practitioner or service provider.
The logic is based on authSource === 'delius' then the user is a PP. 

Provides further details to users the type of user who made the case note.

image

### DIFF
--- a/server/routes/caseNotes/viewAll/caseNotesPresenter.ts
+++ b/server/routes/caseNotes/viewAll/caseNotesPresenter.ts
@@ -12,6 +12,7 @@ interface CaseNotesTableRow {
   sentAtDate: string
   sentAtTime: string
   sentBy: string
+  sentByUserType: 'probation practitioner' | 'service provider'
   subject: string
   body: string
   caseNoteLink: string
@@ -61,6 +62,7 @@ export default class CaseNotesPresenter {
       sentAtDate: DateUtils.formattedDate(caseNote.sentAt),
       sentAtTime: DateUtils.formattedTime(caseNote.sentAt),
       sentBy: officerName,
+      sentByUserType: caseNote.sentBy.authSource === 'delius' ? 'probation practitioner' : 'service provider',
       subject: caseNote.subject,
       body: caseNote.body,
       caseNoteLink: `/${this.loggedInUserType}/case-note/${caseNote.id}`,
--- a/server/routes/caseNotes/viewAll/caseNotesView.ts
+++ b/server/routes/caseNotes/viewAll/caseNotesView.ts
@@ -11,7 +11,7 @@ export default class CaseNotesView {
     href: this.presenter.hrefBackLink,
   }
 
-  private get tableArgs(): TableArgs {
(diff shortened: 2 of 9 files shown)


server/routes/caseNotes/viewAll/caseNotesView.ts
```diff
@@ -24,7 +24,18 @@ export default class CaseNotesView {
...
         return [
           {
-            html: `<p class="govuk-body">${row.sentAtDay}<br>${row.sentAtDate}<br>${row.sentAtTime}<br>${row.sentBy}</p>`,
+            html: `<p class="govuk-body">
+          ${row.sentAtDay}
+          <br>${row.sentAtDate}
+          <br>${row.sentAtTime}
+          <br>${row.sentBy}
+          <br><br>${mojBadgeMacro({
+            text: row.sentByUserType,
```
tomsmyers, line 33: would be preferable to have all the macro args passed to the template

tomsmyers: that way the usertype logic does not bleed into the template

tomsmyers (approved):
nice change - would benefit from a cypress test

server/routes/caseNotes/viewAll/caseNotesPresenter.ts
```diff
@@ -61,6 +62,7 @@ export default class CaseNotesPresenter {
       sentAtDate: DateUtils.formattedDate(caseNote.sentAt),
       sentAtTime: DateUtils.formattedTime(caseNote.sentAt),
       sentBy: officerName,
+      sentByUserType: caseNote.sentBy.authSource === 'delius' ? 'probation practitioner' : 'service provider',
```
Gweaton, line 65: Do you think it'd be a good idea to wrap this `authSource === 'delius'` in a helper function somewhere? We currently repeat this check in a few places:

- server/routes/index.ts
- mockApis/auth.ts
- server/services/hmppsAuthService.ts

I feel like encapsulating it in an `isProbationPractitioner` function would hide the internals about an `auth_source` field that comes back from the API, so we could forget about it, and we'd only need to update it in one place if this value changed.

Ideally, I feel this would live on a `User` model, but we don't tend to do much manipulation of the values we get from the API. We've tended to solve this sort of thing with `Decorator`-type classes in the past, but maybe for now a helper function would do.

I don't think it should block this PR, we could do it in a separate tech-debt style card.

Gweaton (approved):
Looks good, just one thing

andrewburrowsmoj:
Tom - I have added Cypress tests for this PR

server/routes/caseNotes/viewAll/caseNotesPresenter.ts
andrewburrowsmoj: Yeah, I agree. I'll raise it as a new PR? 
We're running out of tickets to do for this sprint so will pick it up after this.

PR #886, opened 2021-10-11
Add polyfills for IE11-unsupported table sort JS

Gweaton:

Ensures functionality for remembering a user's chosen table sort order works on IE11

The new functionality for remembering the user's table sort wasn't working in IE11 and was throwing a few errors in the browser, due to a bunch of helpful new JS methods not being supported yet :(

## Video of this working in IE11

https://user-images.githubusercontent.com/19826940/136823037-a764cdc9-3900-424b-8d26-8f26bf557588.mov

### DIFF
--- a/browser/mojSortableTable.ts
+++ b/browser/mojSortableTable.ts
@@ -54,7 +54,8 @@ class PersistentSortOrder {
   private static observeAttributeChanges(tablePersistentId: string) {
     const headers = document.querySelectorAll('thead th')
 
-    headers.forEach(header => {
+    // IE11 doesn't allow us to call `forEach` on NodeListOf<Element>
+    Array.prototype.forEach.call(headers, header => {
       const headerAriaSortAttribute = header.getAttribute('aria-sort')
 
       if (headerAriaSortAttribute === null) {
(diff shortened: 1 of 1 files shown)


browser/mojSortableTable.ts
```diff
@@ -159,7 +162,25 @@ class PersistentSortOrder {
   private static forceSortOrder(table: HTMLElement, sortOrder: SortOrderDTO) {
     const headers = table.querySelectorAll('thead th') as NodeListOf<HTMLTableCellElement>
 
-    const header = Array.from(headers).find(aheader => aheader.dataset.persistentId === sortOrder.columnPersistentId)
+    let header: HTMLTableCellElement | undefined
+
+    // polyfill for IE11, which doesn't support `.from`
+    if (!Array.from) {
```
andrewburrowsmoj, line 168: What the benefit of having the if statement and instead just using the older functionality?

Gweaton: I left this one as-is (unlike the older functionality above) as it was easier to store this in a variable and use that conditionally.

I think the older functionality is a bit ugly, and not very readable, so thought I'd leave the newer version there for newer browsers.

I'm not actually sure what the best approach is, here - I wonder if there are performance implications for each implementation?

tomsmyers (approved):
LGTM

PR #899, opened 2021-10-14
add improved multi-function primary nav bar which supports different role-based user types

tomsmyers:

add improved multi-function primary nav bar which supports different role-based user types

allow any pages which use the nav bar to use the same shared code. 

allow differnt components to be added to the nav based on user roles, not just 'sp vs pp' types of users.

### DIFF
--- a/server/routes/shared/primaryNavBar/primaryNavBarPresenter.ts
+++ b/server/routes/shared/primaryNavBar/primaryNavBarPresenter.ts
@@ -1,13 +1,15 @@
-import { PrimaryNavItem } from '../../../utils/viewUtils'
 import config from '../../../config'
 import LoggedInUser from '../../../models/loggedInUser'
 
+export type PrimaryNavBarHeading = 'Referrals' | 'Reporting' | 'My cases' | 'Find interventions'
+export type PrimaryNavBarItem = { text: PrimaryNavBarHeading; href: string; active: boolean }
+
 export default class PrimaryNavBarPresenter {
-  constructor(private readonly active: string, private readonly loggedInUser: LoggedInUser) {}
+  constructor(private readonly active: PrimaryNavBarHeading, private readonly loggedInUser: LoggedInUser) {}
 
   // populate the nav bar items based on user roles
-  get items(): PrimaryNavItem[] {
-    const items = []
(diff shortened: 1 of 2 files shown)


server/routes/shared/primaryNavBar/primaryNavBarPresenter.ts
```diff
@@ -0,0 +1,44 @@
+import { PrimaryNavItem } from '../../../utils/viewUtils'
+import config from '../../../config'
+import LoggedInUser from '../../../models/loggedInUser'
+
+export default class PrimaryNavBarPresenter {
+  constructor(private readonly active: string, private readonly loggedInUser: LoggedInUser) {}
```
andrewburrowsmoj, line 6: Would be nice to have active as a `type` rather than `string` to make it a bit safer

i.e. 

export type navItem = `Referrals` | `Reporting` | `My cases` | `Find interventions`

server/routes/shared/primaryNavBar/primaryNavBarPresenter.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  // populate the nav bar items based on user roles
+  get items(): PrimaryNavItem[] {
+    const items = []
+
+    if (this.loggedInUser.token.roles.includes('ROLE_CRS_PROVIDER')) {
```
andrewburrowsmoj, line 12: Would also be nice to have a Decorator(? *) to obtain role information.

*at least that's what we named and use in other places i.e AppointmentDecorator

i.e. 

`if(userDecorator.hasCrsRole)`

andrewburrowsmoj (approved):
Just two optional comments.

server/routes/shared/primaryNavBar/primaryNavBarPresenter.ts
tomsmyers: do you think this is any clearer than the above? i think perhaps it obfuscates what the actual role->content mapping is?

andrewburrowsmoj: oh I was just thinking that we might perhaps need to do this in several places in future.

Maybe leave it until that is the case

PR #900, opened 2021-10-14
add nav bar item for 'My services' for users with valid role

tomsmyers:

add nav bar item for 'My services' for users with role `ROLE_INTERVENTIONS_SERVICE_EDITOR`

allow service editors to navigate to the 'My services' dashboard


tomsmyers:
note to self: needs int tests!

PR #904, opened 2021-10-18
PP Case lists

andrewburrowsmoj:
Adds the following dashboards for PP when they login:

- Open cases - Anything that is not concluded
- Unassigned cases - Anything that is not concluded and unassigned
- Completed cases - Anything that is concluded but not cancelled
- Cancelled cases - Anything that is cancelled

The filtering is done by the back-end by passing in filtering parameters.

The new dashboard looks like:

image

### DIFF
--- a/integration_tests/integration/dashboards.spec.js
+++ b/integration_tests/integration/dashboards.spec.js
@@ -79,6 +79,97 @@ describe('Dashboards', () => {
             },
           ])
       })
+
+      describe('Selecting "Unassigned cases"', () => {
+        it('should see "Unassigned cases"', () => {
+          cy.login()
+          cy.get('h1').contains('Open cases')
+          cy.contains('Unassigned cases').click()
+          cy.get('h1').contains('Unassigned cases')
+          cy.get('table')
+            .getTable()
+            .should('deep.equal', [
+              {
+                'Date sent': '26 Jan 2021',
+                Referral: 'ABCABCA1',
+                'Service user': 'George Michael',
+                'Intervention type': 'Accommodation Services - West Midlands',
+                Provider: 'Harmony Living',
+                Action: 'View',
+              },
+              {
+                'Date sent': '13 Dec 2020',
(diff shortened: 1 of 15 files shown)


server/routes/probationPractitionerReferrals/ppDashboardPresenter.ts
```diff
@@ -7,13 +7,15 @@ import { SortableTableHeaders, SortableTableRow } from '../../utils/viewUtils'
 import PrimaryNavBarPresenter from '../shared/primaryNavBar/primaryNavBarPresenter'
 import LoggedInUser from '../../models/loggedInUser'
 
-export default class MyCasesPresenter {
+export default class PPDashboardPresenter {
```
Gweaton, line 11: I tend to shy away from abbreviations (even well-known ones like `PP`) in class names as much as possible for readability. Bearing in mind this is in a directory called `ProbationPractitionerReferrals`, do you think we could just call it `DashboardPresenter` like we do for the SP view, and consider renaming in future if we move it out of this directory?

server/routes/probationPractitionerReferrals/ppDashboardPresenter.ts
```diff
@@ -7,26 +7,34 @@ import { SortableTableHeaders, SortableTableRow } from '../../utils/viewUtils'
...
     private readonly interventions: Intervention[],
-    private readonly loggedInUser: LoggedInUser
+    private readonly loggedInUser: LoggedInUser,
+    readonly dashboardType: PPDashboardType
   ) {}
 
-  readonly title = 'Open cases'
+  private readonly showAssignedCaseworkerColumn = this.dashboardType === 'Unassigned cases'
+
+  private readonly dashBoardTypePersistentId = `pp${this.dashboardType.replace(/\s/g, '')}`
```
Gweaton, line 21: Sorry, another pernickety naming comment here, could we make `dashboard` all lowercase here to keep it consistent with the `dashboardType`?

PR #917, opened 2021-10-20
[UOl5bHUM] Don't show `cancel` button for ended referrals

Gweaton:

Uses a referral's `concludedAt` and `endRequestedAt` fields to determine whether the PP should be able to cancel a referral.

We were previously showing the "cancel" button for referrals that had a submitted End of Service report.

The link that says "Cancel this referral" should not be visible as soon as:

- PP has prematurely ended an intervention OR
- The SP has submitted an EOSR after a complete intervention

### DIFF
--- a/server/views/probationPractitionerReferrals/interventionProgress.njk
+++ b/server/views/probationPractitionerReferrals/interventionProgress.njk
@@ -54,9 +54,9 @@
     <p>Below you will be able to find the end of service report created by the service provider. Once submitted, you will be able to read and download it.</p>
   {% endif %}
 
-  {% if presenter.referralEndRequested %}
-    <p class="govuk-body">{{ presenter.referralEndRequestedText }}</p>
-  {% else %}
+  {% if presenter.canCancelReferral %}
     <a href='{{ presenter.referralCancellationHref }}'>Cancel this referral</a>
+  {% elseif presenter.referralEndRequested %}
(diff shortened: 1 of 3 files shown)


server/views/probationPractitionerReferrals/interventionProgress.njk
```diff
@@ -54,9 +54,9 @@
     <p>Below you will be able to find the end of service report created by the service provider. Once submitted, you will be able to read and download it.</p>
   {% endif %}
 
-  {% if presenter.referralEndRequested %}
-    <p class="govuk-body">{{ presenter.referralEndRequestedText }}</p>
-  {% else %}
+  {% if presenter.canCancelReferral %}
     <a href='{{ presenter.referralCancellationHref }}'>Cancel this referral</a>
+  {% elseif presenter.referralEndRequested %}
```
tomsmyers, line 59: do we not want something here say if the referral was cancelled? and also something to tell them what this text actually is? like "service user died" on it's own might be a bit jarring?

Gweaton: `referralEndRequestedText`, which we're currently showing is "You requested to end this service on [this date]". We don't actually state the reason. I think it's more of catch-all phrase as the SP might have to complete an EOSR in some cases. Do you think I should run this past Helen, or do you think it's safe for now?

PR #918, opened 2021-10-21
Second order column

andrewburrowsmoj:
Adds a second order column to the dashboard tables.
This performs a second ordering on "Date sent" for PP dashboards and "Date received" for SP dashboards.
It will perform a second ordering on for rows which are identical.
The second ordering is always ascending sort direction.

## Ordered by `Referral`
image

## Reversed ordered by `Referral`
image

### DIFF
--- a/browser/mojSortableTable.ts
+++ b/browser/mojSortableTable.ts
@@ -17,7 +17,7 @@
  *
  * 4. The table can only be sorted by one column at a time.
  */
-
+// eslint-disable-next-line max-classes-per-file
 type ARIASort = 'none' | 'ascending' | 'descending'
 
 interface SortOrderDTO {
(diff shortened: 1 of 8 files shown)


browser/mojSortableTable.ts
```diff
@@ -239,6 +239,53 @@ class PersistentSortOrder {
...
 }
 
+class MOJFrontendTableSecondOrderSorter {
+  private secondOrderColumnNumber: number | undefined
+
+  constructor(private table: HTMLTableElement) {
+    // 'second-order-column' attribute set in server/utils/viewUtils.ts
+    const tableSecondOrderColumnAttr = this.table.getAttribute('second-order-column')
+    if (tableSecondOrderColumnAttr !== null) {
+      const parsedValue = parseInt(tableSecondOrderColumnAttr, 10)
```
Gweaton, line 249: Do you think we could call this something like `parsedColumnNumber`? Took a bit of time to figure out what this meant at first as I wasn't sure what `tableSecondOrderColumnAttr` was without checking `viewUtils`.

browser/mojSortableTable.ts
```diff
@@ -239,6 +239,53 @@ class PersistentSortOrder {
...
+      if (!Number.isNaN(parsedValue)) {
+        this.secondOrderColumnNumber = parsedValue
+      }
+    }
+  }
+
+  secondOrderSort(rows: HTMLTableRowElement[], firstOrderColumnNumber: number): HTMLTableRowElement[] {
+    return rows.sort((rowA: HTMLTableRowElement, rowB: HTMLTableRowElement) => {
+      if (this.secondOrderColumnNumber !== undefined) {
```
Gweaton, line 258: I wonder if it would be clearer to refactor this / break it up a bit to avoid the [arrow anti pattern](http://wiki.c2.com/?ArrowAntiPattern) - I found the multiple nested `if`s a little tricky to read. You could do this by returning early, possibly breaking things out into private functions e.g. here you could remove one level of nesting by doing:
```
if (this.secondOrderColumnNumber === undefined) {
  return 0
}
```

browser/mojSortableTable.ts
```diff
@@ -252,8 +299,22 @@ $(() => {
 
   const table = tables[0]
 
+  class SecondOrderSortableTable extends MOJFrontend.SortableTable {
```
Gweaton, line 302: Is it possible to move this out of the jquery function to stick it with the other class you've added, or is there a reason for it being here?

Gweaton (approved):
Nice work on this - definitely a good idea to add some unit tests in future.

I've added a couple of comments mostly about readability of the code, but I think this approach makes sense to me.

browser/mojSortableTable.ts
andrewburrowsmoj: 👍

PR #947, opened 2021-10-29
Refactor: Add Appointments Controller and move Supplier Assessment & Action Plan session appointments across (Part 1)

Gweaton:

Adds an Appointments Controller and moves functions for scheduling Supplier Assessment & Action Plan session appointments across to it.

Removes some of the attempts to re-use controller methods to make the code more DRY, as these made the code a bit less obvious as to what was going on. The methods are a fair bit longer, but at least more procedural so you can see the exact steps invovled in rendering a page. We may want to revisit this in future to make them a bit shorter and neater, but this feels a lot more readable now.

This is part of a larger refactor around appointments. We're planning to add more functionality around viewing previously-missed appointments and marking appointments that have already taken place in the past and cleaning up some of the DRYness here should help with that in future.

We're also moving to more single responsibility-focused controllers, rather than those based on user type (e.g. Service Provider / Probation Practitioner), as the user type controllers are huge and we'd like to keep things split out a bit.

## What's next?

- Move session feedback functionality from Service Provider and Probation Practitioner controllers to this new controller.
- Move PP functions for viewing appointments to this new controller.


server/routes/serviceProviderRoutes.ts
```diff
@@ -137,22 +145,22 @@ export default function serviceProviderRoutes(router: Router, services: Services
...
   )
   post(router, '/referrals/:id/supplier-assessment/schedule/:draftBookingId/submit', (req, res) =>
-    serviceProviderReferralsController.submitSupplierAssessment(req, res)
+    appointmentsController.submitSupplierAssessmentAppointment(req, res)
   )
   get(router, '/referrals/:id/supplier-assessment', (req, res) =>
-    serviceProviderReferralsController.showSupplierAssessmentAppointment(req, res)
+    appointmentsController.showSupplierAssessmentAppointment(req, res)
   )
   get(router, '/referrals/:id/supplier-assessment/scheduled-confirmation', (req, res) =>
```
andrewburrowsmoj, line 165: Perhaps this should also be moved into the appointments controller as well?

andrewburrowsmoj (approved):
Great change - keen to see this merged....

Left a comment about moving a method from service provider controller to appointment controller

server/routes/serviceProviderRoutes.ts
Gweaton: Oops, missed that one - will do that as part of this PR as it's the same journey. Thanks

PR #952, opened 2021-11-01
Include some cancelled referrals for Completed cases dashboard

andrewburrowsmoj:
Intervention-Service returns a concluded boolean flag for /sent-referrals/summary/service-provider to indicate if a referral has been concluded.

This flag should be used instead of endOfServiceReportSubmitted as we now want to include unfiltered cancelled referrals in the "Completed cases" dashboard. These referrals will not have an end of service report.

## Notes:
This is based on a conversation with Leigh where he provided the following table:
image

This shows that we should be including some cancelled referrals in the Completed cases. intervention service does the filtering so only the correct cancelled referrals will be returned.

### DIFF
--- a/server/models/serviceProviderSentReferralSummary.ts
+++ b/server/models/serviceProviderSentReferralSummary.ts
@@ -7,4 +7,5 @@ export default interface ServiceProviderSentReferralSummary {
   serviceUserFirstName: string | null
   serviceUserLastName: string | null
   endOfServiceReportSubmitted?: boolean
+  concluded: boolean
 }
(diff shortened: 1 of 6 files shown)


server/models/serviceProviderSentReferralSummary.ts
```diff
@@ -7,4 +7,5 @@ export default interface ServiceProviderSentReferralSummary {
   serviceUserFirstName: string | null
   serviceUserLastName: string | null
   endOfServiceReportSubmitted?: boolean
+  concluded: boolean
```
Gweaton, line 10: Do you think we should add this to the `sentReferralSummary` pact tests too? We can probably do that by using the updated factory here: https://github.com/ministryofjustice/hmpps-interventions-ui/blob/21aa181241b9c9b24b3f49d43ac8747293e400de/server/services/interventionsService.test.ts#L1693 instead of creating one from scratch. Would that break the Interventions Service pacts?

andrewburrowsmoj: Good point -- I'll have to do this change after the back-end has been approved.

sldblog:
_Comments from group review_
- Needs in-depth review together with https://github.com/ministryofjustice/hmpps-interventions-service/pull/637

PR #954, opened 2021-11-01
Refactor: move feedback viewing functionality for PPs to Appointments Controller

Gweaton:

Moves functionality for viewing Supplier Assessment and Action Plan session assessments as a PP to the new Apppointments Controller.

So we can reduce the size of the Probation Practitioner Referrals Controller as it's fairly bloated at the moment, and we'd like smaller, more focused controllers.


server/routes/appointments/appointmentsController.ts
```diff
@@ -746,14 +757,52 @@ export default class AppointmentsController {
...
+    return this.viewSubmittedPostSessionFeedback(res, referral, actionPlan.id, sessionNumber, null, 'service-provider')
+  }
+
+  async viewPostSessionFeedbackAsProbationPractitioner(req: Request, res: Response): Promise<void> {
+    const { accessToken } = res.locals.user.token
+    const { referralId, sessionNumber } = req.params
+
+    const referral = await this.interventionsService.getSentReferral(accessToken, referralId)
+
+    if (referral.actionPlanId === null) {
```
andrewburrowsmoj, line 776: Seems that there's no harm having this check on the `viewPostSessionFeedbackAsServiceProvider` so can be moved into the private method `viewSubmittedPostSessionFeedback`?
If you agree that also perhaps have a single method like with `viewSupplierAssessmentFeedback`?

andrewburrowsmoj (approved):
LGTM - I made a small comment; if you agree with it.

server/routes/appointments/appointmentsController.ts
Gweaton: I did initially try to re-use the function but I think my brain wasn't working too well yesterday. I've had another go and think it makes more sense now - I know you've already approved, but I don't suppose you'd mind taking another look as it'd be nice to get a second pair of eyes on the approach.

PR #956, opened 2021-11-02
FIX: Use cypress selector to click on correct element

Gweaton:

Adds a `data-cy` element to the attendance feedback list of radio buttons to ensure we're clicking on the correct element in the cypress tests.

This was failing due to our tests creating an object with the current date, using November as the month and trying to click on "No" and actually clicking on the month, rather than the "No" radio button.


andrewburrowsmoj (approved):
nice solution... lgtm

PR #962, opened 2021-11-04
[CnVfA8tT] Add functions and contracts for appointments linked to referrals, not action plans

Gweaton:

Adds new `/referral/:id/delivery-session-appointment/:id`-consuming functions to reflect the fact that Delivery Session appointments are no longer closely linked to an Action Plan. These no longer refer to an appointment by `sessionNumber` but by an `appointmentId`, now added to the `Appointment` interface.

Adds mocks for these to be used in Cypress tests.

This does not use the new functions or remove the old ones, it's merely to drive the API behaviour via consumer-driven contracts.

- Delivery sessions (what we used to call Action Plan sessions, renamed in the new functions) are no longer tied to an Action Plan, but are linked to a Referral.
- We're going to be displaying multiple appointments against a single session when the person on probation fails to attend an appointment, to allow SPs to reschedule appointments. This means we can no longer rely on the `sessionNumber` to represent an appointment, as we were previously doing. This change will allow us to fetch all appointments for a session and render them on the Intervention Progress page.

### DIFF
--- a/server/services/interventionsService.ts
+++ b/server/services/interventionsService.ts
@@ -383,6 +383,15 @@ export default class InterventionsService {
     })) as ApprovedActionPlanSummary[]
   }
 
+  async getDeliverySessionAppointments(token: string, referralId: string): Promise<ActionPlanAppointment[]> {
+    const restClient = this.createRestClient(token)
+    return (await restClient.get({
+      path: `/referral/${referralId}/delivery-session-appointments`,
+      headers: { Accept: 'application/json' },
+    })) as ActionPlanAppointment[]
+  }
+
+  // Deprecated in favour of `getDeliverySessionAppointments`
   async getActionPlanAppointments(token: string, actionPlanId: string): Promise<ActionPlanAppointment[]> {
     const restClient = this.createRestClient(token)
     return (await restClient.get({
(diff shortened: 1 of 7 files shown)


server/services/interventionsService.ts
```diff
@@ -373,6 +373,15 @@ export default class InterventionsService {
     })) as ApprovedActionPlanSummary[]
   }
 
+  async getDeliverySessionAppointments(token: string, referralId: string): Promise<ActionPlanAppointment[]> {
```
andrewburrowsmoj, line 376: hm, are we expecting to have initial assessment appointments being returned as well?

I think we should either:
1. Make this generic and have the appointment type defined in the DTO
2. Have it accept an `appointmentType` query param that performs filtering.
3. Change the url to make it more clear that only session-appointments are returned

Gweaton: We've changed this to `/delivery-session-appointments` to make it clear that it's just those types being returned.

sldblog:
_Comments from the group review_
- Needs a new owner to push it to merge

sldblog:
As this work is stale, let's close this until the deprecation cleanup comes up in priority again. Then we can reference and reuse this PR.

PR #970, opened 2021-11-08
OAsys risk information - refactor

andrewburrowsmoj:
Combined all oasys risk information details into a single parameter map.
I felt that the old parameter names with ResponseText suffix was too ambiguous about it's intent. This parameter was responsible for showing the No|Yes|Don't know response - esentially a label about the risk information.
I've now moved this parameter to be part of a larger paramter map that includes the risk information coupled with the risk label.

This will be also be used by edit oasys risk information page to avoid duplication with view risk information.

### DIFF
--- a/server/routes/makeAReferral/risk-information/oasys/view/oasysRiskInformationView.ts
+++ b/server/routes/makeAReferral/risk-information/oasys/view/oasysRiskInformationView.ts
@@ -3,6 +3,53 @@ import { Risk } from '../../../../../models/assessRisksAndNeeds/riskSummary'
 import RiskView from '../../../../shared/riskView'
 import { CheckboxesArgs, DetailsArgs, RadiosArgs } from '../../../../../utils/govukFrontendTypes'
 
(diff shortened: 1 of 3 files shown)


server/routes/makeAReferral/risk-information/oasys/view/oasysRiskInformationView.ts
```diff
@@ -3,54 +3,70 @@ import { Risk } from '../../../../../models/assessRisksAndNeeds/riskSummary'
 import RiskView from '../../../../shared/riskView'
 import { CheckboxesArgs, DetailsArgs, RadiosArgs } from '../../../../../utils/govukFrontendTypes'
 
+interface RiskInformationLabelArgs {
+  class: string
+  text: string
```
Gweaton, line 8: This could be `"yes" | "no" | "don't know"` rather than `string`

Gweaton (approved):
Looks good!

PR #974, opened 2021-11-10
Reword the did not attend initial assessment text

andrewburrowsmoj:
Reworded the did not attend initial assessment text from:
The initial assessment has been delivered and feedback added.

to:

Feedback needs to be added on the same day the assessment is delivered.


tomsmyers (approved):
LGTM- but a linked ticket would clarify that this is the correct change to make...

PR #976, opened 2021-11-10
supply query params to SP referrals summary API calls

tomsmyers:

adds query parameters to the '/sent-referrals/summary/service-provider' call to interventions service which specify which dashboard type is being loaded. 

the filtering in the UI is left in tact to ensure full backwards compatability with the existing API (however the API implementation should essentially render these as noops)

allow the UI to streamline queries for SP sent referral requests


andrewburrowsmoj (approved):
LGTM

PR #982, opened 2021-11-11
fix how we construct the query params for SP case lists

tomsmyers:

fix how we construct the query params for SP case lists

the reverse mapping on the enum isn't required for string type enums. before this change the query params are always `undefined`


andrewburrowsmoj (approved):
LGTM. Makes sense.

PR #985, opened 2021-11-12
[Hb65aVRv] Add new routes for new appointment endpoints for Delivery Sessions

Gweaton:

Adds new routes to `serviceProviderRoutes.ts` for viewing Session feedback and scheduling an appointment, now using the new backend structure of an Appointment no longer being linked to an Action Plan (what we're now calling a `DeliverySessionAppointment`). 

These aren't currently linked to by the Intervention Progress page, but in theory users could navigate to the URLs directly - it may be worth putting these behind a feature flag, but it's probably safe to assume they won't be able to guess the IDs etc.

The original routes are still left in place, as we'll want to make sure none of the older links break, e.g. if a user clicks a link in an email to view feedback, we don't want it to break the journey. We should monitor these old routes and remove them once we no longer receive traffic.

To pave the way for having multiple appointment linked to a Session, so we can allow users to reschedule appointments when an Appointment is marked as "Did Not Attend"

### DIFF
--- a/integration_tests/integration/appointmentFeedback.spec.js
+++ b/integration_tests/integration/appointmentFeedback.spec.js
@@ -1,0 +1,307 @@
+import actionPlanFactory from '../../testutils/factories/actionPlan'
+import actionPlanAppointmentFactory from '../../testutils/factories/actionPlanAppointment'
+import deliusServiceUserFactory from '../../testutils/factories/deliusServiceUser'
+import deliusUserFactory from '../../testutils/factories/deliusUser'
+import hmppsAuthUserFactory from '../../testutils/factories/hmppsAuthUser'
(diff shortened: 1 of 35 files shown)


sldblog:
_Comments from the group review_
- Needs a new owner to push it to merge

andrewburrowsmoj:
I have added a commit to this PR which aimed to use the updated back-end end points, unfortunately I was unable to resolve it in time.

Apologies that the last commit squashes a few responsibilities in one and does not have passing tests. I felt it was best to commit what I was able to accomplish rather than leave it uncommitted. 

In order to test this locally, you will need to navigate to the endpoints manually by typing in the url. 
The old ui endpoints should remain as they are, unaffected; and the referral progress page points to the old ui end-points.

PR #993, opened 2021-11-16
[M6tsQ6Au] Improvements to design of the RoSH summary box

jacksonj04:

Updates the RoSH summary box so that it better matches the original designs from ARN by:

- [x] Double-checking that the border and title colour of the box match the highest risk level
- [x] Switching display of actual risk levels to coloured plain text instead of using GOV.UK tag elements
- [ ] Consistently sorts the display order of risks

This improves the consistency of how risk levels are displayed across all endpoints.


server/routes/shared/roshPanelView.ts
```diff
@@ -17,26 +17,40 @@ export default class RoshPanelView {
...
+        return 'rosh-analysis-table__risk-score__high'
       case 'VERY_HIGH':
-        return 'govuk-tag--purple'
+        return 'rosh-analysis-table__risk-score__very-high'
+      default:
+        return ''
+    }
+  }
+
+  private stringForRiskScore(riskScore: string): string {
```
tomsmyers, line 44: nit: it worries me we may end up not displaying any score at all if say, suddently the 'VERY_HIGH' score was returned  as 'VERY HIGH' (i know, not super realistic).. wonder if we can use some rules to convert the input string instead of this switch. e.g. replace '_' with space, and then 'proper case' the result

tomsmyers (approved):
LGTM - would benefit from unit tests for the risk group ordering malarkay

PR #1001, opened 2021-11-19
Edit oasys risk information  draft oasys risk information

andrewburrowsmoj:
This PR introduces the ability to populate the text areaboxes of the `edit oasys risk information` page with draft OAsys risk information. This may be non-null if the user has revisited the page after clicking on `save and continue`. The actual submission to ARN occurs at `check your answers` page, we are using a draft object until that point has been reached.

In addition, a revert `button` allows the user to replace the edited text area with the latest ARN OAsys data.

image

Here is the commit overview for this PR:

commit d240aa57254f79bf7546be4d6b3220e76f32db45 

Added javascript revert functionality for text area boxes on edit oasys risk information page.

When activated the text area box will be replaced with content from OAsys.

commit d14ac10257e88a564281323e3010f13948bac116
DraftOasysRiskInformation populates text area boxes on edit oasys risk information page if it has been persisted to the backend; otherwise it will show the default OASys risk information.

commit c591f7936086fa156b7e2e30c9bdc92aa15b8484

Added GET /referrals/{id}/oasys-risk-information endpoint for interventions service

### note

These changes are behind a feature flag

### DIFF
--- a/server/routes/makeAReferral/makeAReferralController.ts
+++ b/server/routes/makeAReferral/makeAReferralController.ts
@@ -56,6 +56,7 @@ import EditOasysRiskInformationPresenter from './risk-information/oasys/edit/edi
 import DraftReferral from '../../models/draftReferral'
 import ConfirmOasysRiskInformationForm from './risk-information/oasys/confirmOasysRiskInformationForm'
 import EditOasysRiskInformationForm from './risk-information/oasys/edit/editOasysRiskInformationForm'
+import { DraftOasysRiskInformation } from '../../models/draftOasysRiskInformation'
 
 export default class MakeAReferralController {
   constructor(
@@ -575,6 +576,7 @@ export default class MakeAReferralController {
     const { accessToken } = res.locals.user.token
     const referralId = req.params.id
     let error: FormValidationError | null = null
+    let draftOasysRiskInformation: DraftOasysRiskInformation | null = null
     if (req.method === 'POST') {
       const form = await EditOasysRiskInformationForm.createForm(req)
       if (form.isValid) {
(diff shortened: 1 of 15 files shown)


server/routes/makeAReferral/makeAReferralController.ts
```diff
@@ -587,6 +589,21 @@ export default class MakeAReferralController {
...
+      draftOasysRiskInformation = form.editedDraftRiskInformation
+    } else {
+      try {
+        draftOasysRiskInformation = await this.interventionsService.getDraftOasysRiskInformation(
+          accessToken,
+          referralId
+        )
+      } catch (e) {
+        const restClientError = e as RestClientError
+        if (restClientError.status === 404) {
```
tomsmyers, line 601: this looks like a guard against the endpoint not existing in the API - is this nessecary so long as we have contract tests?

andrewburrowsmoj: The first time the user accesses the page the service call will always return NOT_FOUND.

When they click 'save and continue' it saves it in the db and when they return (i.e. from check answers) then we'd expect a draft oasys risk information to be returned.

tomsmyers: 👍

tomsmyers (approved):
nice solution - two things I would like to see (but not blockers)

1. a non-js fallback (reload the page with the text replaced?)
2. a cypress test for the js functionality

andrewburrowsmoj:
@tomsmyers regarding your points:

ARN have not provided a solution for non-js solution; I think by refreshing the page to have text replaced is dangerous since it would remove all their changes and may not be obvious to them.. 
Cypress solution will be in a future PR.

sldblog:
Feels like the revert button can be a separate PR? Not a blocker though.

sldblog:
💭 We could have a non-JS functionality if we have the 'official' risk field as an immutable text somewhere + treat this field as "not relevant" if empty (probably something for the next iteration)

PR #1002, opened 2021-11-19
Bug fix for PP view supplier assessment feedback

andrewburrowsmoj:
There is a bug which prevents PP user from accessing viewing supplier assessment feedback.

This bug is caused because the route specifies the referral id as `:id` yet the controller code expects this to be named `:referralId`.

https://github.com/ministryofjustice/hmpps-interventions-ui/blob/2b2b6de83078cec8496c17860535497c5511d840/server/routes/appointments/appointmentsController.ts#L563


sldblog (approved):
👍

sldblog:
I wonder if we could validate these with some tests?

PR #1009, opened 2021-11-22
Removed filtering for SP dashboards.

andrewburrowsmoj:
Removed filtering for SP dashboards. Filtering is done for us by the backend already.
By keeping the filtering in place it restricts us from changing the filtering strategy on the back end.

This is in preparation for including some cancelled referrals for Completed cases.


tomsmyers (approved):
LGTM

PR #1027, opened 2021-11-25
[Xw3DqmK3] do not allow years bigger than 9999 in forms

tomsmyers:
https://trello.com/c/Xw3DqmK3

do not allow years bigger than 9999 in forms

stop occurances of this error https://sentry.io/organizations/ministryofjustice/issues/2510325235/?project=5807819&project=5807822&query=is%3Aunresolved&statsPeriod=14d

### DIFF
--- a/server/utils/forms/inputs/calendarDayInput.ts
+++ b/server/utils/forms/inputs/calendarDayInput.ts
@@ -92,7 +92,7 @@ export default class CalendarDayInput {
         .notEmpty({ ignore_whitespace: true })
         .withMessage(this.errorMessages.yearEmpty)
         .bail()
-        .isInt()
+        .isInt({ max: 9999 }) // sure, dates bigger than this are real dates, but they're not valid here
         .withMessage(this.errorMessages.invalidDate),
     ]
   }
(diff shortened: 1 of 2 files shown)


server/utils/forms/inputs/calendarDayInput.ts
```diff
@@ -92,7 +92,7 @@ export default class CalendarDayInput {
         .notEmpty({ ignore_whitespace: true })
         .withMessage(this.errorMessages.yearEmpty)
         .bail()
-        .isInt()
+        .isInt({ max: 9999 }) // sure, dates bigger than this are real dates, but they're not valid here
```
andrewburrowsmoj, line 95: suggestion to add this to `calendarDay.ts::isValid()`. We can globally invalidate all years that are greater than 9999.

andrewburrowsmoj: simple unit test would be nice

andrewburrowsmoj (approved):

server/utils/forms/inputs/calendarDayInput.ts
tomsmyers: yeh i toyed with this - but i wasn't sure it was apprpriate there. that method is checking if the date itself is valid. dates in 20222 kinda _are_ valid... but we don't want them passed through our form, which is why i ended up putting it here.

tomsmyers: added unit test

PR #1032, opened 2021-11-26
Made redactedRisk as part of supplementaryRisk to be an optional value.

andrewburrowsmoj:
This value is undefined if supplementaryRisk has never been set.

_Required._

_Required._


sldblog:
Should a test be added that breaks if someone puts this back to non-optional?

andrewburrowsmoj:
@sldblog added a test

sldblog (approved):
Thanks! 👍

PR #1033, opened 2021-11-26
Remove supplementary risk info when editing oasys

andrewburrowsmoj:
commit 364b6e3c72660eb2740451843374b6e90b348146

Removed getSupplementaryRiskInformationForCrn() from assessRisksAndNeedsService.
We should always be getting supplementary risk based on the referral id rather than CRN.

commit df382e4fda5af7f7db9dd5f92e1ee0db70a04c11

Removed additional information from supplementary risk for CRN when viewing OAsys risk information as part of make a referral journey. This was coded with an invalid assumption that the supplementary risk for CRN should be displayed. Only original OAsys data should be displayed on this page.

### DIFF
--- a/server/routes/makeAReferral/risk-information/oasys/arnRiskSummaryView.test.ts
+++ b/server/routes/makeAReferral/risk-information/oasys/arnRiskSummaryView.test.ts
@@ -1,0 +1,182 @@
+import riskSummaryFactory from '../../../../../testutils/factories/riskSummary'
+import supplementaryRiskInformationFactory from '../../../../../testutils/factories/supplementaryRiskInformation'
+import { Risk } from '../../../../models/assessRisksAndNeeds/riskSummary'
+import ArnRiskSummaryView from './arnRiskSummaryView'
+
+describe('ArnRiskSummaryView', () => {
+  describe('supplementaryRiskInformationArgs', () => {
+    describe('additionalRiskInformation', () => {
+      it('returns no information label if no additional risk information has been provided', () => {
+        const riskSummary = riskSummaryFactory.build()
+        const view = new ArnRiskSummaryView(
+          riskSummary,
+          supplementaryRiskInformationFactory.build({
+            riskSummaryComments: undefined,
+          })
+        )
+        expect(view.supplementaryRiskInformationArgs.additionalRiskInformation.label).toHaveProperty('text', 'None')
+      })
+    })
+
+    describe('summary', () => {
+      it('returns null text when undefined redactedRisk provided', () => {
+        const view = new ArnRiskSummaryView(
+          riskSummaryFactory.build(),
+          supplementaryRiskInformationFactory.build({ redactedRisk: undefined })
+        )
+        expect(view.supplementaryRiskInformationArgs.summary.whoIsAtRisk.text).toBeNull()
+        expect(view.supplementaryRiskInformationArgs.summary.natureOfRisk.text).toBeNull()
+        expect(view.supplementaryRiskInformationArgs.summary.riskImminence.text).toBeNull()
+        expect(view.supplementaryRiskInformationArgs.riskToSelf.suicide.text).toBeNull()
+        expect(view.supplementaryRiskInformationArgs.riskToSelf.selfHarm.text).toBeNull()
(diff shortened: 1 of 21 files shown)


sldblog:
> when editing oasys

this is picking at terminology, which I think will come up in later meetings, but could this be specific to _what_ is being edited, please?

we can't edit oasys directly and it might be misleading if this `remove-supplementary-risk-info-when-editing-oasys` changelog appears

server/routes/makeAReferral/risk-information/oasys/oasysRiskSummaryView.test.ts
```diff
@@ -6,10 +6,11 @@ import { Risk } from '../../../../models/assessRisksAndNeeds/riskSummary'
 describe('OasysRiskSummaryView', () => {
   describe('oasysRiskInformationArgs', () => {
     describe('additionalRiskInformation', () => {
-      it('returns special content to display if no additional risk information has been provided', () => {
+      it('always returns null text and undefined label', () => {
```
sldblog, line 9: could this include why this is preferred?

server/routes/makeAReferral/risk-information/oasys/oasysRiskSummaryView.ts
```diff
@@ -103,8 +103,8 @@ export default class OasysRiskSummaryView {
         },
       },
       additionalRiskInformation: {
-        label: this.supplementaryRiskInformation ? undefined : this.noAdditionalRiskInformationLabel,
-        text: this.supplementaryRiskInformation ? this.supplementaryRiskInformation.riskSummaryComments : null,
+        label: this.noAdditionalRiskInformationLabel,
+        text: null,
```
sldblog, line 107: Could we remove the parameter `supplementaryRiskInformation` altogether, if we're not using it?

sldblog:
```
server/routes/shared/showReferralView.ts
16:    this.supplementaryRiskInformationView = new OasysRiskSummaryView(presenter.riskInformation, presenter.riskSummary)
```
seems to be still passing it through but it won't have any effect it seems

server/routes/makeAReferral/risk-information/oasys/oasysRiskSummaryView.test.ts
andrewburrowsmoj: Thanks I've adjusted the tests in this PR to be more descriptive.

server/routes/makeAReferral/risk-information/oasys/oasysRiskSummaryView.ts
andrewburrowsmoj: Yes, you're right. I've improved on this by splitting the functionality into two separate classes:

OAsysRiskInformation and ArnRiskInformation since they're effectively two different things.

server/routes/makeAReferral/risk-information/oasys/arnRiskSummaryView.test.ts
```diff
@@ -0,0 +1,182 @@
...
+        expect(view.supplementaryRiskInformationArgs.summary.whoIsAtRisk.text).toBeNull()
+        expect(view.supplementaryRiskInformationArgs.summary.natureOfRisk.text).toBeNull()
+        expect(view.supplementaryRiskInformationArgs.summary.riskImminence.text).toBeNull()
+        expect(view.supplementaryRiskInformationArgs.riskToSelf.suicide.text).toBeNull()
+        expect(view.supplementaryRiskInformationArgs.riskToSelf.selfHarm.text).toBeNull()
+        expect(view.supplementaryRiskInformationArgs.riskToSelf.hostelSetting.text).toBeNull()
+        expect(view.supplementaryRiskInformationArgs.riskToSelf.vulnerability.text).toBeNull()
+      })
+
+      it('returns correct text when supplementary information provided', () => {
```
sldblog, line 36: _not a blocker_
🙋‍♂️ would be useful to define what is "correct" in case this test regresses -- if I were trying to fix this regression, would I have enough knowledge about what is correct?

sldblog (approved):
Thanks, 👍 looks good to me
No blockers. Perhaps would benefit from a 👍 from someone who's better at node though 😅

PR #1042, opened 2021-11-30
Switch UI deploy to the new cluster on dev

sldblog:

Switch dev environment deployment to the new cluster.

🚨 If this is successful the ingress/deployment/pods will be **removed** from live-1 (manually)

This is controlled by `ingress.contextColour` values in `values-{env}.yaml`

- 🔵 "blue": goes to the current cluster
- 🟢 "green": goes to the new cluster

## Background work

Manually copied all unprovisioned secrets over to the new cluster that blocked boot (`session`)

We need to migrate over to the new cluster this year.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -178,6 +178,7 @@ workflows:
             - vulnerability_scan_and_monitor
           context:
             - hmpps-common-vars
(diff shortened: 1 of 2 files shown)


sldblog:
Closing to open for both dev and preprod

PR #1048, opened 2021-12-01
Show error message for reassign referral

andrewburrowsmoj:
No error message is shown when attempting to reassign a referral and the email address is not found.

## Screenshot of what the error message should show
image


sldblog (approved):
Should this have a test? It sounds like that could've caught the regression that `assignee` is no longer `null`

PR #1055, opened 2021-12-02
Action plan activities out of order for new revised action plan

andrewburrowsmoj:
We have a bug with creating a new action plan from an old one whereby the activities on the new action plan are created out of order compared with the old action plan.

This fix is to order the activities before submitting them to create the new draft action plan.

### DIFF
--- a/server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
+++ b/server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
@@ -818,9 +818,11 @@ export default class ServiceProviderReferralsController {
       accessToken,
       sentReferral.id,
       existingActionPlan.numberOfSessions || undefined,
-      existingActionPlan.activities.map(it => {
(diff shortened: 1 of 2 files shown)


server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
```diff
@@ -818,9 +818,11 @@ export default class ServiceProviderReferralsController {
       accessToken,
       sentReferral.id,
       existingActionPlan.numberOfSessions || undefined,
-      existingActionPlan.activities.map(it => {
-        return { description: it.description }
-      })
+      existingActionPlan.activities
+        .sort((a, b) => (a.createdAt < b.createdAt ? -1 : 1))
```
andrewburrowsmoj, line 822: using same logic used in `orderedActivities()` from `actionPlanPresenter.ts`

sldblog (approved):

PR #1057, opened 2021-12-03
Display caseworker details for appointment feedback

andrewburrowsmoj:
Changed the appointment feedback page to show the firstname, lastname and email address of caseworker for the user who is assigned and the user who submitted the feedback.

These values will always be SP users so it makes sense to only call hmppsAuth to obtain the caseworker details; however if there were any issues in obtaining this information then the username is used as a fallback.

## Screenshot 

image

### DIFF
--- a/server/routes/appointments/appointmentsController.ts
+++ b/server/routes/appointments/appointmentsController.ts
@@ -1,7 +1,11 @@
 import { Request, Response } from 'express'
 import createError from 'http-errors'
 import SupplierAssessmentDecorator from '../../decorators/supplierAssessmentDecorator'
-import { AppointmentSchedulingDetails, InitialAssessmentAppointment } from '../../models/appointment'
+import {
+  ActionPlanAppointment,
+  AppointmentSchedulingDetails,
+  InitialAssessmentAppointment,
+} from '../../models/appointment'
 import DeliusOfficeLocation from '../../models/deliusOfficeLocation'
 import AuthUserDetails from '../../models/hmppsAuth/authUserDetails'
 import CommunityApiService from '../../services/communityApiService'
@@ -41,6 +45,8 @@ import AttendanceFeedbackView from './feedback/shared/attendance/attendanceFeedb
 import BehaviourFeedbackForm from './feedback/shared/behaviour/behaviourFeedbackForm'
 import BehaviourFeedbackView from './feedback/shared/behaviour/behaviourFeedbackView'
 import CheckFeedbackAnswersView from './feedback/shared/checkYourAnswers/checkFeedbackAnswersView'
+import logger from '../../../log'
+import SentReferral from '../../models/sentReferral'
 
 export type DraftAppointmentBooking = null | AppointmentSchedulingDetails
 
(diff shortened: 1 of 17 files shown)


andrewburrowsmoj:
Failing on "Expected to find content: 'LOW COMPLEXITY' but never did." 

Unrelated to changes I've done.

server/routes/appointments/appointmentsController.ts
```diff
@@ -755,6 +755,18 @@ export default class AppointmentsController {
     )
   }
 
+  private async getCaseWorker(accessToken: string, username: string): Promise<AuthUserDetails | string> {
+    try {
+      return this.hmppsAuthService.getSPUserByUsername(accessToken, username)
+    } catch (e) {
+      const interventionsServiceError = e as InterventionsServiceError
+      if (interventionsServiceError.status === 404) {
```
tomsmyers, line 763: do we wanna log a warning or 'invalid assumption' when we hit 404 here? 

i'm guessing we never expect this to happen?

server/utils/summaryList.ts
```diff
@@ -3,9 +3,16 @@ export enum ListStyle {
   bulleted,
 }
 
+interface Caseworker {
```
tomsmyers, line 6: this seems like a strange place to define this interface. 

do we need it at all? can we just use `AuthUserDetails`?

tomsmyers (approved):
nice work - couple of small things noted. 

one thing for me personally, unrelated to the technical changes, is it really doesn't feel clear to me what 'caseworker' is in this context... is it always the currently assigned caseworker? is it the caseworker that was assigned when the appointment feedback was submitted? is it the caseworker that attended the appointment? etc etc. i would like to just quickly loop back around with content on this before pushing to prod.

server/routes/appointments/appointmentsController.ts
andrewburrowsmoj: Sure. I'll add a log here.

jacksonj04 (approved):
Thanks for looping me in, nice to see how this is solved without the rabbit hole I ended up in!

No comments other than those from @tomsmyers, and an agree that "Caseworker" language here isn't particularly great (although not sure it needs fixing in this PR vs just raising another to clarify exactly what it means in any context where "caseworker right now" and "caseworker when this thing was done" might vary.

server/utils/summaryList.ts
andrewburrowsmoj: yeah I had this originally, but felt that it was better to have something less specific and less coupled. 
I've replaced Caseworker with AuthUserDetails.

server/routes/appointments/appointmentsController.ts
andrewburrowsmoj: done

andrewburrowsmoj:
@tomsmyers thanks for your feedback - given that this PR doesn't change the **Caseworker** title in the table for appointment feedback I have merged the changes.
This PR only touches on the name of the caseworker from **username** to **firstName + lastName + email**

I have raised your concern in the interventions-crs-team channel.

PR #1061, opened 2021-12-06
uCSWMBxp: Change case notes return link

andrewburrowsmoj:
The link for "Return to intervention" one confirmation of creating a new case note should return the user to the case notes page rather than interventions progress page.


jacksonj04 (approved):
👍🏼

PR #1085, opened 2021-12-13
CI/security: Replace snyk scannning with trivy scanning

sldblog:

Replace snyk scannning with trivy scanning

🔐 To make this pass, pinned `validator` to mitigate CVE-2021-3765

As an added bonus, now with Slack alerts if nightly scan fails :)

Snyk has been replaced with Trivy as it's faster, smaller, and better
integrated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
       "prettier --write"
     ]
   },
-  "//": "tar is not a direct dependency of our project; we’ve only added it here to remove a security vulnerability introduced by a transitive dependency. This should be removed; see IC-2177",
+  "//": "validator is pinned for CVE-2021-3765; remove when fixed upstream",
   "dependencies": {
     "@ministryofjustice/frontend": "1.0.1",
     "@sentry/node": "^6.16.1",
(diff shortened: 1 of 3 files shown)


sldblog:
📝 This will **fail** with good reason

image

package.json
```diff
@@ -91,7 +91,7 @@
       "prettier --write"
     ]
   },
-  "//": "tar is not a direct dependency of our project; we’ve only added it here to remove a security vulnerability introduced by a transitive dependency. This should be removed; see IC-2177",
+  "//": "validator is pinned for CVE-2021-3765; remove when fixed upstream",
```
jacksonj04, line 94: Is it worth making this a `TODO` so that it shows up when searching for things we need to look at?

jacksonj04 (approved):
One comment about making sure we pick up unpinning validator in the future, otherwise 🚢

package.json
sldblog: 👍 added in c2562d1500f2bf1fdd99819c1c24925038e7b39c

PR #1093, opened 2021-12-14
Past appointments

rgforsyth:
When scheduling a delivery session appointment, if the user provides an appointmentTime that is in the past then the user is directed to a new flow where they will need to enter in session feedback details.

The full journey is using the Redis draft service so only at the final confirmation page are both scheduled appointment details and appointment feedback details submitted to the back end.

This means that the full appointment details for past appointments is sent as a single request to the backend rather than as several requests.

The existing appointment feedback flow remains untouched; only for past appointments is the draft service used.

Care has been taken to ensure that the draft for the appointment scheduling flow does not conflict with the new draft which include feedback details.
In order to ensure that old drafts can work, we've appended feedback details as a new undefinable field in the existing draft so it should be compatible with the old and new draft.


server/routes/appointments/appointmentsController.ts
```diff
@@ -287,7 +291,23 @@ export default class AppointmentsController {
         formError = data.error
         userInputData = req.body
       } else {
-        await this.draftsService.updateDraft(draft.id, data.paramsForUpdate, { userId: res.locals.user.userId })
+        const draftAppointment: DraftAppointment = {
+          ...data.paramsForUpdate,
+          sessionFeedback: {
```
andrewburrowsmoj, line 296: we shouldn't be setting this here - this is because it's possible for a user to be on `check-your-answers` page before submitting and it will direct them to attendance page but the draft won't have an empty sessionFeedback.

We should instead be doing this on the attendance page.

andrewburrowsmoj:
still need write new pact tests for past appointments

server/routes/appointments/appointmentsController.ts
```diff
@@ -890,4 +982,97 @@ export default class AppointmentsController {
...
+      return {
+        sessionFeedback: feedbackDetails,
+        sessionNumber: Number(sessionNumber),
+        ...appointmentSchedulingDetails,
+      }
+    }
+    return null
+  }
+
+  private async getActionPlanAppointment(req: Request, res: Response): Promise<ActionPlanAppointment | null> {
```
andrewburrowsmoj, line 1054: rename to `getActionPlanAppointmentFromDraftOrService`

server/routes/appointments/appointmentsController.ts
```diff
@@ -668,76 +721,115 @@ export default class AppointmentsController {
...
 
-    return res.redirect(
+      if (this.isADraftAppointment(draft.data)) {
+        await this.interventionsService.recordAndSubmitActionPlanAppointmentWithFeedback(
+          accessToken,
+          actionPlanId,
+          Number(sessionNumber),
+          draft.data
+        )
+      }
```
andrewburrowsmoj, line 827: add: 
1. else throw exception if not draft appointment
2. delete draft

server/views/appointments/feedback/shared/postSessionFeedbackCheckAnswers.njk
```diff
@@ -10,6 +10,7 @@
   {{ govukSummaryList(summaryListArgs) }}
 
   <br>
+    {{ summaryListArgs.appointmentTime }}
```
andrewburrowsmoj, line 13: why are we adding this to the page?

We have to validate what it looks like.

server/views/serviceProviderReferrals/actionPlan/actionPlanEditConfirmation.njk
```diff
@@ -6,6 +6,8 @@
     <div class="govuk-grid-row">
         <div class="govuk-grid-column-two-thirds">
             {{ govukBackLink(backLinkArgs) }}
+            {{ govukNotificationBanner(serviceUserNotificationBannerArgs) }}
```
andrewburrowsmoj, line 9: why are we adding this here when it should be in `scheduleAppointmentCheckAnswers.njk`?

Need to check what is looks like on page.

.run/run local.run.xml
```diff
@@ -0,0 +1,25 @@
+<component name="ProjectRunConfigurationManager">
```
andrewburrowsmoj, line 1: remove this file

tomsmyers (approved):
great work guys - thank you

PR #1095, opened 2021-12-16
Reduce dependency update frequency to weekly

sldblog:

Reduce dependency update frequency to weekly

Since we use many packages and the node ecosystem is lively, we get lots
of dependency updates almost every day

To reduce the amount of overview we have to do, we opted to have it run
every week instead, with the trade off of having more dependency PRs

An alternative to this approach would be to setup auto-merging on
dependency PRs if they pass the test suite, but that's not mutually
exclusive with this solution

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,8 @@ updates:
 - package-ecosystem: npm
   directory: "/"
   schedule:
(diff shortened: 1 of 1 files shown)


jacksonj04 (approved):
LGTM, and the rationale makes sense.

PR #1097, opened 2021-12-16
Persist OAsys risk information on draft referral when user selects 'no - I don't want to edit'

andrewburrowsmoj:
This is a bug fix.

On "make a referral" journey If the user selects that they wish not to edit the OAsys risk information then we need to persist the latest for ARN.

Currently a BAD_REQUEST is thrown when the user accesses the final check-answers page because there is an expectation that draft risk information exists on the back-end.


mahalrmoj (approved):

tomsmyers (approved):
LGTM - thanks andrew

PR #1114, opened 2021-12-29
Trim trailing spaces from integer form inputs

tomsmyers:

allows forms which take integer values to accept inputs with leading or trailing spaces.

this was originally reported for the 'date input' component, but the issue was seen in several validators. each is fixed in a seperate commit. 

allow users to submit forms with leading or trailing spaces.


sldblog (approved):
👍 Looks good to me

PR #1119, opened 2022-01-05
Add form validation for appointment scheduling which validates the appointment time is not before *today*.

tomsmyers:

Add form validation for appointment scheduling which validates the appointment time is not before *today*.

The reason for this is that the API does not currently support appointments before today, and if you
continue to submit the form you end up with an error page with a less useful error message. This
change brings the error to the users attention sooner so it can be corrected without the pain of
confirming and then being presented with an error page.

### DIFF
--- a/server/utils/forms/inputs/twelveHourBritishDateTimeInput.ts
+++ b/server/utils/forms/inputs/twelveHourBritishDateTimeInput.ts
@@ -10,6 +10,7 @@ export interface TwelveHourBritishDateTimeErrorMessages {
   calendarDay: CalendarDayErrorMessages
   clockTime: TwelveHourClockTimeErrorMessages
   invalidTime: string
+  pastTime: string
 }
 
 export default class TwelveHourBritishDateTimeInput {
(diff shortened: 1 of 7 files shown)


server/utils/forms/inputs/twelveHourBritishDateTimeInput.ts
```diff
@@ -57,6 +60,35 @@ export default class TwelveHourBritishDateTimeInput {
       }
     }
 
+    if (this.earliestAllowedTime !== null && result < this.earliestAllowedTime) {
+      const isSameDate =
+        this.earliestAllowedTime.getDate() === result.getDate() &&
+        this.earliestAllowedTime.getMonth() === result.getMonth() &&
+        this.earliestAllowedTime.getFullYear() === result.getFullYear()
```
sldblog, line 67: Based on we don't use any of the hour/minute/second fields on `earliestAllowedTime`, would it be more appropriate to call it `earliestAllowedDate`? (And maybe error if non-zero time portions are given?)

sldblog (approved):
👍 LGTM, just one question, not a blocker

(Pending on tests going green)

server/utils/forms/inputs/twelveHourBritishDateTimeInput.ts
tomsmyers: you can pass hour minute second fields in `earliestAllowedTime`, this is just checking whether or not to show the error in the date field or the time field by checking if the `earliestAllowedTime` is on the same calendar day as the input. see the unit tests for exampes.

sldblog: Ah, I see. The `result < this.earliestAllowedTime` is time-aware. Got it, thanks. I was expecting to see explicit hour/minute checks for some reason.

PR #1124, opened 2022-01-07
Update integration tests for past appointments (earlier today vs yesterday)

tomsmyers:

if you book an SAA for a time in the past, you should immediately be presented with the feedback form, the same way you are for action plan appointments. 

this PR adds int tests and the fix for the new failing SAA int test.

consistency between SAAs and action plan appointments.

### DIFF
--- a/integration_tests/integration/serviceProviderReferrals.spec.js
+++ b/integration_tests/integration/serviceProviderReferrals.spec.js
@@ -1943,6 +1943,52 @@ describe('Service provider referrals dashboard', () => {
       })
 
       describe('with appointments in the past', () => {
+        describe('that happened earlier today', () => {
+          it('schedules the appointment', () => {
+            const today = new Date()
+
+            cy.visit(`/service-provider/referrals/${referral.id}/progress`)
+            cy.get('#supplier-assessment-status').contains('not scheduled')
+            cy.contains('Schedule initial assessment').click()
+
+            // schedule page
+            cy.get('#date-day').type(today.getDate().toString())
+            cy.get('#date-month').type((today.getMonth() + 1).toString())
+            cy.get('#date-year').type(today.getFullYear().toString())
+            cy.get('#time-hour').type('0')
+            cy.get('#time-minute').type('0')
+            cy.get('#time-part-of-day').select('AM')
+            cy.get('#duration-hours').type('1')
+            cy.get('#duration-minutes').type('15')
+            cy.contains('Phone call').click()
+
+            cy.contains('Save and continue').click()
+
+            // // schedule check your answers page
+            cy.get('h1').contains('Confirm appointment details')
+            cy.contains("You've chosen a data and time in the past")
+            cy.contains('Midnight to 1:15am')
+            cy.contains('Phone call')
+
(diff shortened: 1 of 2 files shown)


integration_tests/integration/serviceProviderReferrals.spec.js
```diff
@@ -1094,23 +1094,105 @@ describe('Service provider referrals dashboard', () => {
...
-          cy.get('#duration-hours').type('1')
-          cy.get('#duration-minutes').type('15')
-          cy.contains('1:1').click()
-          cy.contains('Phone call').click()
+            // schedule page
+            const today = new Date()
+            cy.get('#date-day').type(today.getDate().toString())
+            cy.get('#date-month').type((today.getMonth() + 1).toString())
+            cy.get('#date-year').type(today.getFullYear().toString())
+            cy.get('#time-hour').type('0')
```
sldblog, line 1106: Should this be `12` (am) for midnight? I'm confused

integration_tests/integration/serviceProviderReferrals.spec.js
```diff
@@ -1826,300 +1908,344 @@ describe('Service provider referrals dashboard', () => {
...
+
+            // schedule page
+            cy.get('#date-day').type(today.getDate().toString())
+            cy.get('#date-month').type((today.getMonth() + 1).toString())
+            cy.get('#date-year').type(today.getFullYear().toString())
+            cy.get('#time-hour').type('0')
+            cy.get('#time-minute').type('0')
+            cy.get('#time-part-of-day').select('AM')
+            cy.get('#duration-hours').type('1')
+            cy.get('#duration-minutes').type('15')
```
sldblog, line 1962: _not a blocker_ Random observation: this is quite a lot of repetition, does cypress support some kind of encapsulation for "blocks" like this?

integration_tests/integration/serviceProviderReferrals.spec.js
rgforsyth: I just looked on the front end form. You cannot put 12:xx AM. You must put 0:xx AM.
For PM, you have to put 12:XX and not 0:XX though. (That may be obvious, I don't know)

rgforsyth (approved):

integration_tests/integration/serviceProviderReferrals.spec.js
```diff
@@ -1943,60 +1943,6 @@ describe('Service provider referrals dashboard', () => {
       })
 
       describe('with appointments in the past', () => {
-        describe('that happened earlier today', () => {
-          it('presents the feedback form as soon as the appointment is confirmed', () => {
```
tomsmyers: can we change this test to validate the current expected behaviour instead of removing it?

integration_tests/integration/serviceProviderReferrals.spec.js
sldblog: 👍 Thanks Ryan. That sounds like an unconventional user experience and a potential bug in the picker code (12 am is definitely used in the world) -- but it overall doesn't matter anyway. No one books midnight appointments

PR #1132, opened 2022-01-11
Handle delius conflicts on appointments in the past

tomsmyers:

Handle delius conflicts on appointments in the past

stop unhandled errors occuring when you submit confirmation for appointments in the past (with feedback) that cause conflicts on delius).


sldblog (approved):
LGTM

PR #1133, opened 2022-01-11
allow past appointments to record 'did not attend' by sending null in the feedback DTO

tomsmyers:
https://trello.com/c/zvyg5T8r/216-submitting-feedback-for-an-appointment-in-the-past-with-did-not-attend-causes-unhandled-error

allow past appointments to record 'did not attend' by sending null in the feedback DTO.

this kind of a crappy fix that does not address the larger problem of too many overlapping types causing confusion. we need to address this tech debt properly. for now, this adds a little more tech debt, but unblocks the release (hopefully)

allow past appointment feedback forms to be submitted with 'did not attend'


sldblog (approved):
Is this in the contract tests?

I wonder if we have any regressions for this occurrence 

LGTM for a quick fix otherwise

tomsmyers:
> 

good points. 

i put in time tomorrow to talk about this in more detail.

PR #1155, opened 2022-01-12
Revert "Bump helmet from 4.6.0 to 5.0.1" until we review new defaults

sldblog:
Reverts ministryofjustice/hmpps-interventions-ui#1135

helmet 5.0 has breaking configuration changes and I have _no idea_ how it would actually impact production, so I'm reverting out of caution


sldblog:
Thanks!

PR #1159, opened 2022-01-14
Allow SAAs to be scheduled in the past

rgforsyth:

Updates the supplier assessment flow to allow for past appointments. This will then continue to ask for feedback at the same time.

To allow past supplier assessment appointments to be created.

### DIFF
--- a/server/routes/appointments/appointmentsController.ts
+++ b/server/routes/appointments/appointmentsController.ts
@@ -448,7 +448,6 @@ export default class AppointmentsController {
       this.interventionsService.getSupplierAssessment(accessToken, referralId),
     ])
     const appointment = await this.getSupplierAssessmentAppointmentFromDraftOrService(req, res, supplierAssessment)
-    // return because draft service already renders page.
 
     if (appointment === null) {
       throw new Error('Attempting to add supplier assessment attendance feedback without a current appointment')
@@ -513,6 +512,7 @@ export default class AppointmentsController {
         }
 
         res.redirect(`${basePath}/${redirectPath}`)
+        return
       }
     }
 
--- a/server/routes/appointments/appointmentsController.ts
+++ b/server/routes/appointments/appointmentsController.ts
@@ -192,33 +192,33 @@ export default class AppointmentsController {
       res.redirect(
         `/service-provider/referrals/${req.params.id}/supplier-assessment/post-assessment-feedback/edit/${draft.id}/attendance`
       )
-    } else {
-      try {
-        await this.interventionsService.scheduleSupplierAssessmentAppointment(
-          res.locals.user.token.accessToken,
-          supplierAssessment.id,
-          draft.data
+      return
+    }
+    try {
+      await this.interventionsService.scheduleSupplierAssessmentAppointment(
+        res.locals.user.token.accessToken,
+        supplierAssessment.id,
+        draft.data
+      )
+    } catch (e) {
+      const interventionsServiceError = e as InterventionsServiceError
+      if (interventionsServiceError.status === 409) {
+        res.redirect(
+          `/service-provider/referrals/${req.params.id}/supplier-assessment/schedule/${draftBookingId}/details?clash=true`
         )
-      } catch (e) {
-        const interventionsServiceError = e as InterventionsServiceError
-        if (interventionsServiceError.status === 409) {
-          res.redirect(
-            `/service-provider/referrals/${req.params.id}/supplier-assessment/schedule/${draftBookingId}/details?clash=true`
-          )
-          return
-        }
-
-        throw e
+        return
       }
 
-      await this.draftsService.deleteDraft(draft.id, { userId: res.locals.user.userId })
-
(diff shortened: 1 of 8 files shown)


server/routes/appointments/appointmentsController.ts
```diff
@@ -183,32 +184,41 @@ export default class AppointmentsController {
...
-        supplierAssessment.id,
-        draft.data
+    const appointmentScheduleDetails = this.extractAppointmentSchedulingDetails(fetchResult.draft)
+    if (appointmentScheduleDetails === null || !appointmentScheduleDetails.appointmentTime) {
+      throw new Error('No scheduling details were filled in on the draft')
+    }
+    if (this.isAPastAppointment(appointmentScheduleDetails)) {
+      res.redirect(
+        `/service-provider/referrals/${req.params.id}/supplier-assessment/post-assessment-feedback/edit/${draft.id}/attendance`
       )
```
tomsmyers, line 194: can we explicitly return after the redirect, instead of relying on this conditional being the last thing in the method?

server/routes/appointments/appointmentsController.ts
```diff
@@ -431,12 +441,15 @@ export default class AppointmentsController {
...
     const referralId = req.params.id
+    const { draftBookingId } = req.params
 
     const [referral, supplierAssessment] = await Promise.all([
       this.interventionsService.getSentReferral(accessToken, referralId),
       this.interventionsService.getSupplierAssessment(accessToken, referralId),
     ])
-    const appointment = new SupplierAssessmentDecorator(supplierAssessment).currentAppointment
+    const appointment = await this.getSupplierAssessmentAppointmentFromDraftOrService(req, res, supplierAssessment)
+    // return because draft service already renders page.
```
tomsmyers, line 451: this comment doesn't seem to tally with what's actually going on?

server/routes/appointments/appointmentsController.ts
```diff
@@ -451,16 +464,55 @@ export default class AppointmentsController {
...
+            accessToken,
+            referralId,
+            data.paramsForUpdate
+          )
+          basePath = `/service-provider/referrals/${referralId}/supplier-assessment/post-assessment-feedback`
+          redirectPath =
+            updatedAppointment.sessionFeedback?.attendance?.attended === 'no' ? 'check-your-answers' : 'behaviour'
+        }
+
+        res.redirect(`${basePath}/${redirectPath}`)
```
tomsmyers, line 515: i think we a re missing a `return` here

server/routes/appointments/appointmentsController.ts
rgforsyth: I added some returns after seeing your previous bug fix but the linter kept saying they weren't needed and automatically removed them when I added.

server/routes/appointments/appointmentsController.ts
```diff
@@ -183,32 +184,41 @@ export default class AppointmentsController {
     if (draft.data === null) {
       throw new Error('Draft data was unexpectedly null when submitting appointment')
     }
-
-    try {
-      await this.interventionsService.scheduleSupplierAssessmentAppointment(
-        res.locals.user.token.accessToken,
-        supplierAssessment.id,
-        draft.data
+    const appointmentScheduleDetails = this.extractAppointmentSchedulingDetails(fetchResult.draft)
```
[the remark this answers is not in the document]
rgforsyth: yeah. This was how it was done in delivery session so I just kept it the same. All that method was doing is returning the scheduling details or null. I guess it could just be done the other way ad keep it all together.

server/routes/appointments/appointmentsController.ts
rgforsyth: done

server/routes/appointments/appointmentsController.ts
```diff
@@ -451,16 +463,56 @@ export default class AppointmentsController {
...
+                behaviour: {
+                  behaviourDescription: null,
+                  notifyProbationPractitioner: null,
+                },
+                submitted: false,
+                submittedBy: null,
+              }
+            }
+          } else {
+            throw new Error('Draft appointment data is missing.')
```
tomsmyers, line 495: this is a situation we never expect to be in, if it happens, we would probably need a more detailed log message to debug why.

server/routes/appointments/appointmentsController.ts
```diff
@@ -1122,6 +1235,61 @@ export default class AppointmentsController {
...
+    if (appointmentSchedulingDetails) {
+      const draftAppointment = draft.data as DraftAppointment
+      if (draftAppointment?.sessionFeedback) {
+        return {
+          sessionFeedback: draftAppointment.sessionFeedback,
+          id,
+          ...appointmentSchedulingDetails,
+        }
+      }
+      return {
```
tomsmyers, line 1253: i think we could use a comment on why we have to construct this empty object here. (especially because it's never used)

tomsmyers (approved):
nice work - i think we have some refactoring to do to reduce repetition here, but we can do this in stages.

PR #1160, opened 2022-01-14
add missing returns for redirects in appointmentsController

tomsmyers:

add missing returns for redirects in appointmentsController.

stop getting `Cannot set headers after they are sent to the client` errors. 

i think fortunately the missing return doesn't really have any functional issues. the code past the return just attempts to render a page, but it can't because we have already sent the redirect response. this code increases the efficiency of this route significantly though because we do not have to call interventions-api.

### DIFF
--- a/server/routes/appointments/appointmentsController.ts
+++ b/server/routes/appointments/appointmentsController.ts
@@ -382,28 +382,30 @@ export default class AppointmentsController {
     if (appointmentScheduleDetails === null || !appointmentScheduleDetails.appointmentTime) {
       throw new Error('No scheduling details were filled in on the draft')
     }
+
     if (this.isAPastAppointment(appointmentScheduleDetails)) {
       res.redirect(
         `/service-provider/action-plan/${actionPlanId}/appointment/${sessionNumber}/post-session-feedback/edit/${draft.id}/attendance`
       )
-    } else {
-      const success = await this.updateSessionAppointmentAndCheckForConflicts(
-        accessToken,
-        actionPlanId,
-        sessionNumber,
-        appointmentScheduleDetails
-      )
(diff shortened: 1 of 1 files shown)


server/routes/appointments/appointmentsController.ts
```diff
@@ -386,24 +386,28 @@ export default class AppointmentsController {
       res.redirect(
         `/service-provider/action-plan/${actionPlanId}/appointment/${sessionNumber}/post-session-feedback/edit/${draft.id}/attendance`
       )
-    } else {
```
tomsmyers, line 389: wow - this isn't meant to be like this - let me check what happened

server/routes/appointments/appointmentsController.ts
```diff
@@ -382,28 +382,30 @@ export default class AppointmentsController {
...
-      const success = await this.updateSessionAppointmentAndCheckForConflicts(
-        accessToken,
-        actionPlanId,
-        sessionNumber,
-        appointmentScheduleDetails
-      )
+      return
```
tomsmyers, line 390: this diff is confusing due to some alignment changes. the actual change here is the addition of this return statement, and the removal of the subsequent 'else' clause (which is no longer needed after the return)

sldblog: `?w=1` to the rescue 🙌 

image

tomsmyers: cool!

i removed the actual functional change by messing around with this - please reload and see the additional return later in this file.

sldblog (approved):
LGTM!

This feels like it could have a test that no further calls are made if the appointment is in the past

PR #1161, opened 2022-01-14
use user id instead of username to look up service provider users in hmpps auth

tomsmyers:

user user id instead of username to look up service provider users in hmpps auth

depending on how the username is used, email address changes in hmpps auth can cause
pages to become broken entirely in the UI (for example where looking up a referral
assignee with an invalid or old username).

this change removes the old hmpps auth service method for looking up records by username,
and replaces it with the equivalent call to look up users using user id. there are lots
of changes to the integration tests, but it's mainly just find and replace.

fix broken referrals for SP users that have changes their email address / username

### DIFF
--- a/server/routes/appointments/appointmentsController.ts
+++ b/server/routes/appointments/appointmentsController.ts
@@ -878,14 +878,14 @@ export default class AppointmentsController {
     )
   }
 
-  private async getCaseWorker(accessToken: string, username: string): Promise<AuthUserDetails | string> {
+  private async getCaseWorker(accessToken: string, userId: string): Promise<AuthUserDetails | null> {
     try {
-      return this.hmppsAuthService.getSPUserByUsername(accessToken, username)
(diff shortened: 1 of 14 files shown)


server/routes/appointments/appointmentsController.ts
```diff
@@ -878,14 +878,14 @@ export default class AppointmentsController {
     )
   }
 
-  private async getCaseWorker(accessToken: string, username: string): Promise<AuthUserDetails | string> {
+  private async getCaseWorker(accessToken: string, userId: string): Promise<AuthUserDetails | null> {
```
tomsmyers, line 881: i simplified this method, so it doesn't return the username on 404. the caller handles it instead.

sldblog (approved):
Nice

sldblog:
Should we also deal with `getUserDetailsByUsername`?

PR #1181, opened 2022-01-24
add missing return after redirect for saa behaviour feedback

tomsmyers:

add missing return after redirect for saa behaviour feedback

stop `Cannot set headers after they are sent to the client` errors


rgforsyth (approved):
good spot

PR #1184, opened 2022-01-25
add backlink on action plan activity 1 page to take back to progress page

rgforsyth:

Adds a backlink to the create action plan page. (add activity 1). This will take the user back to the progress page for the given referral.

Currently there is no backlink on the add activity 1 page. This change is adding that back in, whilst maintaining the subsequent add activity pages.


sldblog (commented):
It seems like the behaviour changed but no tests regressed -- can we add tests for this which would explain why we need this or that behaviour, please?

rgforsyth:
> It seems like the behaviour changed but no tests regressed -- can we add tests for this which would explain why we need this or that behaviour, please?

We don't normally test the view classes. I have moved the logic into the presenter class (where it probably belongs) and added tests.

PR #1208, opened 2022-01-31
ensure that sessions on the action plan progress page are always pulled from the latest approved action plan

tomsmyers:
https://trello.com/c/xN23pxkV/175-bug-fix-sessions-disppears-when-action-plan-is-in-draft-state

ensure that sessions on the action plan progress page are always pulled from the latest approved action plan

stop sessions going missing when the SP creates a new draft action plan

### DIFF
--- a/server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
+++ b/server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
@@ -53,6 +53,7 @@ import createFormValidationErrorOrRethrow from '../../utils/interventionsFormErr
 import EndOfServiceReportPresenter from '../shared/endOfServiceReport/endOfServiceReportPresenter'
 import EndOfServiceReportView from '../shared/endOfServiceReport/endOfServiceReportView'
 import ServiceProviderSentReferralSummary from '../../models/serviceProviderSentReferralSummary'
+import ActionPlanUtils from '../../utils/actionPlanUtils'
 
 export interface DraftAssignmentData {
   email: string | null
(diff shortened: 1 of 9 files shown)


server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
```diff
@@ -173,42 +174,43 @@ export default class ServiceProviderReferralsController {
...
       interventionPromise,
       actionPlanPromise,
+      this.interventionsService.getApprovedActionPlanSummaries(accessToken, id),
       serviceUserPromise,
-      this.interventionsService.getSupplierAssessment(res.locals.user.token.accessToken, sentReferral.id),
+      this.interventionsService.getSupplierAssessment(accessToken, sentReferral.id),
     ])
 
+    // appointments should always come from the latest approved action plan
+    const latestApprovedActionPlanSummary = approvedActionPlans.sort(sortBy('-approvedAt'))?.[0] ?? null
```
sldblog, line 201: 💭 The usage is always `approvedActionPlans.sort(sortBy('-approvedAt'))?.[0]` -- since the older ones are always unused, do we need the list endpoint? Could we have a singular endpoint instead? (YAGNI)

sldblog:
> stop sessions going missing

How are sessions missing though? Don't we migrate the appointments to the new AP?

server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
tomsmyers: this endpoint is used elsewhere, and i think adding a new endpoint just for this use case seems unnecessary.

tomsmyers:
> rate the appointments to the new AP?

have a look in the ticket for more info. but the sessions dissapear from the UI because the draft AP does not get sessions associated with it until approval (where any additional sessions are also added). happy to give a demo.

also, this change fundamentally makes sense. be it new unapproved APs, or new draft APs, or any unexpected situation, the appointments listed on this page should always come from the latest approved AP.

sldblog (approved):
Thanks -- makes sense 👍

server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
sldblog: ah did not know -- thanks. Checked its usage and it's used as a list. All good 👍

PR #1219, opened 2022-02-02
add 'submittedAt' field to ApprovedActionPlanSummary

tomsmyers:

add 'submittedAt' field to ApprovedActionPlanSummary

allows the UI consume this important piece of information from the API to populate better action plan summary tables.

### DIFF
--- a/server/models/approvedActionPlanSummary.ts
+++ b/server/models/approvedActionPlanSummary.ts
@@ -1,4 +1,5 @@
 export default interface ApprovedActionPlanSummary {
   id: string
   approvedAt: string
+  submittedAt: string
 }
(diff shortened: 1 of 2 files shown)


rgforsyth (approved):

server/models/approvedActionPlanSummary.ts
```diff
@@ -1,4 +1,5 @@
 export default interface ApprovedActionPlanSummary {
   id: string
   approvedAt: string
+  submittedAt: string
```
sldblog, line 4: does this need a contract test change too?

tomsmyers: the contract tests use the factory function which is also changed in this PR. 

it's not super obvious, but this is now encoded in the pacts.

sldblog: ah cool -- thanks

PR #1221, opened 2022-02-03
pass query param to specify the page the case note was located on to redirect back to via backlink

rgforsyth:

Adds a new query param to the case note url so that the backlink can redirect to the page it was selected from.

Currently the backlink on a case note page redirects back to the very first page. This should now redirect to the page the case note was selected from.

### DIFF
--- a/server/routes/caseNotes/caseNotesController.ts
+++ b/server/routes/caseNotes/caseNotesController.ts
@@ -135,7 +135,7 @@ export default class CaseNotesController {
     const { accessToken } = res.locals.user.token
     const { caseNoteId } = req.params
     const backlinkPageNumber = req.query.backlinkPageNumber as string
-    if (!Number(backlinkPageNumber)) {
+    if (backlinkPageNumber && !Number(backlinkPageNumber)) {
       // This is to prevent the url being manipulated.
       throw createError(500, `The page number for a case not backlink should only ever be a number.`, {
         userMessage: 'Incorrect url.',
--- a/server/routes/caseNotes/caseNotesController.ts
+++ b/server/routes/caseNotes/caseNotesController.ts
@@ -134,13 +134,20 @@ export default class CaseNotesController {
   ): Promise<void> {
     const { accessToken } = res.locals.user.token
     const { caseNoteId } = req.params
+    const backlinkPageNumber = req.query.backlinkPageNumber as string
+    if (!Number(backlinkPageNumber)) {
+      // This is to prevent the url being manipulated.
+      throw createError(500, `The page number for a case not backlink should only ever be a number.`, {
+        userMessage: 'Incorrect url.',
+      })
+    }
     const caseNote = await this.interventionsService.getCaseNote(accessToken, caseNoteId)
     const sentByUserDetails = await this.hmppsAuthService.getUserDetailsByUsername(
       accessToken,
       caseNote.sentBy.username
     )
     const sentByUserName = sentByUserDetails.name
-    const presenter = new CaseNotePresenter(caseNote, sentByUserName, loggedInUserType)
+    const presenter = new CaseNotePresenter(caseNote, sentByUserName, loggedInUserType, backlinkPageNumber)
     const view = new CaseNoteView(presenter)
     ControllerUtils.renderWithLayout(res, view, null)
   }
--- a/server/routes/caseNotes/caseNotesController.ts
+++ b/server/routes/caseNotes/caseNotesController.ts
@@ -134,7 +134,10 @@ export default class CaseNotesController {
   ): Promise<void> {
     const { accessToken } = res.locals.user.token
     const { caseNoteId } = req.params
-    const backlinkPageNumber = Number(req.query.backlinkPageNumber) ?? null
+
+    const parsedBacklinkPageNumber = Number(req.query.backlinkPageNumber)
+    const backlinkPageNumber =
+      Number.isNaN(parsedBacklinkPageNumber) || parsedBacklinkPageNumber < 1 ? null : parsedBacklinkPageNumber
 
     const caseNote = await this.interventionsService.getCaseNote(accessToken, caseNoteId)
     const sentByUserDetails = await this.hmppsAuthService.getUserDetailsByUsername(
--- a/server/routes/caseNotes/caseNotesController.ts
+++ b/server/routes/caseNotes/caseNotesController.ts
@@ -134,13 +134,8 @@ export default class CaseNotesController {
   ): Promise<void> {
     const { accessToken } = res.locals.user.token
     const { caseNoteId } = req.params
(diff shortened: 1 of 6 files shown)


server/routes/caseNotes/caseNotesController.ts
```diff
@@ -134,13 +134,20 @@ export default class CaseNotesController {
   ): Promise<void> {
     const { accessToken } = res.locals.user.token
     const { caseNoteId } = req.params
+    const backlinkPageNumber = req.query.backlinkPageNumber as string
```
tomsmyers, line 137: what's the reason for the cast to string? 

if the page number is not included i would expect this to still work (just go back to first page), but i think this would return a 500 ('undefined' isn't a number).

`const { backlinkPageNumber } = req.query` gives you a string or undefined that we can pass through safely

tomsmyers: also i would probably not ever return 500. if it's not a number just ignore it.

tomsmyers: but let the presenter handle that

server/routes/caseNotes/view/caseNotePresenter.ts
```diff
@@ -13,10 +13,13 @@ export default class CaseNotePresenter {
...
     private sentByUserName: string,
-    public loggedInUserType: 'service-provider' | 'probation-practitioner'
+    public loggedInUserType: 'service-provider' | 'probation-practitioner',
+    private backlinkPageNumber: string
   ) {}
 
-  backLinkHref = `/${this.loggedInUserType}/referrals/${this.caseNote.referralId}/case-notes`
+  // There is the chance that further case notes are added which pushes the selected case note onto another page.
+  // The backlink will now take them to the page they came from, but potentially not the page the case note now lives.
+  backLinkHref = `/${this.loggedInUserType}/referrals/${this.caseNote.referralId}/case-notes?page=${this.backlinkPageNumber}`
```
tomsmyers, line 22: see comment above. i would expect this to become 

```
   backLinkHref = `/${this.loggedInUserType}/referrals/${this.caseNote.referralId}/case-notes${(Number(this.backlinkPageNumber) !== NaN) ? `?page=${backlinkPageNumber}` : ''}`
   ```

File: server/routes/caseNotes/caseNotesController.ts (hunk 1 above)
Comment (rgforsyth) [reply server/routes/caseNotes/caseNotesController.ts:137]: I cast as a string as the presenter wouldn't accept it.

Just pulling it straight out of the query params as it is shows the error:

`Argument of type 'string | ParsedQs | string[] | ParsedQs[] | undefined' is not assignable to parameter of type 'string'.   Type 'undefined' is not assignable to type 'string'.`

File: server/routes/caseNotes/viewAll/caseNotesPresenter.ts [hunk 3]
```diff
@@ -65,7 +67,8 @@ export default class CaseNotesPresenter {
       sentByUserType: caseNote.sentBy.authSource === 'delius' ? 'probation practitioner' : 'service provider',
       subject: caseNote.subject,
       body: caseNote.body,
-      caseNoteLink: `/${this.loggedInUserType}/case-note/${caseNote.id}`,
+      // caseNoteLink: `/${this.loggedInUserType}/case-note/${caseNote.id}`,
```
tomsmyers, line 70: nit: remove commented code

server/routes/caseNotes/caseNotesController.ts
tomsmyers: 👍 make the presenter take `string | undefined`. casting to string  turns `undefined` into `'undefined'`.

server/routes/caseNotes/viewAll/caseNotesPresenter.test.ts
```diff
@@ -66,7 +66,7 @@ describe('CaseNotesPresenter', () => {
         'service-provider'
       )
       expect(presenter.tableRows[0].caseNoteLink).toEqual(
-        '/service-provider/case-note/025a807d-a631-4559-be74-664ba62db279'
+        '/service-provider/case-note/025a807d-a631-4559-be74-664ba62db279?backlinkPageNumber=1'
```
sldblog, line 69: Could this be `page=1`?

rgforsyth: Saying page=1 would give the impression that we are on 'page 1' of that screen that isn't paginated. The individual case note screen. 
We want to basically say, if we need to go back, this is the page that you were on in the list of case notes.

sldblog: Right -- individual pages. Gotcha. Thanks!

server/routes/caseNotes/view/caseNotePresenter.ts
```diff
@@ -13,10 +13,15 @@ export default class CaseNotePresenter {
...
     private sentByUserName: string,
-    public loggedInUserType: 'service-provider' | 'probation-practitioner'
+    public loggedInUserType: 'service-provider' | 'probation-practitioner',
+    private backlinkPageNumber: number | null
   ) {}
 
-  backLinkHref = `/${this.loggedInUserType}/referrals/${this.caseNote.referralId}/case-notes`
+  // There is the chance that further case notes are added which pushes the selected case note onto another page.
+  // The backlink will now take them to the page they came from, but potentially not the page the case note now lives.
+  backLinkHref = `/${this.loggedInUserType}/referrals/${this.caseNote.referralId}/case-notes${
```
tomsmyers, line 22: sorry to be picky on this - i think this should explicitly be checking for null instead of trying to cast the type `number | null` to number. it's either null or already a number. if you cast null to number you get NaN.

server/routes/caseNotes/caseNotesController.ts
```diff
@@ -134,13 +134,15 @@ export default class CaseNotesController {
   ): Promise<void> {
     const { accessToken } = res.locals.user.token
     const { caseNoteId } = req.params
+    const backlinkPageNumber = Number(req.query.backlinkPageNumber) ?? null
```
tomsmyers, line 137: another nit: `Number` has some slightly weird behaviour e.g. `Number(null)` is `0`. 

i think the safest this here is to be explicit with covering all our bases. 

```
const parsedBacklinkPageNumber = Number(req.query.backlinkPageNumber) // so this either parses a string, something else, undefined, or null into a number. 
const backlinkPageNumber = parsedBacklinkPageNumber === NaN || parsedBacklinkPageNumber < 1 ? null : parsedBacklinkPageNumber
```

in this case `backlinkPageNumber` is always a positive number or null and the presenter can then handle it appropriately

tomsmyers (approved):

PR #1234, opened 2022-02-07
Upgrade base image to use alpine 3.15

sldblog:

Upgrade base image to use alpine 3.15

Alpine 3.13 ends support on 2022-11-01, 3.15 ends support on 2023-11-01


sldblog:
Thanks!

PR #1237, opened 2022-02-11
add pagination to top of dashboard

rgforsyth:

Adds the pagination nav to the top of the pp dashboard, as well as the bottom.

Makes the dashboard more usable if there are up to 500 rows displayed.

### DIFF
--- a/server/views/probationPractitionerReferrals/dashboard.njk
+++ b/server/views/probationPractitionerReferrals/dashboard.njk
@@ -17,6 +17,8 @@
       <h1 class="govuk-heading-xl">{{ presenter.title }}</h1>
 
         {{ mojSubNavigation(subNavArgs) }}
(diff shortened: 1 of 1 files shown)


server/views/probationPractitionerReferrals/dashboard.njk
```diff
@@ -17,6 +17,8 @@
       <h1 class="govuk-heading-xl">{{ presenter.title }}</h1>
 
         {{ mojSubNavigation(subNavArgs) }}
+        {% include "../partials/pagination.njk" %}
+        <br>
```
rgforsyth, line 21: I feel like this isn't the 'proper' way to add the spacing needed

sldblog: 👋 @rwk-moj could you help please?

PR #1250, opened 2022-02-14
Ignore security rules about PHP, shell and Windows

sldblog:

Ignore security rules about PHP, shell and Windows

As we have no exposure to any component running these, but we saw some
requests being blocked on these rules


tomsmyers (approved):
rad - ignore all the thingss

PR #1252, opened 2022-02-16
Sp referral dashboard paging

rgforsyth:

Adds pagination to the sp dashboard.

Part of the mitigation against poorly performing dashboards.


server/config.ts
```diff
@@ -120,6 +120,12 @@ export default {
           completedCases: get('PP_MY_CASES_PAGE_SIZE', '500'),
           cancelledCases: get('PP_CANCELLED_CASES_PAGE_SIZE', '500'),
         },
+        sp: {
+          myCases: get('SP_CANCELLED_CASES_PAGE_SIZE', '500'),
```
tomsmyers, line 124: env var has wrong name here

server/config.ts
```diff
@@ -120,6 +120,12 @@ export default {
...
+          openCases: get('SP_OPEN_CASES_PAGE_SIZE', '500'),
+          unassignedCases: get('SP_UNASSIGNED_CASES_PAGE_SIZE', '500'),
+          completedCases: get('SP_MY_CASES_PAGE_SIZE', '500'),
```
tomsmyers, line 127: and here

server/routes/serviceProviderReferrals/dashboardPresenter.ts
```diff
@@ -1,18 +1,26 @@
...
+import Intervention from '../../models/intervention'
+import Pagination from '../../utils/pagination/pagination'
 
 export type DashboardType = 'My cases' | 'All open cases' | 'Unassigned cases' | 'Completed cases'
 export default class DashboardPresenter {
+  public readonly pagination: Pagination
+
   constructor(
-    private readonly referralsSummary: ServiceProviderSentReferralSummary[],
+    private readonly referralsSummary: Page<SentReferral>,
```
tomsmyers, line 17: nit: this is long longer 'summary' - these are just referrals

server/routes/serviceProviderReferrals/dashboardPresenter.ts
```diff
@@ -51,26 +67,28 @@ export default class DashboardPresenter {
...
+          referralSummary.referral.serviceUser.firstName ?? ''
         }`.toLocaleLowerCase('en-GB'),
         href: null,
       },
-      { text: referralSummary.interventionTitle, sortValue: null, href: null },
+      { text: interventionForReferral.title, sortValue: null, href: null },
       this.showAssignedCaseworkerColumn
         ? null
-        : { text: referralSummary.assignedToUserName ?? '', sortValue: null, href: null },
+        : { text: referralSummary.assignedTo?.username ?? '', sortValue: null, href: null },
```
tomsmyers, line 82: unrelated to these changes, but in the PP dashboard i think we show 'Unassigned' for these types of cases - do we know why this is different?

server/routes/serviceProviderReferrals/dashboardPresenter.ts
```diff
@@ -51,26 +67,28 @@ export default class DashboardPresenter {
...
       { text: 'View', sortValue: null, href: DashboardPresenter.hrefForViewing(referralSummary) },
     ].filter(row => row !== null) as SortableTableRow
   })
 
-  private static hrefForViewing(referralSummary: ServiceProviderSentReferralSummary): string {
```
tomsmyers, line 69: same hit here, this is no longer a referral summary

server/routes/serviceProviderReferrals/dashboardPresenter.ts
```diff
@@ -51,26 +67,28 @@ export default class DashboardPresenter {
...
-    if (referralSummary.assignedToUserName === null || referralSummary.assignedToUserName === undefined) {
-      return `/service-provider/referrals/${referralSummary.referralId}/details`
+  private static hrefForViewing(referralSummary: SentReferral): string {
+    if (referralSummary.assignedTo?.username === null || referralSummary.assignedTo?.username === undefined) {
```
tomsmyers, line 88: you can shortcut this with `== null` - it's up to you if you prefer it or not

server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
```diff
@@ -76,43 +79,51 @@ export default class ServiceProviderReferralsController {
...
+      sort: ['sentAt,DESC'],
+    }
+
+    const cases = await this.interventionsService.getSentReferralsForUserTokenPaged(
+      res.locals.user.token.accessToken,
+      getSentReferralsFilterParams,
+      paginationQuery
+    )
+
+    const dedupedInterventionIds = Array.from(new Set(cases.content.map(referral => referral.referral.interventionId)))
```
tomsmyers, line 121: not a blocker: i could be good to abstract this logic from here and the pp dashboard and put a unit test around it

server/views/serviceProviderReferrals/dashboard.njk
```diff
@@ -14,10 +14,13 @@
       <h1 class="govuk-heading-xl">{{ presenter.title }}</h1>
 
         {{ mojSubNavigation(subNavArgs) }}
+        {% include "../partials/pagination.njk" %}
```
tomsmyers, line 17: imports at the top of the file please

server/routes/serviceProviderReferrals/dashboardPresenter.test.ts
```diff
@@ -22,93 +70,96 @@ describe(DashboardPresenter, () => {
...
-              serviceUserFirstName: 'George',
-              serviceUserLastName: 'Michael',
-            },
-          ]
-          const presenter = new DashboardPresenter(referralsSummary, dashboardType, loggedInUser)
+          const page = pageFactory.pageContent(referrals).build() as Page<SentReferral>
+          const presenter = new DashboardPresenter(page, dashboardType, loggedInUser, interventions)
 
-          expect(presenter.tableRows[0][4]).toMatchObject({ text: 'john.smith' })
+          expect(presenter.tableRows[0][4]).toMatchObject({ text: 'UserABC' })
```
sldblog, line 130: This condition feels a bit magic-y; the previous version explicitly declared `assignedToUserName` but the new data setup does not -- this might condition might pass on all rows

server/routes/serviceProviderReferrals/dashboardPresenter.ts
```diff
@@ -40,8 +48,16 @@ export default class DashboardPresenter {
 
   readonly navItemsPresenter = new PrimaryNavBarPresenter('Referrals', this.loggedInUser)
 
-  readonly tableRows: SortableTableRow[] = this.referralsSummary.map(referralSummary => {
+  readonly tableRows: SortableTableRow[] = this.sentReferrals.content.map(referralSummary => {
     const sentAtDay = CalendarDay.britishDayForDate(new Date(referralSummary.sentAt))
+    const interventionForReferral = this.interventions.find(
```
sldblog, line 53: _not a blocker_; _performance_
💭 `find` here might do a few iterations until it finds the correct element; if we intend the `interventions` object to be a cache for this lookup, we could initialise it for this access, i.e. `{id1: Intervention1, id2: Intervention2}`

just flagging because UI is now leading in CPU consumption (blue is UI):
image

server/routes/serviceProviderReferrals/dashboardPresenter.ts
```diff
@@ -40,8 +48,16 @@ export default class DashboardPresenter {
...
+      intervention => intervention.id === referralSummary.referral.interventionId
+    )
+    if (interventionForReferral === undefined) {
+      throw new Error(
+        `Expected referral to be linked to an intervention with ID ${referralSummary.referral.interventionId}`
```
sldblog, line 58: I think if this occurs, we would benefit from the `referral.id` as well, could we include that, please?

server/routes/serviceProviderReferrals/dashboardPresenter.ts
sldblog: I do not have an answer; but we could push this into config to make it look like an explicit choice -- `dashboards.serviceProvider.unassignedLabel` 🤷

server/config.ts
```diff
@@ -120,6 +120,12 @@ export default {
           completedCases: get('PP_MY_CASES_PAGE_SIZE', '500'),
           cancelledCases: get('PP_CANCELLED_CASES_PAGE_SIZE', '500'),
         },
+        sp: {
```
sldblog, line 123: _not a blocker_
💭 `config.apis.interventionsService.dashboardPageSize.sp.openCases`

I think I see why this is under API config, however, it can also be considered as UI behaviour -- so it might be worth simplifying this to `config.dashboards.serviceProvider.openCases` or similar

server/routes/serviceProviderReferrals/serviceProviderReferralsController.test.ts
```diff
@@ -87,102 +87,195 @@ afterEach(() => {
...
+    interventionsService.getIntervention.mockResolvedValue(intervention)
+    interventionsService.getSentReferralsForUserTokenPaged.mockResolvedValue(page)
     await request(app)
       .get('/service-provider/dashboard/all-open-cases')
       .expect(200)
+      .expect(res => {
+        expect(res.text).toContain('My cases')
+        expect(res.text).toContain('Alex River')
+        expect(res.text).toContain('Accommodation Services - West Midlands')
+        expect(res.text).toContain('')
```
sldblog, line 180: `expect(res.text).toContain('')` will always be true -- do we need it?

sldblog (commented):
Looks good -- couple of minor comments about testing and performance

server/config.ts
tomsmyers: agree with this

server/config.ts
```diff
@@ -125,6 +126,12 @@ export default {
           completedCases: Number(get('PP_MY_CASES_PAGE_SIZE', '500')),
           cancelledCases: Number(get('PP_CANCELLED_CASES_PAGE_SIZE', '500')),
         },
+        sp: {
+          myCases: get('SP_MY_CASES_PAGE_SIZE', '500'),
```
tomsmyers, line 130: there is an unintended change here that the vars are no longer converted to `Number`

server/routes/deprecated/dashboardWithoutPaginationPresenter.ts
```diff
@@ -0,0 +1,76 @@
+import CalendarDay from '../../utils/calendarDay'
```
tomsmyers, line 1: guessing this is just a rename? wonder why the diff doesn't show it as such ...

server/routes/serviceProviderReferrals/dashboardPresenter.test.ts
tomsmyers: i think a counter condition - that an unassigned referral does not show the name on a dashboard like 'all cases' - would be sufficient to validate the test case.

server/routes/deprecated/dashboardWithoutPaginationPresenter.ts
rgforsyth: This was a new fill added. The previous one was amended to include the pagination stuff

server/config.ts
rgforsyth: Done

sldblog (approved):
👍 Let's test it on dev

PR #1268, opened 2022-02-21
Sp dashboard pagination old flow

rgforsyth:

(This pr also contains the work from the previous pr  #1252 which introduces a paginated sp dashboard.)
Adds the old (non-paginated) sp dashboard flow back into the codebase, currently toggled by a flag. This will later be determined by some extra logic.

To control the rollout of the sp paginated dashboard, we will be using both the paginated and non-paginated dashboards temporarily.


sldblog:
Could the base of this PR change to #1252 so we can see the changes simpler, please?

This can be achieved with the "Edit" button:

image

PR #1270, opened 2022-02-23
Sp pagination with non pagination

rgforsyth:

This pr follows on from the previous pr https://github.com/ministryofjustice/hmpps-interventions-ui/pull/1252 which introduces a paginated sp dashboard.
Adds the old (non-paginated) sp dashboard flow back into the codebase, currently toggled by a flag. This will later be determined by some extra logic.

To control the rollout of the sp paginated dashboard, we will be using both the paginated and non-paginated dashboards temporarily.


server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
```diff
@@ -67,6 +71,8 @@ export type DraftAppointmentBooking = null | AppointmentSchedulingDetails
 export default class ServiceProviderReferralsController {
   private readonly deliusOfficeLocationFilter: DeliusOfficeLocationFilter
 
+  private paginationFlow = true
```
tomsmyers, line 74: let's ship this first with the old non-paged version and turn it on in a later commit

PR #1271, opened 2022-02-23
add method to compute a repeatable boolean value to enable features for percentages of users

tomsmyers:

add method to compute a repeatable boolean value to enable features for percentages of users

allow us to enable a feature to say just 10% of the user base in a deterministic fashion.


server/services/featureFlagService.test.ts
```diff
@@ -0,0 +1,57 @@
...
+      })
+    })
+
+    it('always returns true for 100 percent', () => {
+      makeRandomUserIds().forEach(it => {
+        expect(FeatureFlagService.enableForUser(it, 100)).toEqual(true)
+      })
+    })
+
+    retry('returns roughly the right number', 10, () => {
```
rgforsyth, line 31: Not sure there is a better way to test this, but is there still a chance these could fail? I see you've given a bit of leeway.

rgforsyth (approved):
Nice one. Question about the likelihood of the tests failing, but I guess we can monitor that.

server/services/featureFlagService.test.ts
tomsmyers: updated the test so that it users a real sample of user ids and validates a single number.

PR #1286, opened 2022-03-02
fix sorting of ROSH panel by converting input groups to lowercase before comparison with fixed order list

tomsmyers:

fix sorting of ROSH panel by converting input groups to lowercase before comparison with fixed order list

fix ordering of rosh panel box in prod environments where the groups are capitalized.


server/routes/shared/roshPanelPresenter.ts
```diff
@@ -104,7 +104,7 @@ export default class RoshPanelPresenter {
     roshGroupsDisplayOrder.forEach(group => {
       // We get the index here and then push/splice using that index value to save having to `find` on the array twice
       const matchingRoshGroupIndex = unsortedRiskGroups.findIndex(obj => {
-        return obj.riskTo === group
+        return obj.riskTo.toLowerCase() === group
```
sldblog, line 107: Should we lowercase both, for consistency?

sldblog (approved):

PR #1287, opened 2022-03-04
Limit sp dahsboard pagination

rgforsyth:

Adds a new config variable to determine what percentage of users are shown the sp dashboard pagination.

To allow a controlled roll out of the sp dashboard pgination


server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
```diff
@@ -84,7 +85,7 @@ export default class ServiceProviderReferralsController {
   }
 
   async showMyCasesDashboard(req: Request, res: Response): Promise<void> {
-    if (!config.features.spDashboardPaginationFlow) {
+    if (!FeatureFlagService.enableForUser(res.locals.user, config.dashboards.serviceProvider.filteredUserPercentage)) {
```
tomsmyers, line 88: i wonder if this would be easier to read if the order of the branches were swapped and we can remove the `!`?

tomsmyers (approved):
couple of nits, but in general LGTM!

PR #1299, opened 2022-03-07
add else clause to sp pagination branching to fix header issue

rgforsyth:

Adds missing else clause to sp pagination branching.

Fix `Cannot set headers after they are sent to the client` error


sldblog (approved):
Thanks @rgforsyth! 👍 

Is there any way this could be captured by tests?

tomsmyers:
there have been a number of these types of issues (i think 3) in the past couple months. i wonder if there's some way for us to lint for 'no return after render/redirect'?

sldblog:
could we add a default 'to-be-carried-over' header to our mocks?
and/or with mocks we can usually say something like 'exactly one interaction' or similar?

PR #1302, opened 2022-03-10
add App Insights custom event to help measure performance of paged dashboards

tomsmyers:

add App Insights custom event to track some measurements around paged SP dashboards

help measure performance of paged dashboards


rgforsyth (approved):
👍

PR #1325, opened 2022-03-22
never pass 'undefined' to http-error createError function.

tomsmyers:

never pass 'undefined' to http-error createError function.

some errors thrown by 'RestClient' do not have a http status code,
for example timeouts or connection errors. in these cases,
pass a status of 500 instead. this doesn't really have any effect
on the application's behaviour, because 'external' errors always
end up with a 500 response anyway. this is a sensible default either
way because a 500 is the appropriate response code for an unhandled
internal error.

this also stops us getting alerts in sentry that look like
'argument #1 unsupported type undefined'.

### DIFF
--- a/server/services/assessRisksAndNeedsService.ts
+++ b/server/services/assessRisksAndNeedsService.ts
@@ -36,7 +36,7 @@ export default class AssessRisksAndNeedsService {
         return null
       }
 
-      throw createError(restClientError.status, restClientError, {
+      throw createError(restClientError.status || 500, restClientError, {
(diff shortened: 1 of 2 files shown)


rgforsyth (approved):

kth13 (approved):
Looks good

sldblog (approved):
504 (Gateway Timeout) might be appropriate to immediately identify "external" causes

> The server, while acting as a gateway or proxy, did not receive a timely response from an upstream server it needed to access in order to complete the request

PR #1342, opened 2022-03-30
Improve performance of dashboard lookup queries 2

kth13:

- configuring the new API endpoint sentreferrals/summaries which returns only the properties that is required to load the dashboard
- defining the model that supports the new api endpoint
- integrate the dashboards to use the new model.
- fixing the unit tests and integration tests

- The landing dashboard page was taking more time to load when there is more data. This changes are the result of the API changes to fix the performance issue


server/models/sentReferralDashboard.ts
```diff
@@ -0,0 +1,16 @@
+import User from './hmppsAuth/user'
+import ServiceProvider from './serviceProvider'
+import ServiceUser from './serviceUser'
+
+export default interface SentReferralDashboard {
```
tomsmyers, line 5: as we discussed in the backend PR, we are going to rename this class.

server/routes/probationPractitionerReferrals/dashboardPresenter.ts
```diff
@@ -89,13 +87,6 @@ export default class DashboardPresenter {
   }
 
   readonly tableRows: SortableTableRow[] = this.sentReferrals.content.map(referral => {
-    const interventionForReferral = this.interventions.find(
```
tomsmyers, line 92: it's nice that we can remove this code with the DTOs

server/routes/serviceProviderReferrals/dashboardPresenter.ts
```diff
@@ -84,16 +82,16 @@ export default class DashboardPresenter {
...
     const sentAtDay = CalendarDay.britishDayForDate(new Date(referralSummary.sentAt))
-    const interventionForReferral = this.interventions.find(
-      intervention => intervention.id === referralSummary.referral.interventionId
-    )
-    if (interventionForReferral === undefined) {
-      throw new Error(
-        `Expected referral ${referralSummary.id} to be linked to an intervention with ID ${referralSummary.referral.interventionId}`
-      )
-    }
+    // const interventionForReferral = this.interventions.find(
```
tomsmyers, line 87: is this commented code supposed to be here?

testutils/factories/sentReferral.ts
```diff
@@ -11,6 +11,7 @@ const exampleReferralFields = () => {
     createdAt: '2020-12-07T20:45:21.986389Z',
     completionDeadline: '2021-04-01',
     serviceProvider: {
+      id: '1234',
```
tomsmyers, line 14: aha - i imagine this factory class is used in the contract tests, so this addition may be all that's required to validate it. let's double check that though.

tomsmyers (commented):
all looking good - no issues with these changes.

thanks Kumar!

server/models/sentReferralDashboard.ts
kth13: I have renamed this class and all the reference to this class too

server/models/serviceProvider.ts
```diff
@@ -1,3 +1,4 @@
 export default interface ServiceProvider {
+  id: string
```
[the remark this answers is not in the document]
kth13: As you suggested,  I have updated the pactTest regarding this change

PR #1358, opened 2022-04-04
Amend completion deadline

rgforsyth:

Allows pp's to update the completion deadline on a referral.

To reduce the amount of re-referrals by giving the option to update the details.


server/routes/makeAReferral/completion-deadline/completionDeadlineForm.test.ts
```diff
@@ -39,4 +42,97 @@ describe('CompletionDeadlineForm', () => {
...
+        errors: [
+          {
+            errorSummaryLinkedField: 'reason-for-change',
+            formFields: ['reason-for-change'],
+            message: 'Reason for change cannot be empty',
+          },
+        ],
+      })
+    })
+    it('returns an error if completion date is valid but reason for change is invalid', async () => {
```
kth13, line 94: I believe the the test name is completion date is invalid and reason of change is valid?

server/models/referralUpdate.ts
```diff
@@ -0,0 +1,6 @@
+import DraftReferral from './draftReferral'
+
+export interface ReferralDetailsFormUpdate {
```
tomsmyers, line 3: should this live in the models directory or alongside the form?

server/routes/makeAReferral/completion-deadline/completionDeadlineForm.ts
```diff
@@ -1,20 +1,86 @@
...
-  constructor(private readonly request: Request) {}
+  constructor(private readonly request: Request, private readonly isSentReferral: boolean) {}
 
-  async data(): Promise<FormData<Partial<DraftReferral>>> {
+  static readonly completionDeadlineFieldId = 'completion-deadline'
+
+  static readonly reasonForChangeFieldId = 'reason-for-change'
+
+  async data(): Promise<FormData<ReferralDetailsFormUpdate>> {
     const input = new CalendarDayInput(this.request, 'completion-deadline', errorMessages.completionDeadline)
```
tomsmyers, line 19: nit on naming (not a blocker) - would be good to label this var as the date input

server/routes/makeAReferral/completion-deadline/completionDeadlineForm.ts
```diff
@@ -1,20 +1,86 @@
...
-    const result = await input.validate()
+    const completionDeadlineResult = await input.validate()
+    let reasonForChangeResult: FormValidationResult<string> | null = null
```
tomsmyers, line 21: is this var used outside the 'if' block below?

server/routes/makeAReferral/completion-deadline/completionDeadlineForm.ts
```diff
@@ -1,20 +1,86 @@
...
+      : { error: completionDeadlineResult.error, paramsForUpdate: null }
+  }
 
-    if (result.value) {
-      return { error: null, paramsForUpdate: { completionDeadline: result.value.iso8601 } }
+  static get validations() {
+    return [
+      ExpressValidator.body(CompletionDeadlineForm.reasonForChangeFieldId)
+        .notEmpty({ ignore_whitespace: true })
+        .withMessage(errorMessages.completionDeadline.cannotBeEmpty),
```
tomsmyers, line 59: this variable name suggests that the message is for the scenario when the completion deadline field is empty - it is actually for then the reason for change field is empty (i think?)

server/routes/makeAReferral/completion-deadline/completionDeadlineForm.ts
```diff
@@ -1,20 +1,86 @@
...
+      request,
+      validations: CompletionDeadlineForm.validations,
+    })
+
+    const error = CompletionDeadlineForm.extractErrors(validationResult)
+    if (error) {
+      return { value: null, error }
     }
+    return {
+      value: CompletionDeadlineForm.reasonForChangeFieldId,
```
tomsmyers, line 74: slightly confused about what this return value represents. should the value be the actual user-entered value in the form? how is this return value being used?

server/routes/makeAReferral/completion-deadline/completionDeadlinePresenter.ts
```diff
@@ -15,8 +15,9 @@ export default class CompletionDeadlinePresenter {
   readonly hint = 'For example, 27 10 2021'
 
   constructor(
-    private readonly referral: DraftReferral,
+    private readonly completionDeadline: string | null,
     private readonly intervention: Intervention,
+    readonly sentReferral: boolean | undefined = undefined,
```
tomsmyers, line 20: what is this used for? i see it passed in the render args in the view, but not used anywhere.

server/routes/makeAReferral/completion-deadline/completionDeadlineView.ts
```diff
@@ -46,6 +46,23 @@ export default class CompletionDeadlineView {
...
   }
 
+  private get textAreaArgs(): TextareaArgs {
+    return {
+      name: 'reason-for-change',
+      id: 'reason-for-change',
+      label: {
+        text: `What is the reason for changing the completion date?`,
+        classes: 'govuk-label--l',
+        isPageHeading: true,
```
tomsmyers, line 56: is this a page heading?

server/routes/makeAReferral/makeAReferralController.ts
```diff
@@ -797,4 +849,23 @@ export default class MakeAReferralController {
...
+    const caseNote = {
+      referralId,
+      subject,
+      body: message,
+    }
+    await this.interventionsService.addCaseNotes(accessToken, caseNote)
+  }
+
+  async isSentReferral(req: Request, res: Response): Promise<boolean> {
+    try {
```
tomsmyers, line 863: we need to explicitly check for 404 here on the draft referral call. 

also - if the second call fails or also returns 404, what do we do?

tomsmyers: i would guess we probably fail in some way. do we need to fail now? can we assume the referral is sent and have code further down the chain handle the fact that something is wrong? that would save an additonal API call.

server/routes/probationPractitionerReferrals/probationPractitionerReferralsController.ts
```diff
@@ -208,6 +208,8 @@ export default class ProbationPractitionerReferralsController {
   async showReferral(req: Request, res: Response): Promise<void> {
     const { accessToken } = res.locals.user.token
     const sentReferral = await this.interventionsService.getSentReferral(accessToken, req.params.id)
+    const successfulChange = req.query.success
```
tomsmyers, line 211: do we need two separate vars to compyte this boolean?

server/routes/shared/showReferralView.ts
```diff
@@ -61,6 +61,27 @@ export default class ShowReferralView {
     href: this.presenter.referralOverviewPagePresenter.dashboardURL,
   }
 
+  private readonly insetTextArgs = {
+    html: `${ViewUtils.escape(`You can amend the completion date on this referral.`)}<br>
+          ${ViewUtils.escape(`This will send a notification to the service provider`)}`,
+  }
+
+  get notificationBannerArgs(): NotificationBannerArgs {
+    const html = `<b>Referral changes saved</b>
```
tomsmyers, line 70: if this is to style the text in **bold**, the <strong> tag should be used instead.

server/routes/shared/showReferralView.ts
```diff
@@ -61,6 +61,27 @@ export default class ShowReferralView {
...
+          <br>
```
tomsmyers, line 71: prefer govuk horizontal space components to line breaks

server/routes/shared/showReferralView.ts
```diff
@@ -61,6 +61,27 @@ export default class ShowReferralView {
...
+          <p>
+             The service provider has been notified of the changes to this referral.
+          </p>
+          <p>
+              <a href= ${this.presenter.closeHref}> close</a>
```
tomsmyers, line 76: nit: should this be `Close` (with no leading space)?

tomsmyers: also i notice in the designs this is `Dismiss`

server/views/makeAReferral/completionDeadline.njk
```diff
@@ -19,6 +20,10 @@
 
         {{ govukDateInput(dateInputArgs) }}
 
+        {% if sentReferral %}
```
tomsmyers, line 23: you can do without the additional logic for sent referral in this template and just render the input if it's present 

so `{% if textAreaArgs %}... `

server/views/shared/referralDetails.njk
```diff
@@ -2,11 +2,18 @@
...
 {% from "govuk/components/details/macro.njk" import govukDetails %}
+{% from "govuk/components/inset-text/macro.njk" import govukInsetText %}
+{% from "govuk/components/notification-banner/macro.njk" import govukNotificationBanner %}
+
 
 {% set pageTitle = "Referral" %}
 {% set pageSubTitle = "Details" %}
 
 {% block referralPageSection %}
+      {% if presenter.userType == 'probation-practitioner' %}
```
tomsmyers, line 13: same as above here too

tomsmyers: the general concept is leave the logic to the presenter / view classes, these templates should be kept pretty dumb IMO, just render stuff if it's there.

tomsmyers (commented):
a few bits and pieces - nothing super major - nice work!

tomsmyers:
note on change size: this would be easier to review and update if it were split into smaller chunks behind feature flags. e.g. the changes to show the links to update on the referral details page would be a nice logically seperate change. the form changes could also be done seperate and shipped to production (without an url to access the new form, for example). etc. 

in general if something is taking more than a few days, i think we need to take a look at how we can break it down.

server/models/referralUpdate.ts
rgforsyth: In the future would we not use this for other referral details forms(enforceable days)? All of these would live in slightly different places.

server/routes/makeAReferral/completion-deadline/completionDeadlineForm.ts
rgforsyth: This was how one of the other validate methods was doing it. It is returning the result from the validationErrorFromResult into a FormValidationResult, which all the fields will be after their own validations.

server/routes/makeAReferral/completion-deadline/completionDeadlinePresenter.ts
rgforsyth: Was going to be used to determine if to show the backLink for a sent referral. We don't have the backlink for the make a referral(draft) referral flow.

server/routes/shared/showReferralView.ts
rgforsyth: The design i'm looking at says 'close'

server/routes/makeAReferral/completion-deadline/completionDeadlinePresenter.ts
tomsmyers: if we are wanting to show a backlink i would just recommend passing in a backlinkArgs object or similar. see later comments on decoupling logic from templates.

tomsmyers (approved):
LGTM - nice work. let's get it tested fully in dev.

PR #1374, opened 2022-04-12
add changelink into pp referral details page for updating enforceable days

tomsmyers:

changed the query param used to gate displaying the success banner on referral details page from `success` to `detailsUpdated`

add changelink into pp referral details page for updating enforceable days

reduce rates of re-referrals

### DIFF
--- a/server/routes/probationPractitionerReferrals/probationPractitionerReferralsController.ts
+++ b/server/routes/probationPractitionerReferrals/probationPractitionerReferralsController.ts
@@ -208,7 +208,6 @@ export default class ProbationPractitionerReferralsController {
   async showReferral(req: Request, res: Response): Promise<void> {
     const { accessToken } = res.locals.user.token
     const sentReferral = await this.interventionsService.getSentReferral(accessToken, req.params.id)
-    const showSuccess = !!req.query.success
 
     const { crn } = sentReferral.referral.serviceUser
     const [intervention, sentBy, expandedServiceUser, conviction, riskInformation, riskSummary, responsibleOfficer] =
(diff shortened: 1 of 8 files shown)


sldblog:
I'm a bit confused trying to follow this -- are there two changes here?

- adding the change link in `server/routes/shared/showReferralPresenter.ts`
- changing links to only display a banner if the details have been updated

Am I missing something?

server/routes/probationPractitionerReferrals/probationPractitionerReferralsController.ts
```diff
@@ -208,7 +208,6 @@ export default class ProbationPractitionerReferralsController {
   async showReferral(req: Request, res: Response): Promise<void> {
     const { accessToken } = res.locals.user.token
     const sentReferral = await this.interventionsService.getSentReferral(accessToken, req.params.id)
-    const showSuccess = !!req.query.success
```
NeilLuptonMoJ, line 211: would it not be a bit easier to read to keep this `showSuccess` as
```
    const showSuccess = req.query.detailsUpdated === 'true'
```
just shows the reason for this better than determining the boolean as its being passed in to `ShowReferralPresenter`

tomsmyers: fair point. i figured readers of this method don't care why we're pulling that query param out though, since there is no logic that depends on it here. happy to move it into a variable though.

NeilLuptonMoJ: yeah, just a nitpick really, so doesn't really matter, happy for it to stay as is

NeilLuptonMoJ (approved):

PR #1377, opened 2022-04-13
🚸 Retry reads and writes on transient errors

sldblog:

Retry reads and writes to external services on transient errors to improve user experience.

Currently, users experience random network blips as fatal errors (for example, 502 Bad Gateway responses) and we should retry those transient failures.

This change will retry only those. Quoting https://visionmedia.github.io/superagent/#retrying-requests:

> By default the following status codes are retried:
>
>     408
>     413
>     429
>     500
>     502
>     503
>     504
>     521
>     522
>     524
>
> By default the following error codes are retried:
>
>     'ETIMEDOUT'
>     'ECONNRESET'
>     'EADDRINUSE'
>     'ECONNREFUSED'
>     'EPIPE'
>     'ENOTFOUND'
>     'ENETUNREACH'
>     'EAI_AGAIN'

which are good enough defaults for us.

This was previously turned off in b24a3d2ca21fcf5d0e679d13d75cc686e14d9737 to mitigate load-related snowballing.

Further improvements could focus on introducing [circuit breakers](https://microservicesdev.com/circuit-breaker-pattern).

### DIFF
--- a/server/data/restClient.ts
+++ b/server/data/restClient.ts
@@ -29,6 +29,9 @@ interface PostRequest {
 
 export type RestClientError = HttpError
 
+const readRetryCount = 1
+const writeRetryCount = 1
+
 export default class RestClient {
   agent: Agent
 
@@ -83,7 +86,7 @@ export default class RestClient {
       const unauthenticatedRequest = superagent
         .get(`${this.apiUrl()}${path}`)
         .agent(this.agent)
-        .retry(0)
+        .retry(readRetryCount)
         .query(query)
         .set(headers)
         .responseType(responseType)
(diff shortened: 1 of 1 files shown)


server/data/restClient.ts
```diff
@@ -29,6 +29,9 @@ interface PostRequest {
 
 export type RestClientError = HttpError
 
+const readRetryCount = 1
+const writeRetryCount = 1
```
sldblog, line 33: `1` _is_ the default for superagent, according to its docs -- so we could just remove the lines, but I kinda feel it's worth being explicit

NeilLuptonMoJ (approved):

sldblog:
thanks!

PR #1384, opened 2022-04-19
Update node-redis to latest version with backwards incompatible API changes.

tomsmyers:

Update node-redis to latest version with backwards incompatible API changes.

By default the new client uses a new native async API. 'connect-redis' requires
the legacy client version to function, so we have a few places where we have
to use the 'v4' API access on the legacy-mode client. The client used in the
HMPPS Auth service is isolated so doesn't need to do this.

get the latest version of the redis library (and clear out a lingering dependabot PR)


sldblog:
Looks good, but tests are failing at the moment

PR #1387, opened 2022-04-20
Change ordering of draft referrals from ascending to descending

JakeGoodwin22:

Changes the ordering of draft referrals from ascending to descending so that we can see the latest draft at the top of the list.

Users are more likely to want to view and make changes to the most recent draft that they have created, so we show them the most recent at the top of the list to make it easier for them. This is also part of a greater effort to make draft referrals more useful to PP's.


kth13 (approved):
Looks good to me

PR #1388, opened 2022-04-22
Updated readme to include setup for those using Apple silicon laptops

JakeGoodwin22:

Included set up guide for those running the latest apple chips as the current setup doesn't work.

Make it easier for new starters to get set up by following a guide that is designed for their machine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,16 @@ docker-compose pull
 npm install
 ```
 
+### Initial setup part 2 (If using Apple M1 chipset)
+
+```
+Pull down following repositories:
+hmpps-assess-risks-and-needs, community-api, hmpps-auth, and offender-assessments-api
+
+Go into each folder of downloaded repo's and run following command in each:
+docker buildx build --platform linux/arm64 --load --tag quay.io/hmpps/REPLACE-ME-WITH-REPO-NAME:latest .
+```
+
 You'll also need to add the following line to your `/etc/hosts` file:
 
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -31,6 +31,27 @@ bin/start-dependencies
...
+```
+docker-compose pull
+
+(nvm use)
+npm install
+
+Pull down following repositories:
+hmpps-assess-risks-and-needs, community-api, hmpps-auth, and offender-assessments-api
+
+Add '127.0.0.1 hmpps-auth' to your /etc/hosts file
```
sldblog, line 45: This is already mentioned above, I wonder if it's worth adding 'M1' sections in-between the 'general' sections? (Duplication is harder to maintain)

sldblog (commented):
Thanks for doing this!
Comment about duplication

README.md
JakeGoodwin22: Good point! I've changed the layout to make it clearer and to remove duplication

README.md
```diff
@@ -27,10 +37,13 @@ You'll also need to add the following line to your `/etc/hosts` file:
 ### Running the app
 
 ```
-bin/start-dependencies
+If not using Apple chipset use: 'bin/start-dependencies'
```
sldblog, line 40: _not a blocker_
Is this because of the dockerize image to check for "is it up yet"? Otherwise it's using compose as well

sldblog (approved):
Looks good to me, no blockers.

@kth13, IIRC you have an M1 laptop too, could you review as well, please?

PR #1389, opened 2022-04-22
Updated Interventions details page to remove unused tab section

JakeGoodwin22:

Removes tab from the bottom of the Interventions details page as it is not used.

The tab at the bottom of the interventions details page was going to be used to include other information than the service provider however this has never been done so we can safely remove the tab. The information is already easily visible right above where the tab was.

Image post change:

Screenshot 2022-04-22 at 13 37 17


sldblog (approved):

sldblog:
Yay, cleanup 🎉

PR #1390, opened 2022-04-22
Email ro on referral completion deadline changes

kth13:

Amend the `Referral changes saved` message to indicate that responsible officer is also emailed

when the completion deadline is changed , we send email to the case worker and also the responsible officer. So the actual saved message should indicate the user that it is been sent to both caseworker and the responsible officer


server/routes/shared/showReferralView.ts
```diff
@@ -74,7 +74,7 @@ export default class ShowReferralView {
     const html = `<strong>Referral changes saved</strong>
           <hr class="govuk-section-break govuk-section-break--s">
           <p>
-             The service provider has been notified of the changes to this referral.
+          The service provider and your responsible officer have been notified about the change to this referral.
```
sldblog, line 77: should this contain something like "if there is one"? it's not always possible to notify the responsible officer (we have about 200 instances/day where we can't find their RO or they don't have emails)

JakeGoodwin22 (approved):

PR #1402, opened 2022-04-27
Enabling case worker sorting

kth13:

Enables case worker sorting in both probation practitioner and service provider dashboards. Currently its been disabled for all the users. The sorting happens based on the caseworker name.

The dashboard for both probation practitioner and service provider has got a caseworker column. As the number of referrals growing , there is a need to enable sorting so that users can sort and traverse to the intended records sooner.


server/config.ts
```diff
@@ -134,6 +134,7 @@ export default {
   },
   dashboards: {
     probationPractitioner: {
+      percentageOfPaginationUsers: Number(get('SP_PERCENTAGE_OF_DASHBOARD_PAGINATION_USERS', '0')),
```
sldblog, line 137: _nit_ It doesn't look like `dashboards.probationPractitioner.percentageOfPaginationUsers` is used anywhere (PP dashboards are always on) but it's not a blocker for this change

helm_deploy/values-preprod.yaml
```diff
@@ -14,5 +14,9 @@ generic-service:
     GA_ID: G-08D8T0RZZR
     FEATURE_SP_REPORTING_ENABLED: true
     SP_PERCENTAGE_OF_DASHBOARD_PAGINATION_USERS: 100
+    SP_MY_CASES_PAGE_SIZE: 10
+    SP_OPEN_CASES_PAGE_SIZE: 10
+    SP_UNASSIGNED_CASES_PAGE_SIZE: 10
+    SP_COMPLETED_PAGE_SIZE: 10
```
sldblog, line 20: _nit_ Did you mean to put this in dev? In preprod, we usually want to replicate the traffic/volumes/load as on production

sldblog (approved):
LGTM 👍 No blockers, couple of questions

server/config.ts
kth13: Ok sure David, I will remove it.

helm_deploy/values-preprod.yaml
kth13: Dev already the same config and I thought keeping the same thing in pre-prod. In prod, the config is

PR #1405, opened 2022-04-29
Updated errorHandler to convert errors into user friendly messages

JakeGoodwin22:

Changes to the authError template to allow parsing of HTML, allowing more specific and user friendly messages to be displayed to the user which is generated in the errorHandler based on the messages received from the API.

To give the user a better message explaining what the issue is and how they can fix it.

### DIFF
--- a/server/errorHandler.ts
+++ b/server/errorHandler.ts
@@ -3,6 +3,7 @@ import createError, { HttpError } from 'http-errors'
 import ControllerUtils from './utils/controllerUtils'
 import logger from '../log'
 import config from './config'
+import errorMessages from './utils/errorMessages'
 
 export default function createErrorHandler(production: boolean) {
   return (err: Error, req: Request, res: Response, next: NextFunction): void => {
@@ -20,9 +21,33 @@ export default function createErrorHandler(production: boolean) {
       if (err.status === 403 && (!err.external || err.response?.body?.accessErrors)) {
         res.status(403)
 
+        const returnedError = err.response?.body?.message
+        let userMessage
+        let userHeader
+
+        const userHeaderTypes = {
+          userHeaderService: 'You do not have permission to view this service',
+          userHeaderPage: 'You do not have permission to view this page',
+        }
+
+        const lookupError = errorMessages.returnedError[returnedError]
+        if (lookupError) {
+          userMessage = lookupError.mappedMessage
+          userHeader = userHeaderTypes[lookupError.userHeaderType]
+        } else if (returnedError?.includes('unidentified provider')) {
+          userMessage = errorMessages.errorHandlerAccessErrorMessages.providerGroupNotRecognised
+          userHeader = userHeaderTypes.userHeaderService
+        } else if (returnedError?.includes('unidentified contract')) {
+          userMessage = errorMessages.errorHandlerAccessErrorMessages.contractGroupNotRecognised
+          userHeader = userHeaderTypes.userHeaderService
+        } else if (returnedError?.includes('is not accessible to providers')) {
+          userMessage = errorMessages.errorHandlerAccessErrorMessages.groupsDoNotMatch
+          userHeader = userHeaderTypes.userHeaderService
+        }
+
         const args = {
-          message: err.response?.body?.message,
-          accessErrors: err.response?.body?.accessErrors,
+          userHeader,
+          userMessage,
         }
 
         return ControllerUtils.renderWithLayout(res, { renderArgs: ['errors/authError', args] }, null)
--- a/server/views/errors/authError.njk
+++ b/server/views/errors/authError.njk
@@ -4,20 +4,21 @@
     {% block content %}
         <div class="govuk-grid-row">
             <div class="govuk-grid-column-two-thirds">
-                <h1 class="govuk-heading-l">Sorry, you are not authorised to access this page</h1>
-
-                {% if message or accessErrors %}
-                    <p class="govuk-body">The following issues have been identified:</p>
-                    {% if message %}
-                        <p class="govuk-body"><strong>{{ message }}</strong></p>
-                    {% endif %}
-                    {% if accessErrors %}
-                        <ul class="govuk-list govuk-list--bullet">
-                            {% for err in accessErrors %}
-                                <li>{{ err }}</li>
-                            {% endfor %}
-                        </ul>
-                    {% endif %}
+                {% if not userHeader %}
+                    <h1 class="govuk-heading-l">Sorry, you are not authorised to access this page</h1>
+                {% endif %}
+                {% if userHeader %}
+                    <h1 class="govuk-heading-l">{{ userHeader }}</h1>
+                {% endif %}
+                {% if userMessage %}
+                    <p class="govuk-body">{{ userMessage | safe }}</p>
+                {% endif %}
+                {% if accessErrors %}
+                    <ul class="govuk-list govuk-list--bullet">
+                        {% for err in accessErrors %}
+                            <li>{{ err }}</li>
+                        {% endfor %}
+                    </ul>
                 {% endif %}
             </div>
         </div>
--- a/server/utils/errorMessages.ts
+++ b/server/utils/errorMessages.ts
@@ -2,6 +2,54 @@
 // Explicit return types would make this file very messy and not much more
 // informational — everything returned is a string, since they’re messages!
 
+const errorHandlerAccessErrorMessages = {
+  notSetUpCorrectly:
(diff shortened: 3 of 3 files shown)


server/views/errors/authError.njk
```diff
@@ -3,23 +3,23 @@
...
+                        {% endif %}
+                        {% if message or accessErrors %}
+                            {% if message %}
+                                <h1 class="govuk-heading-xl">{{ message }}</h1>
+                            {% endif %}
+                            {% if accessErrors %}
+                                <ul class="govuk-list govuk-list">
+                                    {% for err in accessErrors %}
+                                        <li>{{ err | safe }}</li>
+                                    {% endfor %}
```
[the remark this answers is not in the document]
JakeGoodwin22: safe is used to escape html, this is used as we are passing through <a> tags and we want it to be picked us such rather than a string

server/errorHandler.ts
```diff
@@ -20,9 +20,39 @@ export default function createErrorHandler(production: boolean) {
       if (err.status === 403 && (!err.external || err.response?.body?.accessErrors)) {
         res.status(403)
 
+        const userHeaderService = 'You do not have permission to view this service'
+        const userHeaderPage = 'You do not have permission to view this page'
+        const returnedError = err.response?.body?.message
+        let userMessage
+        let userHeader
+
+        if (returnedError === 'could not map service provider user to access scope') {
```
kth13, line 29: @JakeGoodwin22 is there any way to externalise these messages to an error-mapping.properties (error -> errormessage) or something similar. So, in the future , adding or amending the messages would be easier. This is just a suggestion, please consider this if it is feasible.

rgforsyth: There is an error messages file, if that's what you mean. `server/utils/errorMessages.ts `

server/errorHandler.ts
```diff
@@ -20,9 +20,39 @@ export default function createErrorHandler(production: boolean) {
...
+          userMessage =
+            'Your account is not set up correctly. Ask an admin user in your organisation to fix this in HMPPS Digital Services.'
+          userHeader = userHeaderPage
+        } else if (returnedError === 'cannot find user in hmpps auth') {
+          userMessage =
+            'Your email address is not recognised. If it has changed recently, try signing out and signing in with the correct one. Ask an admin user in your organisation to check what the right email is in HMPPS Digital Services. If that does not work, <a target="_blank" href="https://hmpps-interventions-ui-dev.apps.live-1.cloud-platform.service.justice.gov.uk/report-a-problem">report it as a problem.</a>'
+          userHeader = userHeaderPage
+        } else if (returnedError?.includes('unidentified provider')) {
+          userMessage =
+            'Your provider group is not recognised. Ask an admin in your organisation to check it has been set up correctly in HMPPS Digital Services. They may need to <a target="_blank" href="https://hmpps-interventions-ui-dev.apps.live-1.cloud-platform.service.justice.gov.uk/report-a-problem">report it as a problem.</a>'
```
sldblog, line 43: > href="https://hmpps-interventions-ui-dev.apps.live-1.cloud-platform.service.justice.gov.uk/report-a-problem"

As-is, this link will be used in production as well -- can we use relative links or use the environment-specific host?

server/views/errors/authError.njk
```diff
@@ -4,21 +4,22 @@
...
+                {% if not userHeader %}
+                    <h1 class="govuk-heading-l">Sorry, you are not authorised to access this page</h1>
+                {% endif %}
+                {% if userHeader %}
+                    <h1 class="govuk-heading-l">{{ userHeader }}</h1>
+                {% endif %}
+                {% if userMessage %}
+                    <p class="govuk-body">{{ userMessage | safe }}</p>
+                {% endif %}
                     {% if accessErrors %}
```
sldblog, line 16: _nit, not a blocker_ Indentation seems to be off -- do we have a linter to catch these?

sldblog (commented):
Changes LGTM, the only blocker for me would be using "dev" links that propagate to all environments

_To summarise discussions about this topic in Slack and other PRs_
The change solves the problem but introduces the trade-off of 'surprising' coupling (someone might correct a phrase and behaviour changes in the UI). As a next step, we'd benefit from adding a set of authorisation problem codes that are part of the API

server/errorHandler.ts
JakeGoodwin22: Ah yes sorry my mistake I just copied directly from the requirements! can change this

server/errorHandler.ts
JakeGoodwin22: Happy to move these to the errorMessages file, thanks!

sldblog (approved):
LGTM 👍 
Could someone with more TS/node knowledge also take a look please?

server/errorHandler.ts
```diff
@@ -20,9 +21,34 @@ export default function createErrorHandler(production: boolean) {
...
+
+        if (returnedError === 'could not map service provider user to access scope') {
+          userMessage = errorMessages.errorHandlerAccessErrorMessages.notSetUpCorrectly
+          userHeader = userHeaderService
+        } else if (returnedError === 'user is not a service provider') {
+          userMessage = errorMessages.errorHandlerAccessErrorMessages.notSetUpCorrectly
+          userHeader = userHeaderPage
+        } else if (returnedError === 'cannot find user in hmpps auth') {
+          userMessage = errorMessages.errorHandlerAccessErrorMessages.emailNotRecognised
+          userHeader = userHeaderPage
```
NeilLuptonMoJ, line 38: adding error message will involve adding more and more `else if` checks, could we not move this to a lookup ie in `errorMessages.ts`

```
const errorHandlerAccessErrorMessages = {
  notSetUpCorrectly:
    'Your account is not set up correctly. Ask an admin user in your organisation to fix this in HMPPS Digital Services.',
  emailNotRecognised:
    'Your email address is not recognised. If it has changed recently, try signing out and signing in with the correct one. Ask an admin user in your organisation to check what the right email is in HMPPS Digital Services. If that does not work, <a target="_blank" href="/report-a-problem">report it as a problem.</a>',
  providerGroupNotRecognised:
    'Your provider group is not recognised. Ask an admin in your organisation to check it has been set up correctly in HMPPS Digital Services. They may need to <a target="_blank" href="/report-a-problem">report it as a problem.</a>',
  contractGroupNotRecognised:
    'Your contract group is not recognised. Ask an admin in your organisation to check it has been set up correctly in HMPPS Digital Services. They may need to <a target="_blank" href="/report-a-problem">report it as a problem.</a>',
}
const userHeaderTypes = {
  userHeaderService: 'You do not have permission to view this service',
  userHeaderPage: 'You do not have permission to view this page',
  default: 'Sorry, you are not authorised to access this page',
}
const returnedError = {
  'could not map service provider user to access scope': {
    mappedMessage: errorHandlerAccessErrorMessages.notSetUpCorrectly,
    userHeaderType: userHeaderTypes.userHeaderService,
  },
}
```
(then export these variables)

and change `errorHandler.ts` :

```
        const userHeaderTypes = {
          userHeaderService: 'You do not have permission to view this service',
          userHeaderPage: 'You do not have permission to view this page',
        }

        const lookupError = errorMessages.returnedError[returnedError]
        if (lookupError) {
          userMessage = lookupError.mappedMessage
          userHeader = userHeaderTypes[lookupError.userHeaderType]
        }
```
(and lookup the other userHeader types from userMessages)

doesn't help with the `.includes` errors but should make it easier to add new errors and removes hard coded messages from `errorHandler.ts`

server/errorHandler.ts
```diff
@@ -20,9 +21,33 @@ export default function createErrorHandler(production: boolean) {
...
+        } else if (returnedError?.includes('unidentified provider')) {
+          userMessage = errorMessages.errorHandlerAccessErrorMessages.providerGroupNotRecognised
+          userHeader = userHeaderTypes.userHeaderService
+        } else if (returnedError?.includes('unidentified contract')) {
+          userMessage = errorMessages.errorHandlerAccessErrorMessages.contractGroupNotRecognised
+          userHeader = userHeaderTypes.userHeaderService
+        } else if (returnedError?.includes('is not accessible to providers')) {
+          userMessage = errorMessages.errorHandlerAccessErrorMessages.groupsDoNotMatch
+          userHeader = userHeaderTypes.userHeaderService
+        }
```
NeilLuptonMoJ, line 46: still needs the 
```
        } else {
          userHeader = userHeaderTypes.default
        }
```
otherwise it will return userHeader undefined for any unknown error message

JakeGoodwin22: This works without the else, by default we display the message if it's undefined in authError

NeilLuptonMoJ (approved):

PR #1423, opened 2022-05-03
Updated referral cancelled text to tell PP's they need to inform SP's

JakeGoodwin22:

Updates the text on the referral cancelled page.

Currently, emails are not being sent out when a referral is cancelled - the text needs to be changed as a temporary solution to let the PP's know it is their responsibility to inform SP's

### DIFF
--- a/server/routes/referral/cancellation/confirmation/referralCancellationConfirmationPresenter.test.ts
+++ b/server/routes/referral/cancellation/confirmation/referralCancellationConfirmationPresenter.test.ts
@@ -11,8 +11,9 @@ describe(ReferralCancellationConfirmationPresenter, () => {
 
       expect(presenter.text).toMatchObject({
         confirmationText: 'This referral has been cancelled',
+        headingText: 'What you need to do next',
         whatHappensNextText:
-          "Service provider will be notified about the cancellation. You don't have to do anything else.",
+          'You need to contact the service provider outside the service to let them know about the change.',
       })
     })
   })
--- a/server/routes/referral/cancellation/confirmation/referralCancellationConfirmationPresenter.ts
+++ b/server/routes/referral/cancellation/confirmation/referralCancellationConfirmationPresenter.ts
@@ -9,8 +9,8 @@ export default class ReferralCancellationConfirmationPresenter {
 
   readonly text = {
     confirmationText: 'This referral has been cancelled',
-    whatHappensNextText:
-      "Service provider will be notified about the cancellation. You don't have to do anything else.",
+    headingText: 'What you need to do next',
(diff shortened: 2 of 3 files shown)


server/routes/referral/cancellation/confirmation/referralCancellationConfirmationPresenter.test.ts
```diff
@@ -11,8 +11,9 @@ describe(ReferralCancellationConfirmationPresenter, () => {
 
       expect(presenter.text).toMatchObject({
         confirmationText: 'This referral has been cancelled',
+        headingText: 'What you need to do next',
         whatHappensNextText:
-          "Service provider will be notified about the cancellation. You don't have to do anything else.",
+          'You need to contact the service provider outside the service to let them know about the change.',
```
sldblog, line 16: Can we include the address, to make this more actionable?

JakeGoodwin22: I've updated this to add the email. Speaking with Helen because she wants to check this is the correct approach first before it is merged.

sldblog: The change made me realise that we don't actually have an intended "contact" email for providers, sorry 🤦
The `incomingReferralDistributionEmail` is invalid for 11 contracts...

Possibly something to pick up as a follow-up

JakeGoodwin22: No problem! Can revert for this change, it was a useful learning experience adding that in if nothing else :)

sldblog (approved):
👍 As a next step, we can possibly look at onboarding the contact phone number from the [spreadsheets (sheet 'CAS information' cell C7)](https://github.com/ministryofjustice/hmpps-interventions-seed-data/tree/main/real-data/interventions)

server/routes/referral/cancellation/confirmation/referralCancellationConfirmationPresenter.ts
```diff
@@ -9,8 +9,8 @@ export default class ReferralCancellationConfirmationPresenter {
 
   readonly text = {
     confirmationText: 'This referral has been cancelled',
-    whatHappensNextText:
-      "Service provider will be notified about the cancellation. You don't have to do anything else.",
+    headingText: 'What you need to do next',
+    whatHappensNextText: `You need to contact the service provider outside the service to let them know about the change.`,
```
sldblog, line 13: _nit_ `whatHappensNextText` implies the heading should always be `What you need to do next`

PR #1426, opened 2022-05-06
Add delay after each command to slow videos down

NeilLuptonMoJ:
Creating videos from the Cypress tests results in very high speed output,
this adds delays to each step to make the output eaiser to view.

Adds a delay, configured by an env setting to each command

Should allow us to set the cypress tests up as user journeys, which can then be outputted as videos so that they can be used as demos.


integration_tests/support/delay.js
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable no-restricted-syntax */
+
+// Set COMMAND_DELAY above zero for demoing to stakeholders,
+const COMMAND_DELAY = Cypress.env('COMMAND_DELAY') || 0
```
sldblog, line 4: Genuine question, not a veiled suggestion: should this be namespaced to `CYPRESS_COMMAND_DELAY`?

package.json
```diff
@@ -25,6 +25,7 @@
     "test": "jest --runInBand",
     "int-test": "cypress run --config video=false",
     "int-test-ui": "cypress open",
+    "int-test-video": "CYPRESS_NO_COMMAND_LOG=1 cypress run --config video=true,slowTestThreshold=15000 --env COMMAND_DELAY=1200",
```
sldblog, line 28: Do we need the `slowTestThreshold`? Seems dissimilar to the rest of the changes

sldblog (approved):
LGTM 👍 No blockers

package.json
sldblog: Is this is to avoid the delay triggering slow test warnings?

NeilLuptonMoJ: yeah, adding the delays causes all the tests to trigger the slow warnings, so resetting the threshold for these tests

integration_tests/support/delay.js
NeilLuptonMoJ: Good point, I'll update this

PR #1438, opened 2022-05-10
Updated docker-compose to the most recent version of postgres

JakeGoodwin22:

Uses the most recent version of postgres which is the version used by the service to fix the deprecated image version

Prevents us from having an unsupported postgres image in the future.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -58,7 +58,7 @@ services:
       - ASSESSRISKSANDNEEDS_BASEURL=http://assess-risks-and-needs:8080
 
   postgres:
-    image: postgres:10-alpine
+    image: postgres:14.2-alpine
     container_name: postgres
     networks:
       - hmpps


docker-compose.yml
```diff
@@ -58,7 +58,7 @@ services:
       - ASSESSRISKSANDNEEDS_BASEURL=http://assess-risks-and-needs:8080
 
   postgres:
-    image: postgres:10-alpine
+    image: postgres:14.2
```
sldblog, line 61: for local, can we stick to -alpine version? about ~150MB less pulling to do
```
REPOSITORY TAG                     IMAGE ID       CREATED         SIZE
postgres   14.2                    74b0c105737a   2 weeks ago     376MB
postgres   14.2-alpine             ebf01b748a56   5 weeks ago     211MB
```

sldblog (commented):
LGTM in general (one comment about image size)

However, can we block this until there's a CI pipeline working with the service on PostgreSQL 14? That way only versions that _work_ on v14 and v10 would get built.

NeilLuptonMoJ (approved):

PR #1439, opened 2022-05-10
Add a retry to failing tests

NeilLuptonMoJ:
Intermittent `element is detached from the DOM.` errors are causing the
cypress tests to fail, this will retry the failing tests

Adds retry parameter to failing tests 
(`runMode` for `cypress run` commands and `openMode` for `cypress open` commands)

Also moves cy.stub setup to beforeEach, to try and reduce likelihood of tests running while setup is completing

To try and stop intermittent failures of integration_test pipeline


sldblog (approved):
Thanks! Would be good to understand the root cause as well at some point

PR #1441, opened 2022-05-11
Updated Structured Interventions spreadsheet and content

JakeGoodwin22:

Updates the structured interventions spreadsheet, updates the content and removes the PDF version of the file.

Provide PP's with an updated list of structured interventions.


server/routes/probationPractitionerReferrals/probationPractitionerReferralsController.ts
```diff
@@ -124,8 +124,7 @@ export default class ProbationPractitionerReferralsController {
     const existingDraftReferrals = await this.interventionsService.getDraftReferralsForUserToken(accessToken)
 
     const structuredInterventionsFileDownloadPaths = {
-      xlsx: 'assets/downloads/Structured interventions list v1_3_3.xlsx',
-      pdf: 'assets/downloads/Structured interventions list v1_3_3.pdf',
+      xlsx: 'assets/downloads/Structured interventions.xlsx',
```
sldblog, line 127: Assuming people download this and we update it after -- can we help people understand that they have an outdated version?

server/routes/probationPractitionerReferrals/findStartPresenter.test.ts
```diff
@@ -47,29 +46,18 @@ describe('FindStartPresenter', () => {
...
-        referrals,
-        { xlsx: 'example.xlsx', pdf: 'example.pdf' },
-        { bytes: 0 },
-        loggedInUser
-      )
+      const presenter = new FindStartPresenter(referrals, { xlsx: 'example.xlsx' }, { bytes: 0 }, loggedInUser)
 
       expect(presenter.structuredInterventionsDownloadHrefs).toEqual({
         xlsx: '/example.xlsx',
-        pdf: '/example.pdf',
```
sldblog, line 72: It seems like removing the PDF is against an [acceptance criterion](https://trello.com/c/IxPw3Ovj):

> Given that I am interested in learning what Structured Interventions are available
> When I view the PDF version of the spreadsheet
> Then it matches the current spreadsheet version

Is that ok?

sldblog (commented):
Changes LGTM in general

There's an AC in the [ticket](https://trello.com/c/IxPw3Ovj) that says

> Given that I have bookmarked the 'old' spreadsheet location
> When I attempt to navigate directly to it via my bookmark
> Then I am redirected to the new spreadsheet

which relates to the comment below about the "knowing PP has the latest" question

It feels risky to have local copies for PPs who could think they have the latest, when they don't

server/routes/probationPractitionerReferrals/probationPractitionerReferralsController.ts
JakeGoodwin22: Spoken with Sarah on this - I've added versioning back in so it's now v1.3.4 to show the user that they are using a newer copy. In terms of bookmarking the file, not sure how this could work unless we forever more wanted to hard-code 1.3.3 in which doesn't seem like a good approach?

server/routes/probationPractitionerReferrals/findStartPresenter.test.ts
JakeGoodwin22: Confirmed this is OK, has now been removed from the AC

NeilLuptonMoJ (approved):

PR #1442, opened 2022-05-11
updated the pact test cases to the new sent-referrals

kth13:

updates the pact test for the new API `/sent-referrals/summaries` and remove the pact test for the deprecated API `/sent-referrals`

We have published a new API `/sent-referrals/summaries`, so it requires the pact test to be updated to test the new API. Also, we have deprecated the old API `/sent-referrals`, so it requires the pact test to be removed for that API.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "start:dev": "npm run build && NODE_ENV=development npm run watch",
     "start:test": "export $(cat test.env) && npm run start",
     "test": "jest --runInBand",
+    "test-debug": "npm test --watch -t",
     "int-test": "cypress run --config video=false",
     "int-test-ui": "cypress open",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -23,6 +23,7 @@
     "start:dev": "npm run build && NODE_ENV=development npm run watch",
     "start:test": "export $(cat test.env) && npm run start",
     "test": "jest --runInBand",
+    "test-debug": "npm test --watch -t",
```
rgforsyth, line 26: was this intended to be left in?

rgforsyth (approved):

package.json
kth13: @rgforsyth It is intended as we felt it will be beneficial later for debugging.

PR #1451, opened 2022-05-17
Adding contract type name in to draft referral

Froom2:

This pull request adds the Contract Type Name to the draft referral model, and adds a column to display this in the find interventions draft referral table.
Also updates column heading in the same table to change 'Service user' to 'Name'

The intent is to provide additional information to the user so that they can better identify draft referrals.


JakeGoodwin22 (approved):

JakeGoodwin22:
Approved - but presumably service will need to be merged in first before the UI

PR #1470, opened 2022-05-26
Added accessibility statement to footer

JakeGoodwin22:

Add the accessibility statement to the page footer.

It is good practice and a requirement for us to have an accessibility statement to be visible


JakeGoodwin22:
Change is work in progress - as part of the change I think the cypress tests should be updated to be like a user journey rather than jumping around so want to update this.

Froom2 (approved):

PR #1478, opened 2022-05-31
Fix integration tests for non-chromium browsers

NeilLuptonMoJ:

Fixes integration tests for Firefox

Running int tests using Firefox fails in login.spec.js with:
`CypressError: Cypress detected a cross origin error happened on page load:`
The `chromeWebSecurity` flag is used to get past this in Chrome
This change removes the need for that flag + allows tests to be run in Firefox
when calling `/sign-out` `passport.js` redirects to the auth service endpoint '/auth/sign-out'
this in turn redirects back to `/sign-out/success`
this change bypasses auth service and does the same redirect
removing the need for any cross-origin handling
`chromeWebSecurity` flag removed so Chrome runs the same as Firefox

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -1,5 +1,4 @@
 {
-    "chromeWebSecurity": false,
     "baseUrl": "http://localhost:3007",
     "ignoreTestFiles": "**/!(*.spec).js",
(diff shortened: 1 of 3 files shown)


JakeGoodwin22 (approved):
LGTM

PR #1492, opened 2022-06-06
Use Page Not Available instead of throwing error

NeilLuptonMoJ:

https://trello.com/c/rFs1K8KR
Use soft deleted page instead of outputting error

Stops error message, which includes error stack, from being shown to users
(note error message text will be updated in separate PR)

### DIFF
--- a/server/routes/appointments/appointmentsController.test.ts
+++ b/server/routes/appointments/appointmentsController.test.ts
@@ -164,11 +164,9 @@ describe('Scheduling a supplier assessment appointment', () => {
 
         await request(app)
           .get(`/service-provider/referrals/1/supplier-assessment/schedule/abc/details`)
-          .expect(500)
+          .expect(410)
           .expect(res => {
-            expect(res.text).toContain(
-              'Too much time has passed since you started booking this appointment. Your answers have not been saved, and you will need to start again.'
-            )
+            expect(res.text).toContain('This page is no longer available')
           })
       })
     })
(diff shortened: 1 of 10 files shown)


server/routes/appointments/appointmentsController.test.ts
```diff
@@ -164,11 +164,9 @@ describe('Scheduling a supplier assessment appointment', () => {
 
         await request(app)
           .get(`/service-provider/referrals/1/supplier-assessment/schedule/abc/details`)
-          .expect(500)
+          .expect(410)
```
sldblog, line 167: 👏

server/utils/controllerUtils.ts
```diff
@@ -30,26 +29,16 @@ export default class ControllerUtils {
...
 
-    if (draft === null) {
-      throw createError(500, `UI-only draft ${typeName} with ID ${id} not found`, {
-        userMessage: notFoundUserMessage,
-      })
-    }
-
-    if (draft.softDeleted) {
+    if (draft === null || draft.softDeleted) {
       res.status(410 /* Gone */)
```
sldblog, line 40: _nit_
Is there a constant in Node/Express that could be used?

server/utils/controllerUtils.ts
```diff
@@ -30,26 +29,16 @@ export default class ControllerUtils {
...
-      const view = new DraftSoftDeletedView()
+      const view = new DraftSoftDeletedOrNotFoundView()
```
sldblog, line 41: _semantics_
Along with the notion of "intertwined things increase semantic complexity", would it be worth having separate views for the two sides? Expired/deleted/missing might have separate needs

NeilLuptonMoJ: Good point, I've split it into two views

Froom2 (approved):

PR #1493, opened 2022-06-08
Split referrals to draft and live referrals

kth13:

- It introduces a new table draft referral
- All draft referrals will be residing in this table instead of referral table
- once the draft referrals is sent, it is moved to referral table
- draft referral data will not be deleted as part of the first release
- The existing draft referral in the referral table will be deleted after the migration is done

- referral table was growing and we saw draft referrals holds 30% of data
- By segregating draft referrals , we can fetch only the data that we need. This improves the performance of the queries

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -117,7 +117,6 @@ workflows:
             - integration_test
             - hmpps/helm_lint
             - build_docker_publish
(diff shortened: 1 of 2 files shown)


kth13:
Opened by mistake

PR #1494, opened 2022-06-10
Restrict appointments date input fields to prevent invalid values

JakeGoodwin22:

Adds additional validation to the date input fields where appointments can be scheduled. The change prevents appointments before the date the referral was made and no later than 6 months from today.

Users could be entering invalid dates, it is impossible to schedule an appointment to before the referral has been made so we should prevent this from happening. In addition, we don't want appointments being set too far in the future as it is likely a mistake - both points have been agreed with stakeholders and shouldn't be allowed going forward.

### DIFF
--- a/server/routes/appointments/appointmentsController.test.ts
+++ b/server/routes/appointments/appointmentsController.test.ts
@@ -233,7 +233,7 @@ describe('Scheduling a supplier assessment appointment', () => {
         expect(draftsService.updateDraft).toHaveBeenCalledWith(
           draftBooking.id,
           {
-            appointmentTime: `${dateOfAppointment.toISOString().split('T')[0]}T08:02:00.000Z`,
+            appointmentTime: '2022-03-01T09:02:00.000Z',
             durationInMinutes: 75,
             sessionType: 'ONE_TO_ONE',
             appointmentDeliveryType: 'PHONE_CALL',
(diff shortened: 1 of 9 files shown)


server/routes/appointments/appointmentsController.test.ts
```diff
@@ -610,13 +614,15 @@ describe('Scheduling a delivery session', () => {
...
           .post(`/service-provider/action-plan/${actionPlan.id}/sessions/1/edit/${draftBooking.id}/details`)
           .type('form')
           .send({
-            'date-day': '24',
-            'date-month': '3',
-            'date-year': '2025',
+            'date-year': today.getFullYear(),
+            'date-month': today.getUTCMonth() + 1,
+            'date-day': today.getUTCDate(),
             'time-hour': '9',
```
NeilLuptonMoJ, line 626: I think these tests are going to start failing as soon as the clocks change - the test is now sending in the current date instead of a fixed date, and the following `expect(draftsService.updateDraft)` checks for a time of `T08:02`

JakeGoodwin22: Thanks, had convo with David on Friday about this too - he suggested we should be using static data for testing to keep things predictable and simple for us. I've made that change which will solve this problem too

PR #1509, opened 2022-06-13
Restrict dates input fields

JakeGoodwin22:

Adds additional validation to date fields for appointments to restrict dates earlier than before the referral was made and to no later than 6 month from today.

Prevent users entering erroneous dates, it also fixes a bug where if a user enters a date far in the past it will throw an exception if they go to view the referral again

### DIFF
--- a/server/routes/appointments/appointmentsController.test.ts
+++ b/server/routes/appointments/appointmentsController.test.ts
@@ -209,9 +209,9 @@ describe('Scheduling a supplier assessment appointment', () => {
           .post(`/service-provider/referrals/${referral.id}/supplier-assessment/schedule/${draftBooking.id}/details`)
           .type('form')
           .send({
-            'date-day': '24',
+            'date-day': '5',
             'date-month': '3',
-            'date-year': '2025',
+            'date-year': '2022',
             'time-hour': '9',
             'time-minute': '02',
             'time-part-of-day': 'am',
@@ -229,7 +229,7 @@ describe('Scheduling a supplier assessment appointment', () => {
         expect(draftsService.updateDraft).toHaveBeenCalledWith(
           draftBooking.id,
           {
-            appointmentTime: '2025-03-24T09:02:00.000Z',
+            appointmentTime: '2022-03-05T09:02:00.000Z',
             durationInMinutes: 75,
             sessionType: 'ONE_TO_ONE',
             appointmentDeliveryType: 'PHONE_CALL',
@@ -255,7 +255,7 @@ describe('Scheduling a supplier assessment appointment', () => {
             .send({
               'date-day': '32',
               'date-month': '3',
-              'date-year': '2021',
+              'date-year': 'test',
               'time-hour': '9',
               'time-minute': '02',
               'time-part-of-day': 'am',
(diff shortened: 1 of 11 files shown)


server/routes/appointments/appointmentsController.test.ts
```diff
@@ -209,9 +213,9 @@ describe('Scheduling a supplier assessment appointment', () => {
           .post(`/service-provider/referrals/${referral.id}/supplier-assessment/schedule/${draftBooking.id}/details`)
           .type('form')
           .send({
-            'date-day': '24',
-            'date-month': '3',
-            'date-year': '2025',
+            'date-day': dateOfAppointment.getUTCDate(),
```
NeilLuptonMoJ, line 216: can we keep the test data as hard-coded values - having the values calculated like this can be confusing when looking at this test as you have to go to the referral factory to read what the date is actually being set as

NeilLuptonMoJ (commented):
I think should change the tests to use hard coded values, instead of calculating them, we can use the mockdate package to set the system time to make sure the tests run consistently

server/utils/forms/inputs/calendarDayInput.test.ts
```diff
@@ -182,12 +185,16 @@ describe(CalendarDayInput, () => {
 
   describe('.createErrors', () => {
     it('creates a formError object from the key and message passed in', () => {
-      expect(CalendarDayInput.createError('deadline', 'The deadline must be a real date')).toEqual({
+      const maximumDate = new Date()
```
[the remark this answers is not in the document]
JakeGoodwin22: Don't think we actually need to use a proper date for these tests, it was just used for the error message so we can just type something into the string instead. Mockdate looks useful but not sure it's needed here

server/routes/appointments/appointmentsController.test.ts
JakeGoodwin22: Yep, this is no longer needed now that the mocked referral doesn't use new Date() anymore. It has been hardcoded to 1st Jan 2022 so we're safe to hard-code these as well

server/utils/forms/inputs/calendarDayInput.test.ts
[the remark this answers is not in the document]
JakeGoodwin22: I've found some other tests which I think MockDate will be useful for in another class, so going to update those

NeilLuptonMoJ (approved):

server/utils/forms/inputs/calendarDayInput.ts
```diff
@@ -95,8 +96,52 @@ export default class CalendarDayInput {
...
+          Number(this.request.body[this.keys.month] - 1),
+          Number(this.request.body[this.keys.day]),
+          23,
+          59,
+          59
+        )
+
+        if (this.referralDate != null) {
+          const maximumDate = new Date()
+          maximumDate.setMonth(maximumDate.getMonth() + 6)
```
sldblog, line 116: @JakeGoodwin22 Thanks for these!

As this is a business rule, can we also implement this in the domain API, please? It should reject bookings with appointment times not meeting this criterion.

PR #1511, opened 2022-06-13
Update README.md

JakeGoodwin22:

_Required._

_Required._


sldblog:
_guessing from the branch name_
In case it's useful, the only PR-specific pipeline we have is the auto-approve dependencies job, so committing on a branch is enough for CI checks to run

PR #1514, opened 2022-06-14
Remove risk information for PP user test

NeilLuptonMoJ:

Removes invalid test in showReferralPresenter.test.ts

canShowFullSupplementaryRiskInformation doesn't check user type, so this isn't a valid test
If the `ASSESS_RISKS_AND_NEEDS_API_RISK_SUMMARY_ENABLED` env variable is set to true this test fails


JakeGoodwin22 (approved):

sldblog:
Can we also remove `ASSESS_RISKS_AND_NEEDS_API_RISK_SUMMARY_ENABLED`? It seems to be fully released now.

PR #1527, opened 2022-06-20
Updated ui service contract testing dates for restricting dates greater than 6 months

JakeGoodwin22:

The API now restricts dates greater than 6 months in the future, tests are updated to meet the new restrictions

To maintain the contract tests so that they pass for the API


kth13 (approved):
LGTM

PR #1528, opened 2022-06-23
Skip dev dependencies in security audit

NeilLuptonMoJ:

Skips dev dependencies when running audit

devDependencies are not installed in production, so dev vulnerabilities should not stop nightly checks from running
(raised as a result of vulnerability in a `nodemon` dependency)

### DIFF
--- a/audit-ci.json
+++ b/audit-ci.json
@@ -1,5 +1,6 @@
 {
   "moderate": true,
   "package-manager": "auto",
-  "registry": "https://registry.npmjs.org"
+  "registry": "https://registry.npmjs.org",
+  "skip-dev": true
 }


JakeGoodwin22 (approved):

sldblog:
Some of the corrupted libraries upload things to a shady server. Local usage could then potentially leak things as well. In this case, would it make sense to add an exclusion for a specific CVE?

sldblog (approved):
👍 Got it, thanks. Yeah, since the intent of these jobs is to validate production's security, I think this is fine.

It's another story that it does not validate production 😂 (`:latest`/`main` isn't always deployed)

PR #1543, opened 2022-06-28
Roll out pagination to all dashboard users

sldblog:

Roll out pagination to all dashboard users

Following a production outage today which was caused by unpaginated
queries peaking the production database usage at 100%, causing cascading
outage for all users.

As a mitigation, we have enabled pagination temporarily, which
stabilised the service immediately.

This change makes the rollout permanent.


NeilLuptonMoJ (approved):
👍

PR #1547, opened 2022-06-30
Fix for all open cases showing as search error when no cases exist

JakeGoodwin22:

Implements a fix for removing the search error when no cases exist for the open cases tab

Users should be able to see an empty table rather than a search error to make it clearer that they have no open cases

### DIFF
--- a/server/routes/serviceProviderReferrals/serviceProviderReferralsController.test.ts
+++ b/server/routes/serviceProviderReferrals/serviceProviderReferralsController.test.ts
@@ -269,10 +269,7 @@ describe('GET /service-provider/dashboard/all-open-cases', () => {
       .expect(200)
       .expect(res => {
         expect(res.text).toContain('All open cases')
-        expect(res.text).not.toContain('Alex River')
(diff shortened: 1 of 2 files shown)


NeilLuptonMoJ (approved):
👍

server/routes/serviceProviderReferrals/serviceProviderReferralsController.test.ts
```diff
@@ -269,10 +269,7 @@ describe('GET /service-provider/dashboard/all-open-cases', () => {
       .expect(200)
       .expect(res => {
         expect(res.text).toContain('All open cases')
-        expect(res.text).not.toContain('Alex River')
-        expect(res.text).not.toContain('Accommodation Services - West Midlands')
-        expect(res.text).not.toContain('George River')
-        expect(res.text).toContain(`There are no results for`)
+        expect(res.text).toContain('Date received') // Proves tables existence
```
sldblog, line 272: The test is named "all open cases and no cases when there are no open cases to be viewed" -- I wonder if it should be renamed to cover what the expectation is?

PR #1549, opened 2022-07-01
Differentiate session history

OmotayoTalabi:
Added functionality to differentiate the session history

to create clarity to the users on the sessions history


NeilLuptonMoJ (approved):
Not sure escape is needed for sessionNumber but it doesn't really matter

PR #1576, opened 2022-07-12
Update passport to v0.6.0

NeilLuptonMoJ:

https://trello.com/c/8kv0SLcX

* Updates passport dependency to v0.6.0
breaking change - req.logout now requires callback
(the only place this is mentioned in docs is on the strategies page)
follow [example pattern](https://www.passportjs.org/tutorials/auth0/logout/) for logout processing
for sign-in add option `keepSessionInfo` - the latest version of passport uses session.regenerate to create a clean new session on login, however this is losing `session.returnTo` which is needed for redirect to work correctly 

* use `unset: destroy` express-session option - so `session.destroy` not needed (see [docs](https://github.com/expressjs/session#unset))

Changes required to update to latest version of passport, to fix GHSA-v923-w3x8-wh69 advisory


server/authentication/passport.ts
```diff
@@ -97,26 +97,34 @@ export default function passportSetup(app: Application, hmppsAuthService: HmppsA
   app.get('/sign-in', signIn)
 
   const signInCallback = (redirectUri: string) =>
+    // TODO - remove ts-ignore once keepSessionInfo has been added to types/passport
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore - No overload matches this call. - keep this until @types/passport updated
```
NeilLuptonMoJ, line 102: `keepSessionInfo` is needed to stop req.session from being regenerated on sign in - 
when you pass in `successReturnToOrRedirect` the processing should try to redirect back to the location stored in `req.session.returnTo` on success (or `/` if returnTo is not populated)
However passport has been changed to use session.regenerate on login - this recreates the req.session object resulting in `req.session.returnTo` being dropped - so the redirect will never work correctly

a PR has been raised with definitelyTyped to update the @types/passport definition, but this could take a while to go through approval/release, so temporarily override TS error caused by unknown option

server/authentication/passport.ts
```diff
@@ -97,26 +97,34 @@ export default function passportSetup(app: Application, hmppsAuthService: HmppsA
...
 
-  const signOut = (logoutTerm: string, req: Request, res: Response) => {
+  const signOut = (logoutTerm: string, req: Request, res: Response, next: NextFunction) => {
     const { redirectUrl } = req.query
     const logoutUrl = `${authConfig.url}/${logoutTerm}?client_id=${authConfig.loginClientId}&redirect_uri=${
       config.domain
     }${redirectUrl || `/${logoutTerm}/success`}`
     if (req.isAuthenticated()) {
-      req.logout()
-      req.session!.destroy(() => res.redirect(logoutUrl))
```
NeilLuptonMoJ, line 113: change to add callback function (copied from [example pattern](https://www.passportjs.org/tutorials/auth0/logout/))

note `req.session.destroy` removed as `unset` option added to session in `app.ts`

NeilLuptonMoJ:
[🐛 Passport security vulnerability](https://trello.com/c/8kv0SLcX/441-%F0%9F%90%9B-passport-security-vulnerability)

kth13 (approved):
LGTM with a minor question/comment

rgforsyth (approved):

PR #1577, opened 2022-07-12
refactor dashboard tests

NeilLuptonMoJ:

split tests instead of running under a single block
use standard h1 check to make sure the page loaded
move any common beforeEach blocks up to higher level

try to make tests run more reliably

https://trello.com/c/h0viOeYR

### DIFF
--- a/integration_tests/integration/dashboards.cy.js
+++ b/integration_tests/integration/dashboards.cy.js
@@ -168,24 +168,23 @@ describe('Dashboards', () => {
       })
     })
 
-    describe('table sort headings', () => {
-      beforeEach(() => {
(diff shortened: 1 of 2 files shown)


integration_tests/integration/dashboards.cy.js
```diff
@@ -168,24 +168,23 @@ describe('Dashboards', () => {
       })
     })
 
-    describe('table sort headings', () => {
-      beforeEach(() => {
-        cy.stubGetSentReferralsForUserTokenPaged(pageFactory.pageContent([]).build())
```
NeilLuptonMoJ, line 173: outer beforeEach (line 46) is already setting this, so not needed

rgforsyth (approved):

PR #1600, opened 2022-07-29
Amend complexity level

rgforsyth:

Allows probation practitioners to amend a complexity level on a referral with an un-approved action plan.

To reduce re-referrals.

### DIFF
--- a/server/services/interventionsService.ts
+++ b/server/services/interventionsService.ts
@@ -225,6 +225,21 @@ export default class InterventionsService {
     })) as DraftReferral
   }
 
+  async amendComplexityLevelForServiceCategory(
+    token: string,
+    referralId: string,
+    serviceCategoryId: string,
+    complexityLevel: Partial<ReferralComplexityLevel>
+  ): Promise<null> {
+    const restClient = this.createRestClient(token)
+
+    return (await restClient.post({
+      path: `/sent-referral/${referralId}/service-category/${serviceCategoryId}/amend-complexity-level`,
+      headers: { Accept: 'application/json' },
+      data: complexityLevel,
+    })) as null
+  }
+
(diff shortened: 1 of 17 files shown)


server/views/amendAReferral/complexityLevel.njk
```diff
@@ -0,0 +1,37 @@
...
+
+        {{ govukRadios(radioButtonArgs) }}
+
+        {% if reasonForChangeInputArgs %}
+          {{ govukTextarea(reasonForChangeInputArgs) }}
+        {% endif %}
+
+        <div class="govuk-button-group">
+          {{ govukButton({ text: "Save and continue" }) }}
+          <a class="govuk-link govuk-!-padding-left-5" data-cy="cancel" href="{{ backLinkArgs.href }}">Cancel and go back</a>
```
NeilLuptonMoJ, line 32: don't think you need the padding-left-5 class anymore - the govuk-button-group should space it correctly

server/services/interventionsService.ts
```diff
@@ -210,6 +210,20 @@ export default class InterventionsService {
     })) as DraftReferral
   }
 
+  async amendComplexityLevelForServiceCategory(
```
NeilLuptonMoJ, line 213: needs tests added to interventionsService.test.ts

server/services/interventionsService.ts
```diff
@@ -210,6 +210,20 @@ export default class InterventionsService {
...
+    token: string,
+    referralId: string,
+    complexityLevel: Partial<ReferralComplexityLevel>
+  ): Promise<AmendReferralDetailsUpdate> {
```
NeilLuptonMoJ, line 217: we are going to return 204 no content for the update endpoints, so this return type should be Promise<null>

NeilLuptonMoJ (approved):
LGTM

PR #1608, opened 2022-08-01
add pagination back to the top of all dashboard that aren't open cases for service providers

rgforsyth:

Adds pagination back to the top of all dashboard that aren't open cases for service providers

Allows easy navigation of the paginated results.


NeilLuptonMoJ (approved):
👍

PR #1609, opened 2022-08-02
Re-added missing content changes

JakeGoodwin22:

Adds back the content changes to the search by PoP which were removed when we reverted changes.

Keep content consistent with designs


NeilLuptonMoJ (approved):
👍

PR #1613, opened 2022-08-05
Changelog view

OmotayoTalabi:

change log ui that shows all amend referrals changes from pp

to view all amend referrals changes.

### DIFF
--- a/server/models/changelog.ts
+++ b/server/models/changelog.ts
@@ -1,0 +1,8 @@
+export default interface Changelog {
+  changelogId: string
+  referralId: string
+  topic: string
+  changedAt: string
+  name: string
+  reasonForChange: string
+}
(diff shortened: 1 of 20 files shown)


rgforsyth:
Think this may need a rebase on main?

server/models/changelog.ts
```diff
@@ -0,0 +1,8 @@
+export default interface Changelog {
+  changelogId: string
+  referralId: string
```
kth13, line 3: will this not capture the old and new value

server/routes/amendAReferral/amendAReferralController.ts
```diff
@@ -132,4 +134,20 @@ export default class AmendAReferralController {
 
     return ControllerUtils.renderWithLayout(res, view, serviceUser)
   }
+
+  async getChangelog(req: Request, res: Response): Promise<void> {
```
kth13, line 138: changelog part of logic can be moved to Changelog controller as it is not part of the amend feature

OmotayoTalabi: created the Changelog controller

kth13: But I think you forgot to remove it from here. 😄

kth13 (approved):
LGTM

PR #1623, opened 2022-08-08
Quick fix for interventions service tests

JakeGoodwin22:

Applies a quick fix to the pact tests which fail after a deployment is 2 days old

To stop the contract tests failing if there has been no deployment for 2 or more days


NeilLuptonMoJ (approved):
LGTM

PR #1626, opened 2022-08-11
all case search by reference number

kth13:

-introduce all case search by reference number.

-we had case search by person on probation but we had a requirement where we need provide the capability to the user to search all cases by reference number.


kth13:
Yes @JakeGoodwin22 its still WIP

JakeGoodwin22:
> Yes @JakeGoodwin22 its still WIP

Sorry, I jumped the gun!

JakeGoodwin22 (approved):
LGTM

PR #1639, opened 2022-08-17
Display notification banner if no changes made

NeilLuptonMoJ:

Shows info banner when no changes are made when amending outcomes

If nothing has changed on desired outcomes and the user goes to save changes an error shouldn’t throw on reason for changing
design: https://www.figma.com/file/UPHNDGeHMPKM6hZcAvZjk2/Working---Master-Design?node-id=7340%3A75906

### DIFF
--- a/integration_tests/integration/amend-referral.cy.js
+++ b/integration_tests/integration/amend-referral.cy.js
@@ -292,6 +292,10 @@ context('Amend a referral', () => {
         )
         cy.contains('What are the desired outcomes for Accommodation?')
 
+        cy.get('[data-cy=desired-outcomes]').within(() => {
+          cy.get(':checkbox').last().check()
+        })
+
         cy.contains('Save changes').click()
 
         cy.get('.govuk-error-summary').within(() => {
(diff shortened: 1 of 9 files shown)


NeilLuptonMoJ:
[8.7 Amend desired outcomes](https://trello.com/c/PmbbsdrV/428-87-amend-desired-outcomes)

integration_tests/integration/amend-referral.cy.js
```diff
@@ -292,6 +292,10 @@ context('Amend a referral', () => {
         )
         cy.contains('What are the desired outcomes for Accommodation?')
 
+        cy.get('[data-cy=desired-outcomes]').within(() => {
```
kth13, line 295: Do we have display this banner only for desired outcomes and not for any other amends?

kth13:
Looks good except one clarification

integration_tests/integration/amend-referral.cy.js
NeilLuptonMoJ: its desired outcomes only, that's the only amend that had this requirement (because the view change detail screen needs to split this into 'added' and 'removed' sections, so it wouldn't work if we allowed no changes to the list of outcomes)

kth13: Ok sure @NeilLuptonMoJ

kth13 (approved):
LGTM

PR #1670, opened 2022-09-01
Changelog enabled

OmotayoTalabi:

to enable change log view

allows the users to view amend changes, we disable the functionality initially. we are enabling the feature.


kth13 (approved):
LGTM

PR #1760, opened 2022-10-20
Change referral details tab text to alternative wording

awhodgson256:
Change referral details tab text to alternative wording for probation-practitioners.   Retain existing wording for service providers.

Change wording for tab, conditional on being a probation practitioner, remains as it was for service users

Makes it clearer that the details can be edit / amended by practitioners


kth13 (approved):
LGTM

PR #1770, opened 2022-10-25
Add banner for database migration downtime

sldblog:

Add a banner for database migration downtime

image

image

To allow us to bring the service down while we migrate the database to PostgreSQL 14.


kth13 (approved):
LGTM

PR #1781, opened 2022-11-03
Rewording for message shown on unhandled error

awhodgson256:
Rewording for message shown on unhandled error - note: stack trace is not shown in PROD but is in all other environments.

Rewording for the unhandled error page, to make it clearer what to do in the event an error can be recovered and claifying that work may not have been saved.

Improved clarity for end user in fault conditions.


kth13 (approved):
LGTM

PR #1807, opened 2022-11-15
Fix integration test

kth13:

-Fixing some edge case failure in the integration testing

- Some test cases might fail in the edge case scenario and this test is for fixing it and making it more robuse

### DIFF
--- a/package.json
+++ b/package.json
@@ -177,6 +177,7 @@
     "jest-pact": "^0.9.4",
     "jsonwebtoken": "^8.5.1",
     "lint-staged": "^13.0.3",
+    "moment": "^2.29.4",
     "mocha": "^10.1.0",
     "mocha-junit-reporter": "^2.1.1",
     "mockdate": "^3.0.5",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -127,6 +127,7 @@
     "jquery": "^3.6.1",
     "jwt-decode": "^3.1.2",
     "md5": "^2.3.0",
+    "moment": "^2.29.4",
```
sldblog, line 130: Can this live under `devDependencies`? Do we use it in production code?

sldblog (commented):
Looks good, just one question about dev/prod dependencies

sldblog:
`package-lock` changes seem to be missing, could those be added, please?

package.json
kth13: Yes not needed. I have moved into ```devDependencies```

sldblog:
Yes please. It contains a snapshot of fixed versions, so everyone can use the same exact versions.

The only thing that we have to watch out for is lockfile version: we should use npm that generates the same lockfile

kth13:
> Yes please. It contains a snapshot of fixed versions, so everyone can use the same exact versions.
> 
> The only thing that we have to watch out for is lockfile version: we should use npm that generates the same lockfile

Thanks @sldblog , I have committed the other change you have suggested.

sldblog (approved):

PR #1844, opened 2022-12-06
unassigned search

kth13:

-introduces search feature for the unassigned tab in the dashboard

-users wanted search feature in the unassigned tab in the dashboard

### DIFF
--- a/server/routes/serviceProviderReferrals/dashboardPresenter.ts
+++ b/server/routes/serviceProviderReferrals/dashboardPresenter.ts
@@ -27,10 +27,7 @@ export default class DashboardPresenter {
     readonly searchText: string | null = null,
     private readonly userInputData: Record<string, string> | null = null
   ) {
-    this.pagination = new Pagination(
-      sentReferralSummaries,
-      this.searchText ? `open-case-search-text=${searchText}` : null
-    )
+    this.pagination = new Pagination(sentReferralSummaries, this.searchText ? `case-search-text=${searchText}` : null)
     const [sortField, sortOrder] = this.requestedSort.split(',')
     this.requestedSortField = sortField
     this.requestedSortOrder = ControllerUtils.sortOrderToAriaSort(sortOrder)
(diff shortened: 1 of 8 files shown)


server/views/serviceProviderReferrals/dashboard.njk
```diff
@@ -18,9 +18,9 @@
 
       {{ mojSubNavigation(subNavArgs) }}
 
-      {% if presenter.dashboardType === 'All open cases' %}
+      {% if (presenter.dashboardType === 'All open cases') or (presenter.dashboardType === 'Unassigned cases')  %}
```
sldblog, line 21: These fragments
```ts
{% if (presenter.dashboardType === 'All open cases') or (presenter.dashboardType === 'Unassigned cases')  %}
```
```ts
{% if (presenter.dashboardType === 'My cases') or (presenter.dashboardType === 'Completed cases')%}
```
```
There are no results for "{{presenter.SearchText | escape }}" in {{presenter.casesType}}
```
```
If you still cannot find it, it may not be in {{presenter.casesType}}. Check in other case lists.
```
look like they are all working around that we don't have two abstractions

- categorical dashboard types (`All open cases` and `open cases` as configuration seems wrong)
- the ability to decide if something is searchable

This seems to be corroborated by fragments like
```ts
  get hrefLinkForSearch(): string {
    if (this.dashboardType === 'All open cases') {
      return `/service-provider/dashboard/all-open-cases`
    }
    if (this.dashboardType === 'Unassigned cases') {
      return `/service-provider/dashboard/Unassigned-cases`
    }
    return ''
  }
```
where this could be properties of a class (where open, unassigned, my, completed could be separate instances)

What do you think?

kth13: I agree with you @sldblog , I have changed the code like this

server/routes/serviceProviderReferrals/dashboardPresenter.ts
```diff
@@ -72,6 +80,8 @@ export default class DashboardPresenter {
   private readonly showAssignedCaseworkerColumn =
     this.dashboardType === 'All open cases' || this.dashboardType === 'Completed cases'
 
+  readonly isSearchable = this.dashboardType === 'All open cases' || this.dashboardType === 'Unassigned cases'
```
sldblog, line 83: Anything dependent on `this.dashboardType` could go to `dashboardDetails` to move away from branching in several places

sldblog (approved):
Would be great to include the comment below -- but I'm happy if it happens as a follow-up

server/routes/serviceProviderReferrals/dashboardPresenter.ts
kth13: Sure will take care as part of next PR

sldblog:
image

Noticed that referral number tests are missing

PR #1848, opened 2022-12-10
Add nDelius maintenance banner for today

sldblog:

Add nDelius maintenance banner for today

This was missed and providers are trying (visible in failures in Sentry). This looks like our service is down.


sldblog:
Bypassing approval as it's urgent and it's the weekend.

PR #1863, opened 2022-12-13
Show system-made case notes with "System" sender

sldblog:

Show system-made case notes with "System" sender

https://github.com/ministryofjustice/hmpps-interventions-service/pull/1353

We have a requirement to display case notes from the "system". We created a "fake" user for this purpose, which does not exist in hmpps-auth.

To capture this, we short-circuit the lookup logic for this particular user and use a canned response.

### DIFF
--- a/server/services/hmppsAuthService.test.ts
+++ b/server/services/hmppsAuthService.test.ts
@@ -73,6 +73,20 @@ describe('hmppsAuthService', () => {
       const output = await hmppsAuthService.getUserDetailsByUsername(token.access_token, username)
       expect(output).toEqual(response)
     })
+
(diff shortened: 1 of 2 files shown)


sldblog:
[(To do after go live)Onboarding: Transfer Bexley and Greenwich cases (only) from London North to London S/SE (Lot1) - on 1st December 2022](https://trello.com/c/3CmaRSrw/589-to-do-after-go-liveonboarding-transfer-bexley-and-greenwich-cases-only-from-london-north-to-london-s-se-lot1-on-1st-december-202)

awhodgson256 (approved):
A simple short circuit to use a dummy user for system.  Looks fine.

sldblog:
🎩 🙇‍♂️

PR #1936, opened 2023-01-24
Allow referral to load even with deleted assignee user

sldblog:

IPB-207: Let the referral pages load even if the caseworker (assignee) user no longer exists in hmpps-auth (the identity provider).

| Before | After |
| --- | --- |
| image | image |

What is worse, the current response from auth won't be visible to end-users, so they have no hint towards what is wrong.

Graceful degradation: an external 404 Not Found response should block other users from viewing or progressing with the referral.

### DIFF
--- a/server/models/hmppsAuth/authUserDetails.test.ts
+++ b/server/models/hmppsAuth/authUserDetails.test.ts
@@ -1,0 +1,17 @@
+import AuthUserDetails, { authUserFullName } from './authUserDetails'
+
+describe('authUserFullName', () => {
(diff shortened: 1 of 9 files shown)


kth13:
@sldblog Is the content team ok with this change?

sldblog:
@kth13 I don't have approval yet.

However, a potentially poorly named "deleted user" is a major win against the current situation (none of the pages load), so waiting for approval will delay people waiting to do work.

kth13 (approved):
LGTM

PR #1937, opened 2023-01-25
Recover from errors on session/draft storage

sldblog:

Partially cover IPB-204: Recover from errors on session/draft storage.

The `redisClient` in `hmppsAuthService.ts` recovers from all intermittent outages, but this main one does not, and the only difference is the `on('error', ...)` handler.

Tested locally, and without this handler, the connection remains dead. With it, it recovers.

(This does not replace a proper liveness check, though -- that one should ping all Redis connections -- rest of IPB-204 is to expose healthchecks)

So that intermittent Redis connection errors will not get the application in a "stuck" state.

Based on the current evidence, I think this will mitigate this pattern:

image
image

### DIFF
--- a/server/app.ts
+++ b/server/app.ts
@@ -118,12 +118,16 @@ export default function createApp(
     },
     password: config.redis.password,
   })
+
   redisClient
     .connect()
     .then(() => logger.info('App Redis connected'))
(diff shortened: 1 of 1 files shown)


kth13 (approved):
LGTM

PR #1942, opened 2023-01-31
Ipb 10/expected release date

kth13:

Creates a new page to capture the expected release date for a custodial referral

We are introducing a new feature to show the users whether the referral is community or custodial. So, If a referral is custodial, its a requirement to capture the expected release date


awhodgson256 (approved):
Looks ok, noted discussion on feature flags, but agreed this won't need flagging as it's the last element of work (and can be reverse merged therefore in the event of a problem).

PR #1943, opened 2023-02-03
Ipb 4/knowing custody status

rgforsyth:

Adds the current location flow to the draft referral form and a start on the check your answers/referral details page. All behind a feature flag.

To allow the ability to store current location and prison information if the user is in custody.


awhodgson256 (commented):
A GET endpoint with a name containing "submit" suggests it should be a POST or PUT.  Is this a pattern that is used elsewhere in the app?

Specifically .... >  describe('GET /referrals/:id/submit-current-location', () => {

rgforsyth:
This is purely due to the copydeck design asking for the url to be `submit-current-location`

awhodgson256:
> This is purely due to the copydeck design asking for the url to be `submit-current-location`

Fair enough if there is a design reasoning...

kth13 (approved):
LGTM

PR #1946, opened 2023-02-07
new cancel referral design

kth13:

Introducing a new way of displaying cancel a referral link. Its now displayed with title and a description to make it more explicit

The cancel referral link was not placed well in the referral details page. That is the intent behind bringing this change


awhodgson256 (approved):
Looks good, I assume this link takes us back to the previous cancel flow.

sldblog:
_suggestion_
For this kind of work, it would be helpful to posterity to provide before/after screenshots in the pull request.

PR #1951, opened 2023-02-13
minor text case fix

kth13:

updates the work `Current Establishment` to `Current establishment`

To be inline with the design specification


awhodgson256 (approved):
Simple key change, looks ok.

PR #1953, opened 2023-02-16
addressing the feedback from UAT

kth13:

- set the back link in expected release date to submitting current location
- showing the correct option when expected release date page is already selected and the user comes again to that page.

- Fixing the issues raised during the UAT


awhodgson256 (approved):
Looks fine - backlink and some minor content changes.

PR #1957, opened 2023-02-17
Wiremock stub for prisoner-offender-search get prisoner endpoint

marcocolluramoj:

Adds wiremock stub for prisoner-offender-search GET /prisoner/{id} endpoint.

To mock prisoner-offender-search API with stub data for local development.


kth13 (approved):
LGTM

rgforsyth:
Looks good. I still believe this should be done in the service however, as this will now mean that the service can no longer be run by its self.

PR #1958, opened 2023-02-20
override the front for autocomplete select options

rgforsyth:

Overrides the font on the of the accessible-autocompletes css classes to be consistent with the rest of the application.

To give consistency to all font on the page.


kth13 (approved):
LGTM

PR #1959, opened 2023-02-22
enabling submit current location and expected release date

kth13:

- Enable submitting the current location feature in pre-prod and prod
- Enable submitting the expected release date feature in pre-prod and prod

- Submitting the current location and expected release date in pre-prod and prod


marcocolluramoj (approved):
LGTM

PR #1960, opened 2023-02-23
configuring different prison register envs for different intervention…

kth13:

- configuring the relevant  prison register envs for the intervention envs

- To configure the relevant prison register envs instead of using prod prison register envs


awhodgson256 (approved):
Looks good Kumar

PR #1966, opened 2023-03-09
all open cases pagination fix for sp dashboard

rgforsyth:

Fix to the service provider all open cases dashboard pagination when there is a search term. 
We will now pull and store the search in session data. Upon accessing the page without the new query param, previous stored data will be cleared so that the page isn't loaded with any previous search terms.

Allows a user to search and also use the pagination. This was broken when we moved to using POST for the search.

### DIFF
--- a/server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
+++ b/server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
@@ -151,6 +151,7 @@ export default class ServiceProviderReferralsController {
   }
 
   async showUnassignedCasesDashboard(req: Request, res: Response): Promise<void> {
+    this.handlePaginatedSearchText(req)
     const searchText = (req.body['case-search-text'] as string) ?? null
     if (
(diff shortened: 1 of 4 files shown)


awhodgson256 (approved):
Looks good Ryan, as you say test in dev with some representative data.

server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
```diff
@@ -110,7 +110,22 @@ export default class ServiceProviderReferralsController {
     this.renderDashboardWithoutPagination(req, res, referralsSummary, 'My cases')
   }
 
+  private handlePaginatedSearchText(req: Request) {
+    if (req.query.paginated === 'true') {
+      req.body['case-search-text'] = req.session.searchText
```
kth13, line 115: why we need to tamper the request body here. instead you can return the search text and assign it in line no 129.
```const searchText = handlePaginatedSearchText(params...)```

PR #1970, opened 2023-03-16
Ipb 367/case search pagination loss

rgforsyth:

fixes the loss of search when the dashboard table is re-ordered.

Currently the dashboard will resort back to an unfiltered table when a new sort order is provided. This change is to prevent that and retain the search


awhodgson256 (approved):
Looks good Ryan

kth13 (approved):
LGTM

PR #1988, opened 2023-04-25
Increase timeout value for auth connections to 20s

awhodgson256:
To reduce timeout errors that are observed in PROD.

Increases timeout value for the UI and AUTH connection

To reduce or eliminate timeout errors when AUTH is responding slowly.


kth13 (approved):
LGTM

PR #2011, opened 2023-06-20
setting right content while capturing pp details

kth13:

- If delius return us with undefined value, we need to replace with Not found
- If the user is not inputting the probation office, then set to Not provided

- The intent is set the right content


marcocolluramoj (approved):
LGTM

PR #2021, opened 2023-07-05
update cancelled banner message

rgforsyth:

Updates the content within the intervention ended banner.

image

The current message displays that an eosr needs to be completed within 10 days, which is incorrect. This should be 5 days.


marcocolluramoj (approved):
LGTM

PR #2023, opened 2023-07-06
fixing referral details showing responsible officer and referring officer

kth13:

- Shows the responsible officer with the information captured from the confirm probation practitioner page
- Shows the referring probation practitioner details only when there is a custody

- When community with com was released the referring probation practitioner was not showing the right information
- The probation practitioner summary card was showing wrong information. So this PR is to correct it


marcocolluramoj (approved):
LGTM

PR #2024, opened 2023-07-06
Ipb 553/bring back community with com

kth13:

- bring back community with com feature for further enhancement

- community with com feature was reverted to allow other features to go in as there was a bug in this. Now since the other features have gone in, we are bringing this back


marcocolluramoj (approved):
LGTM

PR #2038, opened 2023-07-13
fixing null reference error for summary list args

kth13:

- Adding additional checks to check if the line in summary list args is not null
- removing the redundant flag check

- we are receiving null reference error, so this check will safe guard against any null reference


marcocolluramoj (approved):
LGTM

PR #2040, opened 2023-07-14
use feedback submitted by rather than assigned caseworker for appointment feedback page

rgforsyth:

Displays the caseworker that submitted the feedback for an appointment, rather than the currently assigned caseworker.

Currently the wrong caseworker is displayed on the check your answers and feedback page.


server/routes/appointments/appointmentSummary.ts
```diff
@@ -121,8 +121,10 @@ export default class AppointmentSummary {
   }
 
   private caseworkerName(): string {
-    return typeof this.assignedCaseworker === 'string'
-      ? this.assignedCaseworker
-      : `${(<AuthUserDetails>this.assignedCaseworker).firstName} ${(<AuthUserDetails>this.assignedCaseworker).lastName}`
+    return typeof this.feedbackSubmittedBy === 'string'
```
kth13, line 124: wanted to understand if `this.feedbackSubmittedBy` can be a String as well

PR #2045, opened 2023-07-21
fixing referral details page to get phone details from community manager alone

kth13:

Retrieves the phone number and team phone number from community manager instead of checking who is the responsible officer and getting the details

Due to some issue in Delius API which is being investigated, we fall back to use community manager details


awhodgson256 (approved):
LGTM

PR #2049, opened 2023-07-27
configuring more ramdelius mocks for local setup

kth13:

- configures ramdelius api mocks for the local set up to work

- community api is migrated to ramdelius api. so we need to set up mocks for the application to work


rgforsyth (approved):
LGTM

PR #2053, opened 2023-07-28
configure new manage users api

kth13:

- configure hmpps-manage-users to authenticate and authorise external users

- hmpps-auth has been rearchitected which caused their api is to be split into multiple apis.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,16 +15,76 @@ services:
     networks:
       - hmpps
     container_name: hmpps-auth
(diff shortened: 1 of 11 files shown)


amolphalnikar:
Looks good to me but will wait for other devs to double check it.

amolphalnikar (approved):
Looks good to me but will wait for other devs to double check it.

marcocolluramoj (approved):
LGTM

PR #2058, opened 2023-08-02
Fix NPE when submitting feedback with late attendance

marcocolluramoj:

Updates conditional statement so that session feedback is also included in the call to submit appointment feedback when attendance is set to late. 

To fix NullPointerExceptions occurring when submitting feedback for historic appointments marked as late.


kth13 (approved):
LGTM

PR #2062, opened 2023-08-07
Remove Delivery Schedule link text from footer

marcocolluramoj:

Removes the Delivery Schedule link text from footer

Removes the Delivery Schedule link text from footer


awhodgson256 (approved):
LGTM

PR #2068, opened 2023-08-14
add service user banner to case note check your answers page

rgforsyth:

Adds the blue service user banner to the check your answers page when adding a case note.

The banner is currently missing.


awhodgson256 (approved):
LGTM

PR #2076, opened 2023-09-04
enabling info level to test the reporting

kth13:

- Enable level to info to check the s3 bucket url

- The report is not downloading properly. we need to check what is the url generated for s3


marcocolluramoj (approved):
LGTM

PR #2084, opened 2023-09-08
Display service outage banner

marcocolluramoj:

Displays the downtime banner with the following message:
"Refer and monitor an intervention will not be available between 10am on Saturday 09 September and 9am on Sunday 10 September 2023."

To notify users of planned downtime from Saturday 9th September from 10am as nDelius will be shut down and moved into read only access for the introduction of new improvements. As a result Refer and Monitor MAY be impacted.


kth13 (approved):
LGTM

PR #2094, opened 2023-09-20
Update draft oasys risk info when users selects not to edit

marcocolluramoj:

Updates the draft oasys risk information table with the most recent changes (if any) when the users selects that they do not want to edit. 

To prevent the original RiskSummary from ARN overriding the most recent changes when the user navigates back to the risk information page and selects that they do not wish to edit after having already made changes.


kth13 (approved):
LGTM

PR #2095, opened 2023-09-20
fixing the bugs raised in testing

kth13:

- Fix the issue where the user was able to submit when the auto complete was blank
- Fix the issue where the user go backs from the referral type form and prison select form and type the new crn again, it was failing
- Fix the date validations message order

- Fix the bugs raised during testing


server/app.ts
```diff
@@ -96,6 +96,12 @@ export default function createApp(
             "'sha256-pu37IY9Ig/H/vwy6rKyd1hbGd0HCmI3jhdUIhUV1l2A='",
             "'sha256-v8tCeJtLZcGYwtw7d1C1rXv3Tv2Clp+NyDAzPWZvtaY='",
             "'sha256-MoTaAoB/V8e6T3/hagtzBDVnvuUwgTboZeP40OVsw4k='",
+            "'sha256-5jMuXDZya0nKx/E7KeO2htVzsDvBbBGHk54V7ZEFCwM='",
```
rgforsyth, line 99: We probably need to make sure we are removing the ones that are no longer used when updating old scripts. Not necessarily now, but this list will keep growing otherwise, where a bunch are redundant.

rgforsyth (approved):

server/app.ts
kth13: @rgforsyth I agree to your point. probably after this goes in, we can take up that exercise.

rgforsyth: 👍

PR #2097, opened 2023-09-26
Change session references to appointment for SAA feedback

marcocolluramoj:

Change session references to appointment for SAA feedback

Change session references to appointment for SAA feedback


kth13 (approved):
LGTM

PR #2100, opened 2023-09-28
fixing broken pact test

kth13:

- Sets the ppLocationType to null

- Fixes the pact test after the introduction of a new field


awhodgson256 (approved):
LGTM

PR #2101, opened 2023-09-28
fixing pact test 3

kth13:

- Sets the ppLocationType to null in the pact test as it is not required to be validated

- Fixing the failed pact test


awhodgson256 (approved):
LGTM

PR #2106, opened 2023-10-10
change prison release form to new design

kth13:

- change the prison release type form according to the new design

- change the prison release type form according to the new design


awhodgson256 (approved):
LGTM

PR #2107, opened 2023-10-11
changing the content for the prison release form

kth13:

- changing the content for the prison release form

- the content needs to be changed after the stakeholders input


awhodgson256 (approved):
LGTM

PR #2115, opened 2023-10-19
Update session concerns question

marcocolluramoj:

Updates session concerns question

Updates session concerns question


kth13 (approved):
LGTM

PR #2116, opened 2023-10-23
capturing community allocated pp info

kth13:

- capturing the information of whether the referral has community allocated PP in a new page

- This is to mitigate the unstable data from community api. To mitigate the issue, we are capturing the information in R&M itself.


marcocolluramoj (approved):
LGTM

PR #2121, opened 2023-10-26
Enable downtime banner

marcocolluramoj:

Enables the downtime banner.

To notify users of the planned downtime due to nDelius maintenance.


rgforsyth (approved):
LGTM

PR #2122, opened 2023-10-26
Update banner content and make it full width

marcocolluramoj:

Updates downtime banner content and make it full width.

Updates downtime banner content and make it full width.


kth13 (approved):
LGTM

PR #2129, opened 2023-10-31
handling the forbidden user

kth13:

- Handling the forbidden user error (403) by Oasys risk api

- Now Oasys risk is throwing 403 for some functionalities instead of 404. so, we need to handle that


marcocolluramoj (approved):
LGTM

PR #2139, opened 2023-11-24
What's New Banner

marcocolluramoj:

Adds a "What's New" notification banner, linking to a what's new page for more details, which is hidden when dismissed until the banner content is updated with a new version number.  The banner is displayed on most pages and can be updated with userType specific content (SP or PP).

To inform Service Providers and Probation Practitioners of recent releases.

### DIFF
--- a/server/routes/appointments/appointmentsController.ts
+++ b/server/routes/appointments/appointmentsController.ts
@@ -73,7 +73,7 @@ export default class AppointmentsController {
   }
 
   async scheduleSupplierAssessmentAppointment(req: Request, res: Response): Promise<void> {
-    const fetchResult = await this.fetchDraftBookingOrRenderMessage(req, res)
+    const fetchResult = await this.fetchDraftBookingOrRenderMessage(req, res, 'service-provider')
     if (fetchResult.rendered) {
       return
     }
@@ -141,7 +141,7 @@ export default class AppointmentsController {
     )
 
     const view = new ScheduleAppointmentView(presenter)
-    ControllerUtils.renderWithLayout(res, view, serviceUser)
+    await ControllerUtils.renderWithLayout(req, res, view, serviceUser, 'service-provider')
   }
 
   async checkSupplierAssessmentAnswers(req: Request, res: Response): Promise<void> {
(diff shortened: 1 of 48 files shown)


server/routes/appointments/appointmentsController.ts
```diff
@@ -471,7 +486,7 @@ export default class AppointmentsController {
         let basePath
         let redirectPath
         if (draftBookingId) {
-          const fetchResult = await this.fetchDraftAppointmentOrRenderMessage(req, res)
+          const fetchResult = await this.fetchDraftAppointmentOrRenderMessage(req, res, 'service-provider')
```
kth13, line 489: I am seeing you are passing the `userType` variable many places, is it possible to identity from the  request?

server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
```diff
@@ -109,7 +109,7 @@ export default class ServiceProviderReferralsController {
       res.locals.user.token.accessToken,
       SPDashboardType.MyCases
     )
-    this.renderDashboardWithoutPagination(req, res, referralsSummary, 'My cases')
+    await this.renderDashboardWithoutPagination(req, res, referralsSummary, 'My cases')
```
kth13, line 112: why we need to change this to async?

server/routes/appointments/appointmentsController.ts
marcocolluramoj: > I am seeing you are passing the `userType` variable in many places, is it possible to identity from the request?. This comment holds good for all the places where you are passing this.

I don't think it's possible, the 'userType' is already being passed through to some controllers like this and I applied the same approach.

server/routes/serviceProviderReferrals/serviceProviderReferralsController.ts
marcocolluramoj: > why we need to change this to async?. Same question holds good for other places as well

Because we are calling an async function in ReferenceDataService to load the what's new banner content from a csv from `ControllerUtils`.  `MakeAReferralController` also uses the ReferenceDataService and all the functions are async.

server/routes/appointments/appointmentsController.ts
kth13: Ok no worries then.

kth13 (approved):
LGTM

PR #2144, opened 2023-12-01
Make whats new banner content full width

marcocolluramoj:

Makes whats new banner content full width

To make whats new banner content full width


awhodgson256 (approved):
LGTM

PR #2148, opened 2023-12-11
short circuit locals.user to avoid errors when destructuring

marcocolluramoj:

Prevents an error when trying to get userId from res.locals.user for the what's new banner cookie key when it is undefined.

Prevents an error when trying to get userId from res.locals.user for the what's new banner cookie key when it is undefined.


awhodgson256 (approved):
LGTM

PR #2155, opened 2024-01-03
Amend tests based on expected date following turn of the new year

awhodgson256:

Fixes failing tests due to change of year

Keep mainline build working / passing tests

### DIFF
--- a/integration_tests/integration/probationPractitionerReferrals.cy.js
+++ b/integration_tests/integration/probationPractitionerReferrals.cy.js
@@ -472,7 +472,7 @@ describe('Probation practitioner referrals dashboard', () => {
       .should('contain', 'English')
       .should('contain', 'Male')
       .should('contain', 'Agnostic')
-      .should('contain', '1 Jan 1980 (43 years old)')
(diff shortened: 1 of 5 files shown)


integration_tests/integration/probationPractitionerReferrals.cy.js
```diff
@@ -472,7 +472,7 @@ describe('Probation practitioner referrals dashboard', () => {
       .should('contain', 'English')
       .should('contain', 'Male')
       .should('contain', 'Agnostic')
-      .should('contain', '1 Jan 1980 (43 years old)')
+      .should('contain', '1 Jan 1980 (44 years old)')
```
kth13, line 475: @awhodgson256 I feel its worth making this dynamic so that it won't fail the next year.

kth13: @awhodgson256 we have used moment framework for testing the dates dynamically. we can use moment here as well

kth13 (approved):
LGTM

PR #2158, opened 2024-01-08
Fix for updated image allowing existing username login locally

awhodgson256:

Moves credentials for bernard.beaks in offline configuration into wiremock mapping

Restore user credentials typically used in offline test


kth13 (approved):
LGTM

PR #2167, opened 2024-01-26
Ipb 897/change pp phone number

kth13:

- introduce new page for updating pp phone number
- save the pp details retrieved from delius initially

- The changes are done in conjunction to the new design


awhodgson256 (approved):
LGTM

PR #2180, opened 2024-02-14
Add notify probation practitioner of poor behaviour and concerns

marcocolluramoj:

Adds notify probation practitioner of poor behaviour and concerns checkbox to session feedback forms.

To give the user the option to report poor behaviour to the probation practitioner and differentiate from concerning behaviour.


kth13 (approved):
LGTM

PR #2182, opened 2024-02-21
Ipb 3/allocate prioritise cases

kth13:

- new confirm probation practitioner design page
- different update pages for different fields

- The design of the probation practitioner design page is updated and hence forth the changes


awhodgson256 (approved):
LGTM, discussed lookup data change and agreed approach

PR #2183, opened 2024-02-22
using ppName instead of delius ppname for backwards compatability

kth13:

- Amending to save the values in a different place

- It is required to store both pp and delius values in the DB


awhodgson256 (approved):
Looks good Kumar, similar logic to how we were dealing with responsible officer originally

PR #2184, opened 2024-02-22
Enable service outage banner for Auth outage 29th Feb

awhodgson256:

Enables outage banner for upcoming outage

Inform users of service outage


rgforsyth (approved):
Looks good 👍

PR #2187, opened 2024-02-23
fixing pact test

kth13:

- set the allocatedCommunityPP to true
- this is done to set for the pact contract testing to pass

- To fix the pact test


marcocolluramoj (approved):
LGTM

PR #2188, opened 2024-02-27
Add option for sending case note emails to probation practitioners

marcocolluramoj:

Adds a radio button to select whether or not to send an email to the probation practitioner when adding a case note.

To avoid probation practitioners from getting too many emails.


kth13 (approved):
LGTM

PR #2192, opened 2024-02-29
Add case note email opt-out what's new banner entry

marcocolluramoj:

Add case note email opt-out what's new banner entry

Add case note email opt-out what's new banner entry


kth13 (approved):
LGTM

PR #2198, opened 2024-03-01
Fix notify PP of concerns for draft feedback when session didn't happen

marcocolluramoj:

Correctly updates DraftAppointment object with notifyProbationPractitionerOfConcerns  when session didn't happen

Notifies PP of concerns via email as expected when session concerns is selected


kth13 (approved):
LGTM

PR #2201, opened 2024-03-04
fix link to view appointment feedback for probation practitioners

rgforsyth:

Adds the appointmentID so the correct url can be constructed to view appointment feedback as a pp.

When viewing appointment feedback as a pp, the wrong appointment is being shown in certain cases.


kth13 (approved):
LGTM

PR #2204, opened 2024-03-11
referral details page with the refined design

kth13:

- referral details page with the new design
- showing the released date
- integrating with the new intervention api which gives the prisoner data

- changes for the new referral details design


awhodgson256 (approved):
LGTM

PR #2205, opened 2024-03-11
Fix send case note email answer

marcocolluramoj:

Fixes an issue where the case note answer for whether or not to send an email doesn't match what was actually selected.

Fixes an issue where the case note answer for whether or not to send an email doesn't match what was actually selected.


kth13 (approved):
LGTM

PR #2207, opened 2024-03-13
Ipb 986/confirm current location

kth13:

- changing the design of submit current location. 
- We have a new radio button to say if the prison name is already known, if not input the prison name

- the design of the submit current location is changed


marcocolluramoj (approved):
LGTM

PR #2209, opened 2024-03-15
redirecting correctly when changing the reason for referral

kth13:

- fixing the issue of accessing reason for referral when change link
- The change link is present in check all referral information page and referral details page

- a bug was reported that changing the reason for referral is not redirecting to the right page


marcocolluramoj (approved):
LGTM

PR #2214, opened 2024-03-21
Ipb 986/confirm current location 2

kth13:

- fall back to old design if prisoner details is not returned
- change the path of prisoner details url

- falling back to old design if prisoner details is not returned


marcocolluramoj (approved):
LGTM

PR #2216, opened 2024-03-25
Expected release date

marcocolluramoj:

Replaces the expected release date form's radio buttons with a confirm option displaying the release date pulled from Prisoner Search API and adds new dedicated pages for the options to enter the expected release date or the reason it is not known.

Replaces the expected release date form's radio buttons with a confirm option displaying the release date pulled from Prisoner Search API and adds dedicated pages for the options to enter the expected release date or the reason it is not known.


kth13 (approved):
LGTM

PR #2225, opened 2024-04-08
Amend error messages as defined in copydecks

awhodgson256:

Amends error feedback on main point of contact form

Clarify wording to match the form's behavior


kth13 (approved):
LGTM

PR #2227, opened 2024-04-10
calling a new api to save new user

kth13:

- call `/intervention/user` to save any new user before the dashboard api call

- When the dashboard api was pointing to read replica db, the new users trying to access the db, the new users were not able to store in our auth db.
So this new api will store the users first and then the dashboard api will not need to store the new user which will allow them to point it back to read replica


awhodgson256 (approved):
LGTM

PR #2230, opened 2024-04-16
do not show deactivated users name

kth13:

- Do not show the deactivated users
- show as `Deactived R&M account`

- Due to GDPR issue, we should not show deactivated user name


awhodgson256 (approved):
LGTM

PR #2231, opened 2024-04-17
Ipb 1147/radio button validation state error

kth13:

- maintain the state when there is validation error in the input boxes inside a radio button

- maintain the state when there is validation error in the input boxes inside a radio button


awhodgson256 (approved):
LGTM

PR #2233, opened 2024-04-19
remove the email opt in summary for PP

kth13:

- hide email opt in summary in case notes for PP

- That message is intended for PP, so we don't need to show it in case notes summary created by PP


awhodgson256 (approved):
LGTM

PR #2238, opened 2024-05-08
Ipb 1121/withdraw referral

rgforsyth:

Adds the withdraw referral flow for pre-ICA referrals

Allow referrals to be withdrawn

### DIFF
--- a/server/routes/appointments/feedback/shared/sessionFeedback/sessionFeedbackView.ts
+++ b/server/routes/appointments/feedback/shared/sessionFeedback/sessionFeedbackView.ts
@@ -146,6 +146,7 @@ export default class SessionFeedbackView {
         },
         {
           divider: 'or',
+          class: 'govuk-label--m',
         },
         {
           id: 'noNotifyPPCheckbox',
--- a/server/utils/nunjucksSetup.ts
+++ b/server/utils/nunjucksSetup.ts
@@ -24,4 +24,11 @@ export default function nunjucksSetup(app: express.Application, path: pathModule
   Object.keys(mojFilters).forEach(filterName => {
     njkEnv.addFilter(filterName, mojFilters[filterName])
   })
+
+  // eslint-disable-next-line func-names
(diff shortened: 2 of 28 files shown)


server/routes/appointments/feedback/shared/sessionFeedback/sessionFeedbackView.ts
```diff
@@ -146,6 +146,7 @@ export default class SessionFeedbackView {
         },
         {
           divider: 'or',
+          class: 'govuk-label--m',
```
kth13, line 149: Is this related to this change or you are fixing something that you found wrong?

server/utils/nunjucksSetup.ts
```diff
@@ -24,4 +24,11 @@ export default function nunjucksSetup(app: express.Application, path: pathModule
   Object.keys(mojFilters).forEach(filterName => {
     njkEnv.addFilter(filterName, mojFilters[filterName])
   })
+
+  // eslint-disable-next-line func-names
+  njkEnv.addFilter('setAttribute', function (dictionary, key, value) {
```
kth13, line 29: why we need add this filter

kth13 (commented):
Overall looking good. Have got questions and some small suggestions

server/routes/appointments/feedback/shared/sessionFeedback/sessionFeedbackView.ts
rgforsyth: I think the default divider text was too small and didn't match the design.

server/utils/nunjucksSetup.ts
rgforsyth: This was to solve an issue of having multiple text boxes share the same data field. It allowed us to update them with unique id's and set the govuk text box for each radio within the njk file itself, whilst also keeping the radio list separated into different sections, but treated as one list.

PR #2242, opened 2024-05-21
Ipb 1122/withdrawl referral post ica

rgforsyth:

Adds functionality for Post and Pre-ica withdrawals

To allow referrals to be withdrawn rather than cancelled.


kth13 (commented):
most of stuffs are brought from the other PR. So, lets push that changes and check again

kth13 (approved):
LGTM

PR #2245, opened 2024-05-22
changing the prison api url

kth13:

- update prison API urls for prod and pre-prod

- update prison API urls for prod and pre-prod


awhodgson256 (approved):
LGTM

PR #2247, opened 2024-05-28
capturing expected probation office and expected office unknown reason

kth13:

- capturing the expected probation office for the referral
- capturing the expected probation unknown reason for the referral
- displaying the captured information in summary and referral details page

- PP needs to input the expected probation office where there referral is going to go
- This will help SP to gain information about the referral


awhodgson256 (approved):
LGTM

PR #2252, opened 2024-06-06
capturing the reason for referral before allocation

kth13:

- adding a new page to capture reason for referral creation before allocation
- This is only for referrals with unallocated com

- this is due to a new requirement where the pp wants to input the reason for referral creation for unallocated referral


awhodgson256 (approved):
LGTM

PR #2257, opened 2024-06-17
amend the reason for referral feature

kth13:

- amending the reason for referral from referral details page

- PP should be given the feature to update the reason for referral


awhodgson256 (approved):
LGTM

PR #2260, opened 2024-06-19
downtime banner between 22-23 june

kth13:

- changing the downtime banner message

- changing the downtime banner message


awhodgson256 (approved):
LGTM

PR #2262, opened 2024-06-20
sp dashboard completed to show unassigned referrals

kth13:

- show unassigned referrals in sp dashboard completed tab

- after withdrawn feature is introduced , we need to show the cancelled referrals in the sp completed view irrespective of whether its assigned or not


awhodgson256 (approved):
LGTM

PR #2273, opened 2024-07-08
update code WOR to correct text for intervention ended box

rgforsyth:

update code WOR to correct text for intervention ended box.

Text is currently incorrect.


awhodgson256 (approved):
LGTM

PR #2274, opened 2024-07-08
change log details for expected release date

kth13:

- deriving details for the expected release date
- changes for different scenarios of amending expected release date

- change log needs to be displayed after introducing amendment for expected release date


awhodgson256 (approved):
LGTM

PR #2287, opened 2024-07-25
add saa history to pp progress page

rgforsyth:

Shows saa appointment history on the pp progress page.

Shows more information to the pp's about appointments that have taken place.


kth13 (approved):
LGTM

PR #2292, opened 2024-07-30
Show expected PDU when expected probation office not selected

marcocolluramoj:

Show the expected PDU instead of the expected probation office in the show referral view when the expected probation office was not provided in the referral creation.

Show the expected PDU instead of the expected probation office in the show referral view when the expected probation office was not provided in the referral creation.


kth13 (approved):
LGTM

PR #2294, opened 2024-08-01
add new banner and whats new content

rgforsyth:

add new banner and whats new content to pp view.

Keep pp users updated on recent changes.


server/views/probationPractitionerReferrals/whatsNew.njk
```diff
@@ -11,6 +11,16 @@
       <h1 class="govuk-heading-l">What's new</h1>
       <p>Find out about recent changes to Refer and monitor an intervention.</p>
 
+    {% set whatsNewVersion6Html %}
+        <h3 class="govuk-heading-s">Amending the probation office</h3>
```
kth13, line 15: @rgforsyth we had other amends like amending prison establishment and expected release date was also released. Do we need to add that too?

kth13 (approved):

PR #2310, opened 2024-09-02
Filter out cancelled referrals from My Cases and Unassigned tabs

marcocolluramoj:

Sets cancelled to false in GetSentReferralsFilterParams for the My Cases and Unassigned Cases dashboards

To filter out cancelled referrals from the My Cases and Unassigned Cases dashboard tabs.


awhodgson256 (approved):
LGTM

PR #2322, opened 2024-10-08
QA fixes for history of rescheduled session

marcocolluramoj:

Includes rescheduled in the appointments filter to return rescheduled appointments for PP's and checks attendance is null in addition to the appointment superseded check when determining whether session status is rescheduled. 

To fix the following issues:
- PP cannot see the history of rescheduled sessions
- For the delivery session, the status Did Not Attend and Did Not Happen is being overwritten by Rescheduled


awhodgson256 (approved):
LGTM

PR #2323, opened 2024-10-08
Ipb 1063/update pp email address

awhodgson256:

Adds ability to change email address to the amend a referral flow

Allow PP user to amend their email address used within R&M

### DIFF
--- a/server/models/referralProbationPractitionerEmail.ts
+++ b/server/models/referralProbationPractitionerEmail.ts
@@ -1,0 +1,3 @@
+export interface AmendProbationPractitionerEmailUpdate {
+  ppEmail: string
+}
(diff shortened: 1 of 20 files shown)


server/models/referralProbationPractitionerEmail.ts
```diff
@@ -0,0 +1,3 @@
+export interface AmendProbationPractitionerEmailUpdate {
+  ppEmail: string
```
rgforsyth, line 2: I noticed in the service, the dto included:
val preventEmailNotification: Boolean? = null

Should this be included here, or not in the service?
I may have missed something though.

awhodgson256: It was required by the interface, but I don't have any guidance either way if it's needed or not.  I've assumed we include it but default to null in this case and add an actual value if it becomes necessary in a later change.

@rgforsyth

rgforsyth: 👍

rgforsyth (approved):
LGTM

PR #2389, opened 2025-04-15
Revert accidental change of 'participant' in pact circle ci config from ACP UI back to Interventions UI

Jack-Dempster:
This pr reverts the accidental change of 'participant' in the circle ci config when I was upgrading the pact version. It currently points to `Accredited Programmes UI` when it should point to `Interventions UI`


thomaswilsonxyz (approved):
lgtm

PR #2483, opened 2025-09-22
IPB-1552: Removed call to display downtime banner

Davansah:

Removes downtime banner from the UI

To no longer display the downtime banner as normal service is resumed


kth13 (approved):
LGTM

PR #2563, opened 2026-04-15
adding a query parameter to intervention api

kth13:

- adding a query parameter check end date to intervention api to make
- pass the query parameter only when I call find intervention page

- This is to fix the issue caused by introducing referral end date


moj-keith-au (approved):
LGTM

PR #2569, opened 2026-05-05
Revert "downtime banner (#2566)"

kth13:
This reverts commit 79306d1bd39f73391c630b9815b94e45a13b8001.

- removes the downtime banner added for IPB-2287

- The downtime banner time is expired. So removing the banner


TSwann-justice (approved):
Looks good!

PR #2570, opened 2026-05-06
fixing the slack notification issue

kth13:

- Fixes the slack notification failure when its notifying slack on prod deployment

- Slack notification is currently broken due to an error in string interpolation


Davansah (approved):
LGTM

PR #2571, opened 2026-05-13
downtime banner 13th May 2026

kth13:

- Set down time banner of R&M coinciding with ndelius maintenance

- Ndelius will be down and due to that we have let the users know


moj-keith-au (approved):
Look good

PR #2576, opened 2026-06-16
enabling mod security in dev environment

kth13:

- configure modsec security in the dev ingress.

- This has been raised by the devops as the application is accessible from the internet


moj-keith-au (approved):
LGTM

PR #2579, opened 2026-07-20
add delius location id to Leeds waterloo house

kth13:

- adding location id to Leeds waterloo house

- Since location id was not there, waterloo house was not visible


moj-keith-au (approved):
LGTM

PR #2580, opened 2026-07-21
down time banner for 24th July

kth13:

- Enables the down time banner for apps to be down by 24th July and will be back by 27th July

- Delius is down on maintenance during that period


Davansah (approved):
LGTM

PR #2583, opened 2026-07-27
remove outage banner

kth13:

- Remove outage banner due to delius maintenance

- Remove outage banner due to delius maintenance


Davansah (approved):
LGTM

